PR #2, opened 2019-01-21
Added the initial gradle build with scala support.

meln1k:
Summary: now we have a build structure with scala application support.


jeschkies:
Should we change the `.gitignore` for Gradle files?

src/main/scala/com/mesosphere/usi/Main.scala
```diff
@@ -0,0 +1,7 @@
+package com.mesosphere.usi
```
jeschkies: I think we should start right away with a multi-project setup. What do you think about
* `interface`
* `core`
* `hello-world`

With `core` depending on `interface` and `hello-world` depending on both and being a runnable?

src/main/scala/com/mesosphere/usi/Main.scala
```diff
@@ -0,0 +1,7 @@
...
+
+object Main extends App {
+
+  println("Starting orchestrator")
```
zen-dog: Don't think we're going to "start an orchestrator" here ;)

src/test/scala/com/mesosphere/usi/DummySuite.scala
```diff
@@ -0,0 +1,15 @@
...
+ */
+package com.mesosphere.usi
+
+import org.junit.runner.RunWith
+import org.scalatest.FunSuite
+import org.scalatest.junit.JUnitRunner
+
+@RunWith(classOf[JUnitRunner])
+class DummySuite extends FunSuite {
+  test("true is always true") {
```
zen-dog: Isn't it a philosophical question? :D

zen-dog (approved):
🚢

jeschkies (changes requested):

src/main/scala/com/mesosphere/usi/Main.scala
meln1k: Sounds good, will do in a second

build.gradle
```diff
@@ -0,0 +1,52 @@
...
+ * This file was generated by the Gradle 'init' task.
+ *
+ * This generated file contains a sample Scala library project to get you started.
+ * For more details take a look at the Scala plugin chapter in the Gradle
+ * user guide available at https://docs.gradle.org/5.1.1/userguide/scala_plugin.html
+ */
+
+plugins {
+    // Apply the scala plugin to add support for Scala
+    id 'scala'
```
jeschkies: Why is this in `build.gradle` and not `example/build.gradle`?

build.gradle
```diff
@@ -0,0 +1,52 @@
+/*
+ * This file was generated by the Gradle 'init' task.
+ *
+ * This generated file contains a sample Scala library project to get you started.
+ * For more details take a look at the Scala plugin chapter in the Gradle
+ * user guide available at https://docs.gradle.org/5.1.1/userguide/scala_plugin.html
```
jeschkies: I don't think we need these comments.

core/build.gradle
```diff
@@ -0,0 +1,8 @@
+plugins {
+    // Apply the scala plugin to add support for Scala
+    id 'scala'
+}
+
+dependencies {
+    compile project(':interface')
```
jeschkies, line 2: Nice!

example-orchestrator/build.gradle
```diff
@@ -0,0 +1,15 @@
+plugins {
```
jeschkies: We called the example `hello-world` if I'm not mistaken.

settings.gradle
```diff
@@ -0,0 +1,14 @@
+/*
+ * This file was generated by the Gradle 'init' task.
+ *
+ * The settings file is used to specify which projects to include in your build.
+ *
+ * Detailed information about configuring a multi-project build in Gradle can be found
+ * in the user guide at https://docs.gradle.org/5.1.1/userguide/multi_project_builds.html
```
jeschkies: I don't think these are required.

interface/src/main/scala/com/mesosphere/usi/interface/DummySchedulerInterface.scala
```diff
@@ -0,0 +1,20 @@
+package com.mesosphere.usi.interface
+
+import com.mesosphere.usi.interface.DummySchedulerInterface._
```
jeschkies: No wildcard imports 😉

jeschkies (changes requested):
Thanks. This looks much better to me. The `build.gradle` in root should be much leaner, though. At least it feels like that to me.

jeschkies (approved):
Given you discuss the example name with @meichstedt this looks great to me!

PR #9, opened 2019-01-24
Publish Scaladoc to Github pages.

jeschkies:
Summary:
This will aggregate all Scala documentations into one under the `api` folder on the
`gh-pages` branch. It will also copy all content from `./docs` to the `gh-pages`.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,12 +1,11 @@
 language: scala
 scala:
    - 2.12.2
+script:
+  - ./gradlew ci
 before_cache:
   - rm -f  $HOME/.gradle/caches/modules-2/modules-2.lock
   - rm -fr $HOME/.gradle/caches/*/plugin-resolution/
-after_script:
-  - if [[ ! -z "$GIT_PASSWORD" ]]; then ./gradlew githubReport; fi
-  - if [[ "$TRAVIS_BRANCH" = "master" -a "$TRAVIS_PULL_REQUEST" = "false" ]]; then ./gradlew gitPublishPush; fi
 cache:
   directories:
     - $HOME/.gradle/caches/
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,7 +5,8 @@ before_cache:
   - rm -f  $HOME/.gradle/caches/modules-2/modules-2.lock
   - rm -fr $HOME/.gradle/caches/*/plugin-resolution/
 after_script:
-  - 'if [[ ! -z "$GIT_PASSWORD" ]]; then ./gradlew githubReport; fi'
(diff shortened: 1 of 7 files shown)


.travis.yml
```diff
@@ -5,7 +5,8 @@ before_cache:
   - rm -f  $HOME/.gradle/caches/modules-2/modules-2.lock
   - rm -fr $HOME/.gradle/caches/*/plugin-resolution/
 after_script:
-  - 'if [[ ! -z "$GIT_PASSWORD" ]]; then ./gradlew githubReport; fi'
+  - if [[ ! -z "$GIT_PASSWORD" ]]; then ./gradlew githubReport; fi
```
zen-dog: Do you think, we should move this bash `if [[ ! -z ??? ]]` logic into the tasks?

.travis.yml
```diff
@@ -5,7 +5,8 @@ before_cache:
...
+  - if [[ "$TRAVIS_BRANCH" = "master" -a "$TRAVIS_PULL_REQUEST" = "false" ]]; then ./gradlew gitPublishPush; fi
```
zen-dog: Same idea here - how about passing the branch to the task: `gitPublishPush "$TRAVIS_BRANCH"`?

zen-dog (changes requested):
See my comments

.travis.yml
jeschkies: I was wondering the same. Should I add `./gradlew ci` or `./gradlew after_script`? On the other side this is fairly Travis specific.

.travis.yml
jeschkies: 🤔 that's a good idea. The issues is that `TRAVIS_BRANCH` is the target branch for PRs unless the build is not a PR. I want this to be executed only when we are building master. Should we add the logic to Gradle as we have for Marathon?

```groovy
task master {
  description "Build target for master branch."
}
task pr {
  description "Build target for pull requests."
}
task ci {
  description "Build target for CIs such as Travis CI or Jenkins."
  if (isMasterBuild) dependsOn(master)
  else dependsOn(pr)
}
```

zen-dog: Sgtm

alenkacz: +1

alenkacz (approved):
Oh gradle... I like you so much.

Thanks Karsten, looking good (just the comment I +1d would be nice to change)!

meln1k (approved):

zen-dog (approved):
Nice!

kensipe:
I reorganized the build file.   I would love to see us keep `build.gradle` clean if we can.   If you don't like... please revert.

kensipe (approved):

kensipe:
if you revert... there are several redundant code snip its and a misspelling that we should resolve after the revert.

PR #11, opened 2019-01-25
Added a persistence and core-models project modules.

meln1k:
Persistence contains an in-memory implementation of the KV store, does serialization and range scans.

JIRA issues: DCOS-47508


core-models/src/main/scala/com/mesosphere/usi/core/models/PodRecord.scala
```diff
@@ -0,0 +1,16 @@
+package com.mesosphere.usi.core.models
+
+import java.time.Instant
+
+
+case class PodRecord(sequenceNr: Long,
```
jeschkies: Please add documentation.

core-models/src/main/scala/com/mesosphere/usi/core/models/PodRecord.scala
```diff
@@ -0,0 +1,16 @@
...
+                     podId: PodId,
+                     launchedAt: Instant,
+                     unserviceable: Option[UnserviceableDetails])
```
jeschkies: Do we want to use this formatting? We probably should add scalairform.

persistence/src/main/scala/com/mesosphere/usi/persistence/InMemoryRepositoryImpl.scala
```diff
@@ -0,0 +1,42 @@
+package com.mesosphere.usi.persistence
+
+import akka.{Done, NotUsed}
+import akka.stream.scaladsl.Source
+import com.mesosphere.usi.core.models.{PodRecord, ReservationRecord}
+
+import scala.collection.concurrent.TrieMap
+import scala.concurrent.Future
+
+class InMemoryRepositoryImpl extends Repository {
```
jeschkies: Should the `impl` be part of this? As an example this should be fine but the Zookeeper one should be in `persistence-zookeper` or am I missing something? @zen-dog

persistence/src/main/scala/com/mesosphere/usi/persistence/Repository.scala
```diff
@@ -0,0 +1,19 @@
+package com.mesosphere.usi.persistence
+
+import akka.{Done, NotUsed}
+import akka.stream.scaladsl.Source
+import com.mesosphere.usi.core.models.{PodRecord, ReservationRecord}
+
+import scala.concurrent.Future
+
+trait Repository {
```
jeschkies: Please add docs.

persistence/src/main/scala/com/mesosphere/usi/persistence/Repository.scala
```diff
@@ -0,0 +1,19 @@
...
+
+  def storePodRecord(record: PodRecord): Future[Done]
+
+  def storeReservationRecord(record: ReservationRecord): Future[Done]
+
+  def podRecords(fromSequenceNr: Long = 0, toSequenceNr: Long = Long.MaxValue): Source[PodRecord, NotUsed]
```
jeschkies: What are these parameters?

persistence/src/main/scala/com/mesosphere/usi/persistence/Repository.scala
```diff
@@ -0,0 +1,19 @@
...
+
+import akka.{Done, NotUsed}
+import akka.stream.scaladsl.Source
+import com.mesosphere.usi.core.models.{PodRecord, ReservationRecord}
+
+import scala.concurrent.Future
+
+trait Repository {
+
+  def storePodRecord(record: PodRecord): Future[Done]
```
jeschkies: What kind of interface do we want? CRUD or something else? Should we follow @zen-dog, template storage more?

persistence/src/main/scala/com/mesosphere/usi/persistence/storage/InMemoryStorage.scala
```diff
@@ -0,0 +1,23 @@
+package com.mesosphere.usi.persistence.storage
+
+import java.util.concurrent.ConcurrentHashMap
+
+import akka.Done
+import akka.util.ByteString
+
+import scala.concurrent.Future
+
+class InMemoryStorage extends KVStorage {
```
jeschkies: I'm a little confused. Is this an interface or an implementation?

persistence/src/main/scala/com/mesosphere/usi/persistence/storage/Serialization.scala
```diff
@@ -0,0 +1,18 @@
+package com.mesosphere.usi.persistence.storage
```
jeschkies: Should this be really under `storage`? Seems more like a `serialization` package to me.

jeschkies (changes requested):
I'm not sure if I'd call it `persistence`. I like `repository` but that's just me.

Please add documentation. Especially the models and interfaces need these.

I did not look too close at the code since I had a lot of questions for the package names etc. It is not clear to me what the interface is and what's implementation details. Eg is `KVStorage` an interface a Zookeeper implementation should implement or is this just a detail for you in-memory implementation?

jeschkies:
Oh, could you also remove `interfaces`.

persistence/src/main/scala/com/mesosphere/usi/persistence/InMemoryRepositoryImpl.scala
zen-dog: That's what I would do. And it's a personal preference but I always hated the `Impl` suffixes.

persistence/src/main/scala/com/mesosphere/usi/persistence/Repository.scala
zen-dog: I guess this is very much influenced by FDB? These implementation details should not leak into the interface.

persistence/src/main/scala/com/mesosphere/usi/persistence/KVRepositoryImpl.scala
```diff
@@ -0,0 +1,58 @@
...
+    storage.write(key, value)
+  }
+
+  override def storeReservationRecord(record: ReservationRecord): Future[Done] = {
+    val value = Serialization.toBinary(record)
+    val key = reservatonRecordSeqNr2Bytes(record.sequenceNr)
+    storage.write(key, value)
+  }
+
+  override def podRecords(fromSequenceNr: Long, toSequenceNr: Long): Source[PodRecord, NotUsed] = {
```
zen-dog: what are those sequence numbers with pod records? where do they come from? I don't know how we identify individual records but I don't think it'll be a `Long`.

persistence/src/main/scala/com/mesosphere/usi/persistence/SerializationFormats.scala
```diff
@@ -0,0 +1,43 @@
...
+import java.io.{ByteArrayInputStream, ByteArrayOutputStream, ObjectInputStream, ObjectOutputStream}
+
+import com.mesosphere.usi.core.models.{PodRecord, ReservationRecord}
+
+import scala.reflect.ClassTag
+
+object SerializationFormats {
+
+  /**
+    * Java serialization for in-memory tests. Don't use in production!
```
zen-dog: Why would we use different serializations in test and prod?

core-models/src/main/scala/com/mesosphere/usi/core/models/PodRecord.scala
```diff
@@ -0,0 +1,27 @@
...
+/**
+  * Pod record is an immutable event of what happened to all pods.
+  * @param sequenceNr number used for ordering the pod events.
+  * @param podId id of the pod
+  * @param launchedAt when the pod was launched
+  * @param unserviceable if not none, pod is not in a serviceable state
+  */
+case class PodRecord(sequenceNr: Long,
+                     podId: PodId,
+                     launchedAt: Instant,
```
meichstedt: I'd like to challenge the decision to name this `launchedAt` as it's ambiguous: does it refer to the point in time the podRecord/Pod was created? Or launched in Mesos? Or matched against an offer? The stations in a pod's life are something like
- created
- provisioned (matched against an offer)
- reported as staging from Mesos
- reported as starting
- reported as running

So, when was the pod **launched**? @timcharper how does `podRecord.launchedAt` relate to `podStatus.statusTimestamp`?

core-models/src/main/scala/com/mesosphere/usi/core/models/PodRecord.scala
```diff
@@ -0,0 +1,27 @@
+package com.mesosphere.usi.core.models
+
+import java.time.Instant
+
+/**
+  * Pod record is an immutable event of what happened to all pods.
+  * @param sequenceNr number used for ordering the pod events.
+  * @param podId id of the pod
+  * @param launchedAt when the pod was launched
+  * @param unserviceable if not none, pod is not in a serviceable state
```
jeschkies: What is a `serviceable` state?

core-models/src/main/scala/com/mesosphere/usi/core/models/PodRecord.scala
```diff
@@ -0,0 +1,27 @@
+package com.mesosphere.usi.core.models
+
+import java.time.Instant
+
+/**
+  * Pod record is an immutable event of what happened to all pods.
+  * @param sequenceNr number used for ordering the pod events.
```
jeschkies: Not sure if I understand. Is this a sequential id?

core-models/src/main/scala/com/mesosphere/usi/core/models/PodRecord.scala
```diff
@@ -0,0 +1,27 @@
...
+  * @param launchedAt when the pod was launched
+  * @param unserviceable if not none, pod is not in a serviceable state
+  */
+case class PodRecord(sequenceNr: Long,
+                     podId: PodId,
+                     launchedAt: Instant,
+                     unserviceable: Option[UnserviceableDetails])
+
+/**
+  * Details about the pod being in an unserviceable state
```
jeschkies: That does not really explain what `unserviceable` means.

core-models/src/main/scala/com/mesosphere/usi/core/models/ReservationRecord.scala
```diff
@@ -0,0 +1,13 @@
+package com.mesosphere.usi.core.models
+
+/**
+  * * ReservationRecord is an immutable event of created reservations .
```
jeschkies: Not sure if this really explains a reservation. Should we link to the reservation definition?

Nit: One `*` to many 🙂

core-models/src/main/scala/com/mesosphere/usi/core/models/ReservationRecord.scala
```diff
@@ -0,0 +1,13 @@
...
+  * @param sequenceNr number used for ordering the reservation record events.
```
jeschkies: `sequenceNr` is an unknown concept for me. Could you explain it to me?

core-models/src/main/scala/com/mesosphere/usi/core/models/Resources.scala
```diff
@@ -0,0 +1,10 @@
+package com.mesosphere.usi.core.models
+
+/**
+  * Resources that were reserved with a reservation
```
jeschkies: Not sure if that makes sense. Isn't this a [Mesos resource](http://mesos.apache.org/documentation/attributes-resources/)? Should we reference them?

persistence/src/main/scala/com/mesosphere/usi/persistence/KVRepositoryImpl.scala
```diff
@@ -0,0 +1,63 @@
...
+import akka.stream.scaladsl.Source
+import com.mesosphere.usi.core.models.{PodRecord, ReservationRecord}
+import com.mesosphere.usi.persistence.storage.KVStorage
+import com.mesosphere.usi.persistence.serialization.SerializationFormats.{javaSerializationPodRecord, javaSerializationReservationRecord}
+import com.mesosphere.usi.persistence.serialization.Serialization
+
+import scala.concurrent.Future
+
+/**
+  * Implementation of the repository that uses an underlying key-value storage for storing records.
```
jeschkies: Could you link respository
```suggestion
  * Implementation of the [[com.mesosphere.usi.persistence.Repository]] that uses an underlying key-value storage for storing records.
```

persistence/src/main/scala/com/mesosphere/usi/persistence/Repository.scala
```diff
@@ -0,0 +1,50 @@
...
+
+import akka.{Done, NotUsed}
+import akka.stream.scaladsl.Source
+import com.mesosphere.usi.core.models.{PodRecord, ReservationRecord}
+
+import scala.concurrent.Future
+
+
+/**
+  * Base trait for storing records which are used by the Scheduler.
```
jeschkies: Should we link to the concept of `records`?

jeschkies (commented):
Some comments on the docs. Thanks for adding them!

core-models/src/main/scala/com/mesosphere/usi/core/models/PodRecord.scala
meln1k: sounds good, I think it's the default in intellij, I don't have a strong opinion about this.

kensipe: I really want the default format of intellij... or details on how to change it to match scalarform

core-models/src/main/scala/com/mesosphere/usi/core/models/PodRecord.scala
meichstedt: Suggestion: let's leave out `unserviceable` and `UnserviceableDetails` for now. Let's not support (health) checks in the initial interface, and add it later with a fully designed spec.

Questions:
- would readiness also be rendered as `unserviceable`?
- How do we express that a pod was last seen unhealthy and is now unreachable?
- How would we expose the payload of a (health) check?
- How would we expose the combination of a health check **and** a check? (Mesos supports having each of these.)

As a reference, see the [`CheckStatusInfo`](https://github.com/apache/mesos/blob/40210dbcfb218d6ef3f647b047afbdda24133f4d/include/mesos/v1/mesos.proto#L2464), which will have a payload depending on the type of check.

I'd defer this part for later design and decision.

core-models/src/main/scala/com/mesosphere/usi/core/models/PodRecord.scala
kensipe: I don't understand this either

persistence/src/main/scala/com/mesosphere/usi/persistence/InMemoryRepositoryImpl.scala
kensipe: I don't care as long as we are consistent and it is intuitive

core-models/src/main/scala/com/mesosphere/usi/core/models/ReservationRecord.scala
```diff
@@ -0,0 +1,13 @@
...
+  * @param reservationId id of the reservation
+  * @param agentId id of the agent
+  * @param resources reserved resources
+  */
+case class ReservationRecord(sequenceNr: Long,
```
meichstedt: Same as for `unserviceable` – I'd leave this whole class out for now since we're not going to use/implement it right away. It'll just create clutter that needs to be touched and refactored. Let's add it once we add support for persistent volumes..

persistence/src/main/scala/com/mesosphere/usi/persistence/InMemoryRepositoryImpl.scala
meichstedt: Yeah please rename to `InMemoryRepository` and move it to a dedicated package.

kensipe (changes requested):
It seems there are at least 3 of us that don't understand the details surrounding seq nums.

I want to make sure we get formatting as close to right as early as possible.   I would prefer it to be close to the default of intellij which is a sensible format IMO.   if not... I want an export of the format on a wiki / s3 we can advertise for devs to use.

Good start.   I

persistence/src/main/scala/com/mesosphere/usi/persistence/InMemoryRepositoryImpl.scala
```diff
@@ -0,0 +1,46 @@
+package com.mesosphere.usi.persistence
```
meichstedt: For the sake of resolving the related JIRA, I'd go with the core modules only, and add persistence as a follow-up and its own JIRA. There's this fancy persistence impl from @zen-dog in the Marathon project and I wonder whether we should add any dummy impl here now, if we can as well move that one over, with benchmarks and all …

@zen-dog @timcharper thoughts?

persistence/src/main/scala/com/mesosphere/usi/persistence/Repository.scala
meichstedt: Should we _define_ this interface in this PR at all?

meichstedt (changes requested):
I appreciate your efforts @meln1k but I suggest to keep this PR short – i.e. leave out the persistence trait and impls, as well ad models we won't use for now (reservation and health checks). The task @timcharper is working on will likely result in changes to these models.

The JIRA only requires a module being setup, and I'd like to re-use the persistence impl @zen-dog already did in the Marathon project.

core-models/src/main/scala/com/mesosphere/usi/core/models/PodRecord.scala
jeschkies: That should be possible with https://plugins.gradle.org/plugin/com.github.hierynomus.scalariform

persistence/src/main/scala/com/mesosphere/usi/persistence/Repository.scala
meln1k: nope it's not FDB, I misunderstood Tim's idea. I'll refactor that.

persistence/src/main/scala/com/mesosphere/usi/persistence/Repository.scala
meln1k: @zen-dog can easily use his storage module as the underlying implementation

persistence/src/main/scala/com/mesosphere/usi/persistence/Repository.scala
meln1k: Sounds good, but we need to write it first.

core-models/src/main/scala/com/mesosphere/usi/core/models/PodRecord.scala
meln1k: Let's implement it later when there are more agreement around this. I'll remove it for now.

core-models/src/main/scala/com/mesosphere/usi/core/models/Ids.scala
```diff
@@ -0,0 +1,7 @@
+package com.mesosphere.usi.core.models
+
+
+case class PodId(value: String)
+case class ReservationId(value: String)
+case class AgentId(value: String)
+case class DeploymentId(value: String)
```
alenkacz, line 7: Will there be any notions of deployments in the USI? I though that it not a case...

persistence/src/main/scala/com/mesosphere/usi/persistence/repository/Repository.scala
```diff
@@ -0,0 +1,41 @@
...
+
+import com.mesosphere.usi.core.models.{PodId, PodRecord, ReservationId, ReservationRecord}
+
+import scala.concurrent.Future
+
+/**
+  * Base trait for storing records which are used by the Scheduler.
+  *
+  */
+trait Repository {
```
alenkacz, line 11: should there maybe be a separate repository for pod records and reservation records? I always considered repository pattern as being per entity...

alenkacz (changes requested):
I would personally have two repositories for two types of entities. But I am open to discussing that...

meln1k:
I'll create separate PR's for persistence and core-models

PR #13, opened 2019-01-26
Add `metrics` and `metrics-dropwizard` modules to the project

zen-dog:
Summary:
Ported Marathon's metrics interface and it's Dropwizard implementation:
- `metrics` contains a common interface for metrics implementation. It defines a common `Metrics` interface as well as a set of interfaces for common metrics primitives like `Counter`, `Gauge` etc.
- `metrics-dropwizard` contains Dropwizard metrics implementation
- Additionally ported Marathon's metrics reporter: StatsD, DataDog (API and UDP) and Prometheus

JIRA: DCOS-47507

Co-authored-by: Ivan Chernetsky <ichernetsky@mesosphere.io>

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -26,15 +26,25 @@ allprojects {
     }
 
     dependencies {
+        // Scala version for all scala dependencies
+        def scalaVersion = '2.12'
+
         // Use Scala 2.12 in our library project
         implementation 'org.scala-lang:scala-library:2.12.7'
 
         // Use Scalatest for testing our library
-        testImplementation 'junit:junit:4.12'
-        testImplementation 'org.scalatest:scalatest_2.12:3.0.5'
+        testImplementation "junit:junit:4.12"
+        testImplementation "org.scalatest:scalatest_$scalaVersion:3.0.5"
 
         // Need scala-xml at test runtime
-        testRuntimeOnly 'org.scala-lang.modules:scala-xml_2.12:1.1.1'
+        testRuntimeOnly "org.scala-lang.modules:scala-xml_$scalaVersion:1.1.1"
+
+        // Logging
+        compile group: 'ch.qos.logback', name: 'logback-classic', version: '1.2.3'
+        compile group: 'com.typesafe.scala-logging', name: "scala-logging_$scalaVersion", version: '3.9.2'
+
+        // Akka
+        compile group: 'com.typesafe.akka', name: "akka-http_$scalaVersion", version: '10.0.11'
     }
 
     test {
--- a/build.gradle
+++ b/build.gradle
@@ -11,6 +11,7 @@ allprojects {
     version = '0.1'
     apply plugin: "scala"
 
+    sourceCompatibility = 1.8
     targetCompatibility = 1.8
     compileScala.targetCompatibility = 1.8
 
--- a/build.gradle
+++ b/build.gradle
@@ -33,7 +33,7 @@ allprojects {
         implementation 'org.scala-lang:scala-library:2.12.7'
 
         // Use Scalatest for testing our library
-        testImplementation "junit:junit:4.12"
+        testImplementation 'junit:junit:4.12'
         testImplementation "org.scalatest:scalatest_$scalaVersion:3.0.5"
 
         // Need scala-xml at test runtime
--- a/build.gradle
+++ b/build.gradle
@@ -40,7 +40,6 @@ allprojects {
         testRuntimeOnly "org.scala-lang.modules:scala-xml_$scalaVersion:1.1.1"
 
         // Logging
-        compile group: 'ch.qos.logback', name: 'logback-classic', version: '1.2.3'
         compile group: 'com.typesafe.scala-logging', name: "scala-logging_$scalaVersion", version: '3.9.2'
 
         // Akka
--- a/metrics-dropwizard/README.md
+++ b/metrics-dropwizard/README.md
@@ -1,0 +1,140 @@
+# Metrics
+
+This is a [Dropwizard Metrics](https://github.com/dropwizard/metrics) implementation 
+of the USI metrics interface. It supports multiple exporters ([DataDog(https://www.datadoghq.com/)],
+[StatsD](https://github.com/etsy/statsd)) as well as text representation in 
+[Prometheus](https://prometheus.io/docs/instrumenting/exposition_formats/) format.
+
+Histograms and timers are backed with reservoirs leveraging
+[HdrHistogram](http://hdrhistogram.org/).
+
+## Metric names
+
+All metric names are prefixed with `usi` by default. This can be changed
+using `usi.metrics.name-prefix` configuration parameter. Metric name components 
+are joined with dots. Components may have dashes in them.
+
+A metric type and a unit of measurement (if any) are appended to
+a metric name. A couple of examples:
+
+* `usi.scheduler.offers.counter`
+* `usi.scheduler.offers.processing-time.seconds`
+
+## Prometheus reporter
+
+The Prometheus reporter can be used to return a text representation of 
+current metrics and their values. Dots and dashes in metric names are 
+replaced with underscores.
+
+## StatsD reporter
+
+The StatsD reporter can be enabled with `usi.metrics.statsd-reporter` parameter. 
+It sends metrics over UDP to the host and port specified with
+`usi.metrics.reporters.statsd.host` and `usi.metrics.reporters.statsd.port` respectively.
+
+Counters are forwarded as gauges.
+
+## DataDog reporter
+
+The DataDog reporter can be enabled with `usi.metrics.datadog-reporter` parameter. 
+It sends metrics over UDP to the host and port specified with `usi.metrics.reporters.datadog.host` 
+and `usi.metrics.reporters.datadog.port` respectively.
+
+Metrics can be sent to a DataDog agent over UDP, or directly to
+the DataDog cloud over HTTP. It is specified using
+`usi.metrics.reporters.datadog.protocol`. Its possible values are `udp` (default)
+and `api`. If `api` is chosen, your DataDog API key can be supplied with
+`usi.metrics.reporters.datadog.api-key`.
+
+Dashes in metric names are replaced with underscores. Counters are
+forwarded as gauges.
+
+
+### JVM-specific metrics
+A selected set of JVM related metrics is collected by default.
+
+#### JVM buffer pools
+
+* `usi.jvm.buffers.mapped.gauge` — an estimate of the number of
+  mapped buffers.
+* `usi.jvm.buffers.mapped.capacity.gauge.bytes` — an estimate of
+  the total capacity of the mapped buffers in bytes.
+* `usi.jvm.buffers.mapped.memory.used.gauge.bytes` an estimate of
+  the memory that the JVM is using for mapped buffers in bytes, or `-1L`
+  if an estimate of the memory usage is not available.
+* `usi.jvm.buffers.direct.gauge` — an estimate of the number of
+  direct buffers.
+* `usi.jvm.buffers.direct.capacity.gauge.bytes` — an estimate of
+  the total capacity of the direct buffers in bytes.
+* `usi.jvm.buffers.direct.memory.used.gauge.bytes` an estimate of
+  the memory that the JVM is using for direct buffers in bytes, or `-1L`
+  if an estimate of the memory usage is not available.
+
+#### JVM garbage collection
+
+* `usi.jvm.gc.<gc>.collections.gauge` — the total number
+  of collections that have occurred
+* `usi.jvm.gc.<gc>.collections.duraration.gauge.seconds` — the
+  approximate accumulated collection elapsed time, or `-1` if the
+  collection elapsed time is undefined for the given collector.
+
+#### JVM memory
+
+* `usi.jvm.memory.total.init.gauge.bytes` - the amount of memory
+  in bytes that the JVM initially requests from the operating system
+  for memory management, or `-1` if the initial memory size is
+  undefined.
+* `usi.jvm.memory.total.used.gauge.bytes` - the amount of used
+  memory in bytes.
+* `usi.jvm.memory.total.max.gauge.bytes` - the maximum amount of
+  memory in bytes that can be used for memory management, `-1` if the
+  maximum memory size is undefined.
+* `usi.jvm.memory.total.committed.gauge.bytes` - the amount of
+  memory in bytes that is committed for the JVM to use.
+* `usi.jvm.memory.heap.init.gauge.bytes` - the amount of heap
+  memory in bytes that the JVM initially requests from the operating
+  system for memory management, or `-1` if the initial memory size is
+  undefined.
+* `usi.jvm.memory.heap.used.gauge.bytes` - the amount of used heap
+  memory in bytes.
+* `usi.jvm.memory.heap.max.gauge.bytes` - the maximum amount of
+  heap memory in bytes that can be used for memory management, `-1` if
+  the maximum memory size is undefined.
+* `usi.jvm.memory.heap.committed.gauge.bytes` - the amount of heap
+  memory in bytes that is committed for the JVM to use.
+* `usi.jvm.memory.heap.usage.gauge` - the ratio of
+  `usi.jvm.memory.heap.used.gauge.bytes` and
+  `usi.jvm.memory.heap.max.gauge.bytes`.
+* `usi.jvm.memory.non-heap.init.gauge.bytes` - the amount of
+  non-heap memory in bytes that the JVM initially requests from the
+  operating system for memory management, or `-1` if the initial memory
+  size is undefined.
+* `usi.jvm.memory.non-heap.used.gauge.bytes` - the amount of used
+  non-heap memory in bytes.
+* `usi.jvm.memory.non-heap.max.gauge.bytes` - the maximum amount of
+  non-heap memory in bytes that can be used for memory management, `-1`
+  if the maximum memory size is undefined.
+* `usi.jvm.memory.non-heap.committed.gauge.bytes` - the amount of
+  non-heap memory in bytes that is committed for the JVM to use.
+* `usi.jvm.memory.non-heap.usage.gauge` - the ratio of
+  `usi.jvm.memory.non-heap.used.gauge.bytes` and
+  `usi.jvm.memory.non-heap.max.gauge.bytes`.
+
+#### JVM threads
+
+* `usi.jvm.threads.active.gauge` — the number of active threads.
+* `usi.jvm.threads.daemon.gauge` — the number of daemon threads.
+* `usi.jvm.threads.deadlocked.gauge` — the number of deadlocked
+  threads.
+* `usi.jvm.threads.new.gauge` — the number of threads in `NEW`
+   state.
+* `usi.jvm.threads.runnable.gauge` — the number of threads in
+  `RUNNABLE` state.
+* `usi.jvm.threads.blocked.gauge` — the number of threads in
+  `BLOCKED` state.
+* `usi.jvm.threads.timed-waiting.gauge` — the number of threads in
+  `TIMED_WAITING` state.
+* `usi.jvm.threads.waiting.gauge` — the number of threads in
+  `WAITING` state.
+* `usi.jvm.threads.terminated.gauge` — the number of threads in
+  `TERMINATED` state.
--- a/metrics-dropwizard/build.gradle
+++ b/metrics-dropwizard/build.gradle
@@ -1,22 +1,8 @@
-plugins {
-    id 'scala'
-}
-
-repositories {
-    mavenCentral()
-}
-
 dependencies {
-    implementation project(':usi-metrics')
+    implementation project(':metrics')
+    compile group: 'com.github.vladimir-bukhtoyarov', name: 'rolling-metrics', version:  '2.0.4'
+    compile group: 'com.iheart', name: 'ficus_2.12', version: '1.4.3'
     compile group: 'io.dropwizard.metrics', name: 'metrics-core', version: '4.0.2'
     compile group: 'io.dropwizard.metrics', name: 'metrics-jvm', version: '4.0.2'
-    compile group: 'com.github.vladimir-bukhtoyarov', name: 'rolling-metrics', version:  '2.0.4'
     compile group: 'org.hdrhistogram', name: 'HdrHistogram', version:  '2.1.10'
-
-    // Common project dependencies
-    compile group: 'org.rogach', name: 'scallop_2.12', version: '3.1.2'
-    compile group: 'ch.qos.logback', name: 'logback-classic', version: '1.2.3'
-    compile group: 'com.typesafe.scala-logging', name: 'scala-logging_2.12', version: '3.9.2'
-    compile group: 'com.typesafe.akka', name: 'akka-http_2.12', version: '10.0.11'
-    compile group: 'com.iheart', name: 'ficus_2.12', version: '1.4.3'
 }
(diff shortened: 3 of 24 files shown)


metrics-dropwizard/src/main/scala/com/mesosphere/usi/metrics/dropwizard/DropwizardMetricsModule.scala
```diff
@@ -0,0 +1,41 @@
...
+import akka.actor.ActorRefFactory
+import com.codahale.metrics.MetricRegistry
+import com.mesosphere.usi.metrics.Metrics
+import com.mesosphere.usi.metrics.dropwizard.conf.{DataDogApiReporterSettings, DataDogUdpReporterSettings, MetricsSettings}
+import com.mesosphere.usi.metrics.dropwizard.reporters.{DataDogAPIReporter, DataDogUDPReporter, StatsDReporter}
+import com.typesafe.config.Config
+
+class DropwizardMetricsModule(config: Config) {
+
+  val metricsSettings = MetricsSettings.fromSubConfig(config)
```
zen-dog, line 15: We should discuss this pattern of configuring modules and components throughout the project. This is heavily influenced by the [typesafe recommendations](https://github.com/lightbend/config#schemas-and-validation) and by some existing implementation e.g. [spray](https://github.com/spray/spray/blob/master/spray-can/src/main/scala/spray/can/server/ServerSettings.scala). In a nutshell:
- modules/components are initialized with the part of the `Config` that is relevant for them, e.g in this case `MetricsModule` gets the `usi.metrics` part of the config tree. This way, we can rearrange parts of the configuration tree if needed
- `Config` is parsed into the `Settings` class. All the other parts of the module only know the `Settings` class. This is also where validation happens (if needed). See [MetricsSettings](https://github.com/mesosphere/usi/pull/13/files#diff-b26429af3946938b3f805558d1fce910) class below.

settings.gradle
```diff
@@ -2,6 +2,10 @@ rootProject.name = 'usi'
 
 include 'interfaces'
 include 'core'
+include 'metrics'
+findProject(':metrics')?.name = 'usi-metrics'
+include 'metrics-dropwizard'
+findProject(':metrics-dropwizard')?.name = 'usi-metrics-dropwizard'
```
jeschkies: What's the output of `./gradlew projects`? What do you think about setting the project name in `subprojects`?

metrics-dropwizard/build.gradle
```diff
@@ -0,0 +1,22 @@
+plugins {
+    id 'scala'
+}
+
+repositories {
+    mavenCentral()
```
jeschkies: Why not `jcenter()`? Should we add `mavenCentral()` for all subprojects?

metrics-dropwizard/build.gradle
```diff
@@ -0,0 +1,22 @@
...
+}
+
+dependencies {
+    implementation project(':usi-metrics')
+    compile group: 'io.dropwizard.metrics', name: 'metrics-core', version: '4.0.2'
+    compile group: 'io.dropwizard.metrics', name: 'metrics-jvm', version: '4.0.2'
```
jeschkies: What do you think about sorting alphabetical: `com.github...`, `io.dropwizard...`?

metrics-dropwizard/build.gradle
```diff
@@ -0,0 +1,22 @@
...
+    compile group: 'com.github.vladimir-bukhtoyarov', name: 'rolling-metrics', version:  '2.0.4'
+    compile group: 'org.hdrhistogram', name: 'HdrHistogram', version:  '2.1.10'
+
+    // Common project dependencies
```
jeschkies: What do you mean?

metrics-dropwizard/src/main/java/com/mesosphere/usi/metrics/dropwizard/GarbageCollectorMetricSet.java
```diff
@@ -0,0 +1,52 @@
...
+
+import static com.codahale.metrics.MetricRegistry.name;
+
+/**
+ * A set of gauges for the counts and elapsed times of garbage collections.
+ *
+ * This is a copy of GarbageCollectorMetricSet from Dropwizard metrics, slightly adjusted to support USI metric name
+ * conventions.
+ */
+public class GarbageCollectorMetricSet implements MetricSet {
```
jeschkies, line 27: Is this from Ivan? Why is it in Java?

metrics-dropwizard/src/main/resources/reference.conf
```diff
@@ -0,0 +1,63 @@
+usi.metrics {
```
jeschkies, line 1: Shouldn't this be in `metrics` instead of `metrics-dropwizard`?

metrics-dropwizard/src/main/resources/reference.conf
```diff
@@ -0,0 +1,63 @@
...
+
+  # A prefix that is used when constructing metric names
+  name-prefix: "usi"
+
+  # Histogram realated settings
+  histogram {
+    # The highest trackable value of histograms and timers
+    reservoir-highest-trackable-value: 3600000000000
```
jeschkies, line 9: What is the unit?

metrics-dropwizard/src/main/resources/reference.conf
```diff
@@ -0,0 +1,63 @@
...
+  }
+
+  # Enable StatsD reporter. If enabled, the `statsd-reporter` block should be enabled and
+  # host and port should be provided
+  statsd-reporter: off
+
+  # StatsD reporter
+  # reporters.statsd {
+  #   # A remote hostname for the StatsD reporter
+  #   host:
```
jeschkies, line 31: I like the solution of commenting this block since there is probably not a reasonable default.

metrics-dropwizard/src/main/resources/reference.conf
```diff
@@ -0,0 +1,63 @@
...
+  #    port: 0
+  #
+  #    # A protocol to use with the DataDog reporter (default: udp; supported protocols: udp, api).
+  #    # When using "api" protocol, `api-key` should be provided. With "udp" protocol  host and port
+  #    # should be set.
+  #
+  #    protocol: "api"
+  #
+  #    # A DataDog API key
+  #    api-key: "FAKE-API-KEY"
```
jeschkies: This should be an environment variable. We should not encourage users to set a key in any text file unless it's encrypted such as credentials in Travis CI files.

metrics-dropwizard/src/main/scala/com/mesosphere/usi/metrics/dropwizard/conf/MetricsSettings.scala
```diff
@@ -0,0 +1,100 @@
...
+
+import scala.concurrent.duration.Duration
+import scala.concurrent.duration._
+import net.ceedubs.ficus.Ficus.{finiteDurationReader, toFicusConfig}
+
+case class HistorgramSettings(reservoirHighestTrackableValue: Long = 3600000000000L,
+                              reservoirSignificantDigits: Int = 3,
+                              reservoirResetPeriodically: Boolean = true,
+                              reservoirResettingInterval: FiniteDuration = 5.seconds,
+                              reservoirResettingChunks: Int = 0
```
jeschkies, line 13: Again the formatting. I guess we should get scalariform working.

metrics-dropwizard/src/main/scala/com/mesosphere/usi/metrics/dropwizard/conf/MetricsSettings.scala
```diff
@@ -0,0 +1,100 @@
...
+                             ) {
+  require(reservoirHighestTrackableValue > 1L, "reservoir-highest-trackable-value: should be > 1")
+  require(reservoirSignificantDigits >= 0 && reservoirSignificantDigits <= 5, "reservoir-significant-digits should be >= 0 and <= 5")
+  require(reservoirResettingInterval > Duration.Zero, "reservoir-resetting-interval should be > 0")
+  require(reservoirResettingChunks == 0 || reservoirResettingChunks >= 2, "reservoir-resetting-chunks should be == 0 or >= 2")
+}
+
+object HistorgramSettings {
+  def fromSubConfig(c: Config) = apply (
+    c.getLong("reservoir-highest-trackable-value"),
```
jeschkies, line 23: What happens when one value is `null`? Should we validate first? Ah, you do so in the constructor...

metrics/README.md
```diff
@@ -0,0 +1,163 @@
+# Metrics
+
+This is a [Dropwizard Metrics](https://github.com/dropwizard/metrics) implementation
```
jeschkies: ...yet the `README.md` is in `metrics`❓ I wonder whether this should be part of `./docs` or even the Scaladoc for the package `usi-metrics`:

```scala
package com.mesosphere.usi

/**
 * ...
 * == Metrics Types ==
 */
package metrics {}
```

metrics/build.gradle
```diff
@@ -0,0 +1,12 @@
+plugins {
+    id 'scala'
+}
+
+repositories {
+    mavenCentral()
```
jeschkies: See my comment above.

metrics/src/main/scala/com/mesosphere/usi/metrics/Metrics.scala
```diff
@@ -0,0 +1,59 @@
+package com.mesosphere.usi.metrics
+
+import java.time.Clock
+import akka.stream.scaladsl.Source
+import scala.concurrent.Future
+
+trait Metric
+
+trait Counter extends Metric {
```
jeschkies: Please add documentation.

jeschkies (changes requested):
Thanks. I did not review the code much as I assume that happened in Marathon.

metrics-dropwizard/build.gradle
alenkacz: the jcenter is already defined in the base build.gradle for all buildprojects. I would stick with that or add maven central if necessary in there... As Karsten proposes :)

metrics-dropwizard/build.gradle
```diff
@@ -0,0 +1,22 @@
+plugins {
+    id 'scala'
```
alenkacz: why not `apply plugin scala`? Also this is also defined in the base build.gradle for all projects so we should drop it from here as it's redundant

jeschkies: This is the new way. `apply plugin scala` is the deprecated way.

zen-dog: What he said ^^

metrics/README.md
zen-dog: I think every sub-project should have its own README (which can also be part of the docs). I expect that much when browsing subprojects in GitHub

metrics-dropwizard/src/main/java/com/mesosphere/usi/metrics/dropwizard/GarbageCollectorMetricSet.java
zen-dog: As the comment there says: 
```
 This is a copy of GarbageCollectorMetricSet from Dropwizard metrics, slightly adjusted to support USI metric name conventions.
```
Same with other collectors. And the reporters exits for the same reason - it's impossible to find one that doesn't screw up with metric names.

metrics-dropwizard/build.gradle
zen-dog: Oh, thanks for the reminder. I wanted to move them out as dependencies for all projects (logging, akka etc. will be used everywhere)

metrics-dropwizard/build.gradle
zen-dog: alphabetical by the group name? artifact name?

settings.gradle
zen-dog:
```
$ gradle projects
Root project 'usi'
+--- Project ':core'
+--- Project ':hello-world'
+--- Project ':interfaces'
+--- Project ':usi-metrics'
\--- Project ':usi-metrics-dropwizard'
```

metrics-dropwizard/build.gradle
jeschkies: First group name then artifact name.

settings.gradle
zen-dog: I removed it completely

metrics-dropwizard/src/main/resources/reference.conf
zen-dog: It's `Long` but if you put a unit there, it'll be interpreted as a string.

metrics-dropwizard/src/main/resources/reference.conf
zen-dog: Yep, everything else doesn't seem to work with these type of settings

metrics/README.md
jeschkies: Hm, I'm not sure. I understand the idea but I find browsing READMEs in subdirectories a strange practice. A single point of entry for docs is much nicer. Also for us who will have to maintain all the docs and READMEs. Having a `./README.md`, `metrics/README.md`, `./dcos/metrics.md` and Scaladoc for `com.mesosphere.usi.metrics` seems too much. Where should we document what?

build.gradle
```diff
@@ -20,15 +20,25 @@ allprojects {
...
+        // Scala version for all scala dependencies
+        def scalaVersion = '2.12'
+
         // Use Scala 2.12 in our library project
         implementation 'org.scala-lang:scala-library:2.12.7'
 
         // Use Scalatest for testing our library
-        testImplementation 'junit:junit:4.12'
-        testImplementation 'org.scalatest:scalatest_2.12:3.0.5'
+        testImplementation "junit:junit:4.12"
```
jeschkies: What's the idea for this change? Logging still has single-quotes.

build.gradle
```diff
@@ -20,15 +20,25 @@ allprojects {
...
+        testImplementation "org.scalatest:scalatest_$scalaVersion:3.0.5"
 
         // Need scala-xml at test runtime
-        testRuntimeOnly 'org.scala-lang.modules:scala-xml_2.12:1.1.1'
+        testRuntimeOnly "org.scala-lang.modules:scala-xml_$scalaVersion:1.1.1"
+
+        // Logging
+        compile group: 'ch.qos.logback', name: 'logback-classic', version: '1.2.3'
```
jeschkies: This should go. Libraries should only depend on SL4J or similar interfaces. Scalalogging uses SL4J so that's fine. If we need a backend for tests or examples they should define it as a runtime dependency.

jeschkies (changes requested):

metrics-dropwizard/src/main/resources/reference.conf
zen-dog: Good catch. It was and then I replaced it with this fake string for tests.

meln1k (approved):

metrics-dropwizard/build.gradle
zen-dog: Have fun enforcing this rule across the project :D I'll do that but just because it's you!

build.gradle
jeschkies: It’s fine to declare it as a test runtime dependency.

metrics/README.md
zen-dog: As discussed offline - let's move forwards with READMEs for now and have a bigger discussion on docs structure, templates, content etc.

metrics-dropwizard/src/main/resources/reference.conf
zen-dog: I'd say everything in this conf is implementation specific: reporters, histogram settings (based by HdrHistogram) and even name prefix - original interface doesn't define any of those (for a good reason)

zen-dog: Think about `DummyMetris` used for tests that don't need any of these settings

metrics-dropwizard/README.md
```diff
@@ -0,0 +1,140 @@
...
+## Metric names
+
+All metric names are prefixed with `usi` by default. This can be changed
+using `usi.metrics.name-prefix` configuration parameter. Metric name components 
+are joined with dots. Components may have dashes in them.
+
+A metric type and a unit of measurement (if any) are appended to
+a metric name. A couple of examples:
+
+* `usi.scheduler.offers.counter`
```
kensipe, line 20: initially I thought this was a huge mistake... but it looks like the `name-prefix` is a way around my concern.   I do NOT think it is a good idea to have a non-initialized default for this.     all metrics are dumped into a common repo (Prometheus etc).   Even having 2 schedulers using USI pushing to metrics is bad IMO.    It should absolutely be a requirement... in order to use USI you have to initialize with a scheduler name which is a prefix potential to a number of things but certainly to metrics.

kensipe: this would be a blocker for me if we didn't have a way to configure the `name-prefix`.   At this point it isn't a blocker but I would like to make the ability to use USI impossible without initializing the scheduler name etc.

metrics/src/main/scala/com/mesosphere/usi/metrics/Metrics.scala
```diff
@@ -0,0 +1,76 @@
+package com.mesosphere.usi.metrics
```
kensipe, line 1: this is exactly what I didn't want...    why can we not have a `usi-commons` which has the common interfaces and models instead of having a individual project for everything we want to make extensible?

metrics/README.md
```diff
@@ -0,0 +1,28 @@
+# Metrics
+
+This module defines a set of metrics interfaces. For a concrete implementation 
+see [metrics-dropwizar](http://github.com/mesosphere/use/metrics-dropwizard)
```
kensipe: `[metrics-dropwizar]` -> `[metrics-dropwizard]`

kensipe: `http://github.com/mesosphere/use/metrics-dropwizard` -> `http://github.com/mesosphere/usi/metrics-dropwizard`

metrics-dropwizard/build.gradle
```diff
@@ -0,0 +1,8 @@
+dependencies {
+    implementation project(':metrics')
+    compile group: 'com.github.vladimir-bukhtoyarov', name: 'rolling-metrics', version:  '2.0.4'
+    compile group: 'com.iheart', name: 'ficus_2.12', version: '1.4.3'
+    compile group: 'io.dropwizard.metrics', name: 'metrics-core', version: '4.0.2'
+    compile group: 'io.dropwizard.metrics', name: 'metrics-jvm', version: '4.0.2'
```
kensipe: fails to compile for Java 11... seems like a blocker

```
FAILURE: Build failed with an exception.

* What went wrong:
Execution failed for task ':metrics-dropwizard:compileJava'.
> warning: source release 11 requires target release 11

```

metrics-dropwizard/src/main/java/com/mesosphere/usi/metrics/dropwizard/MemoryUsageGaugeSet.java
```diff
@@ -0,0 +1,106 @@
...
+import com.codahale.metrics.Gauge;
+import com.codahale.metrics.Metric;
+import com.codahale.metrics.MetricSet;
+import com.codahale.metrics.RatioGauge;
+
+import java.lang.management.ManagementFactory;
+import java.lang.management.MemoryMXBean;
+import java.lang.management.MemoryPoolMXBean;
+import java.lang.management.MemoryUsage;
+import java.util.*;
```
kensipe: our standard is no `*`

metrics-dropwizard/src/main/java/com/mesosphere/usi/metrics/dropwizard/GarbageCollectorMetricSet.java
```diff
@@ -0,0 +1,52 @@
+package com.mesosphere.usi.metrics.dropwizard;
+
+import com.codahale.metrics.Gauge;
+import com.codahale.metrics.Metric;
+import com.codahale.metrics.MetricSet;
+
+import java.lang.management.GarbageCollectorMXBean;
+import java.lang.management.ManagementFactory;
+import java.util.*;
```
kensipe: no `*`

kensipe (changes requested):
Not compiling on Java 11 is a blocker to me... regardless of the fact that we want to stay with bytecode to a lesser version.

I also strongly dislike the creation of another interface project that has 2 class files in it.

metrics-dropwizard/src/main/java/com/mesosphere/usi/metrics/dropwizard/GarbageCollectorMetricSet.java
```diff
@@ -0,0 +1,52 @@
...
+    }
+
+    @Override
+    public Map<String, Metric> getMetrics() {
+        final Map<String, Metric> gauges = new HashMap<>();
+        for (final GarbageCollectorMXBean gc : garbageCollectors) {
+            final String name = WHITESPACE.matcher(gc.getName()).replaceAll("-").toLowerCase(Locale.US);
+            gauges.put(name(name, "collections.gauge"), (Gauge<Long>) gc::getCollectionCount);
+            gauges.put(name(name, "collections.duration.gauge.seconds"),
+                    (Gauge<Double>) () -> (new Long(gc.getCollectionTime()).doubleValue()) / 1000.0);
```
kensipe: this is a crazy line of code... 
1. get double value from `gc.getCollectionTime()`
2. convert that double to a `Long`
3. get the `doubleValue` of that long value that was a double
4. divide by a float to be autoboxed into a `Double` part of `Gauge<Double>`

kensipe (approved):
Resolved all blocking issues I had... updating to approve of PR.

* upgrading to latest dependencies for dropwizard 
* enable java 11 compiles (to java 8 bytecode)
* expanded * imports
* removing redundant casting of types for gc metrics. 
* added gradle linter to expand use of deprecate warnings

build.gradle
```diff
@@ -20,20 +21,36 @@ allprojects {
...
+        // Scala version for all scala dependencies
+        def scalaVersion = '2.12'
+
         // Use Scala 2.12 in our library project
         implementation 'org.scala-lang:scala-library:2.12.7'
 
         // Use Scalatest for testing our library
         testImplementation 'junit:junit:4.12'
-        testImplementation 'org.scalatest:scalatest_2.12:3.0.5'
+        testImplementation "org.scalatest:scalatest_$scalaVersion:3.0.5"
```
jeschkies, line 38: Why dobule quotes?

metrics-dropwizard/src/main/scala/com/mesosphere/usi/metrics/dropwizard/DropwizardMetricsModule.scala
jeschkies: Shouldn't `config` be of type `MetricsSettings` or are we doing it here so that the caller does not have to care about what the actual module needs?

metrics/src/main/scala/com/mesosphere/usi/metrics/Metrics.scala
jeschkies: Hehe, that is what I called `interfaces` but was overturned.

jeschkies (approved):
I accept but we really have to agree on
* common way of defining settings.
* pattern for READMEs

build.gradle
zen-dog: For `$scalaVersion` substitution to work

metrics/src/main/scala/com/mesosphere/usi/metrics/Metrics.scala
zen-dog: Yep, that's the whole "interfaces" debate we had last week.

PR #14, opened 2019-01-28
Mesos V1 initial implementation

alenkacz:
Most of the code is copy-paste from Marathon repo. Notable changes:
- using java POJOs from the `org.apache.mesos:mesos` jar instead of generating case classes from the protobufs manually
- using typesafe config for configuration

Co-authored-by: Aleksey Dukhovniy <alex.dukhovniy@googlemail.com>


mesos-client/src/main/scala/mesosphere/mesos/client/MesosCalls.scala
```diff
@@ -0,0 +1,265 @@
+package mesosphere.mesos.client
+
+import akka.util.ByteString
+import com.google.protobuf
+import org.apache.mesos.v1.Protos.{AgentID, ExecutorID, FrameworkID, Filters, KillPolicy, OfferID, Request, TaskID}
+import org.apache.mesos.v1.scheduler.Protos.Call.{Accept, Acknowledge, Decline, Kill, Message, Reconcile, Revive}
+import org.apache.mesos.v1.scheduler.Protos.{Call, Event}
+import collection.JavaConverters._
+
+class MesosCalls(frameworkId: FrameworkID) {
```
alenkacz: most of the changes happened in this class

mesos-client/src/main/scala/mesosphere/mesos/client/MesosClient.scala
```diff
@@ -0,0 +1,378 @@
...
+    * `subscribe.framework_info.id` is not set, master considers the scheduler as a new one and subscribes it by
+    * assigning it a FrameworkID. The HTTP response is a stream in RecordIO format; the event stream begins with a
+    * SUBSCRIBED event.
+    *
+    * Note: this method is used by mesos client to establish connection to mesos master and is not supposed to be called
+    * directly by the framework.
+    *
+    * http://mesos.apache.org/documentation/latest/scheduler-http-api/#subscribe-1
+    */
+  private def newSubscribeCall(frameworkInfo: FrameworkInfo): Call = {
```
alenkacz: change here

meln1k:
What do you think about using the generated protobufs from https://mvnrepository.com/artifact/org.apache.mesos/mesos/1.7.1 instead of having .proto in the repository?

meln1k (approved):

mesos-client/src/main/scala/mesosphere/mesos/conf/MesosClientConf.scala
```diff
@@ -0,0 +1,18 @@
...
+/**
+  * Configuring mesos v1 client
+  *
+  * @param master The URL of the Mesos master
+  * @param sourceBufferSize Buffer size of the mesos source im messages
+  * @param redirectRetries Number of retries to follow mesos master redirect
+  * @param idleTimeout Time in seconds between two processed elements exceeds the provided timeout then the connection to mesos
+  *                    is interrupted. Is usually set to approx. 5 hear beats.
+  */
+case class MesosClientConf(
```
zen-dog: Should we apply the same configuration patter here as in [metrics](https://github.com/mesosphere/usi/pull/13/files#r251202266) e.g. using Typesafe config and this renamed to `MesosClientSettings`?

mesos-client/src/main/scala/mesosphere/mesos/examples/UselessFramework.scala
```diff
@@ -0,0 +1,65 @@
...
+
+/**
+  * Run Foo framework that:
+  *  - successfully subscribes
+  *  - declines all offers.
+  *
+  *  Not much, but shows the basic idea. Good to test against local mesos.
+  *
+  */
+object UselessFramework extends App with StrictLoggingFlow {
```
kensipe: we have a project for examples... should we move examples to that ?

mesos-client/README.md
```diff
@@ -0,0 +1,128 @@
+# Mesos V1 Client
```
jeschkies, line 1: I had a discussion with @zen-dog about READMEs in project folders. We should agree on a common pattern. IMHO most of the content here should be in `./docs` instead since it can be subject to change. The readme should link to these docs. What do you htink?

mesos-client/build.gradle
```diff
@@ -0,0 +1,16 @@
+plugins {
+    id 'application'
+}
+
+dependencies {
+    implementation 'com.typesafe.akka:akka-stream_2.12:2.5.19'
+    implementation 'org.apache.mesos:mesos:1.7.1'
```
jeschkies, line 2: Why not `implementation 'org.apache.mesos:mesos:1.7.1:shaded-protobuf'`? Do we have other dependencies?

mesos-client/build.gradle
```diff
@@ -0,0 +1,16 @@
...
+    implementation 'com.typesafe.akka:akka-http_2.12:10.1.7'
+    implementation 'com.lightbend.akka:akka-stream-alpakka-simple-codecs_2.12:0.20'
+
+    runtime 'org.slf4j:slf4j-simple:1.7.25'
```
jeschkies: Why is this required? Shouldn't this be a `testRuntimeOnly` dependency?

mesos-client/build.gradle
```diff
@@ -0,0 +1,16 @@
+plugins {
+    id 'application'
+}
+
+dependencies {
+    implementation 'com.typesafe.akka:akka-stream_2.12:2.5.19'
+    implementation 'org.apache.mesos:mesos:1.7.1'
+    implementation 'com.typesafe.akka:akka-http_2.12:10.1.7'
+    implementation 'com.lightbend.akka:akka-stream-alpakka-simple-codecs_2.12:0.20'
```
jeschkies, line 3: What do you think about sorting the dependencies alphabetically?

mesos-client/build.gradle
```diff
@@ -0,0 +1,16 @@
...
+
+    runtime 'org.slf4j:slf4j-simple:1.7.25'
+}
+
+application {
+    mainClassName = "mesosphere.mesos.examples.UselessFramework"
```
jeschkies: This is surprising for a library.

kensipe (dismissed):
1. +1 to Alek comment on consistency with metrics regardign confs
2. I'm not sure why we would have an example dummy class built as part of our distro

nice work.. nice docs!!

mesos-client/src/main/scala/mesosphere/mesos/conf/MesosClientConf.scala
jeschkies: +1

mesos-client/src/main/scala/mesosphere/mesos/examples/UselessFramework.scala
jeschkies: Should this be moved to `./examples` instead?

jeschkies (changes requested):
I'd like to hear your opinion on some structure suggestions. I know this is painful at the beginning but I'd like to have consistency.

mesos-client/src/main/scala/mesosphere/mesos/examples/UselessFramework.scala
jeschkies: Man, I just posted the same. 😂 Good to know that we all converge to the same structure.

zen-dog: Here is why it could make sense: the `usi/examples` folder is for examples _of the whole thing_ e.g. full-fledged frameworks capable of starting tasks on Mesos (at least that is my understanding of it). This `mesos-client` module is however usable on its own - it can't start anything but it can connect to Mesos.

mesos-client/build.gradle
zen-dog: Yep, remnants of the old codebase.

mesos-client/src/main/scala/mesosphere/mesos/examples/UselessFramework.scala
alenkacz: Yeah, this lives in this repo just to ease the initial development. In the future, this could be a separate repo and separate library used on its own - that's how it was designed from the beginning. Originally we planned to plug it in Marathon (which is still open possibility).

So I think considering that, examples on this level make sense and I like that @zen-dog did it this way

mesos-client/README.md
alenkacz: Hmm I am not sure. I picture this piece as a standalone library at one point and then having it in readme makes sense. I guess I would probably like to have both but not sure how to do that - so that we don't have to have two copies

mesos-client/build.gradle
alenkacz: this is required for the main class to run (and to be able to log). I am fine dropping this but then everytime someone does `gradle run` on this, no logs will be printed

mesos-client/build.gradle
alenkacz: this is to run the example framework

mesos-client/build.gradle
alenkacz: when I do that, I get e.g. this build failure 🤔 

```MesosCalls.scala:166: type mismatch;
 found   : ByteString (in com.google.protobuf) 
 required: ByteString (in org.apache.mesos.protobuf) 
      .setAcknowledge(Call.Acknowledge.newBuilder().setAgentId(agentId).setTaskId(taskId).setUuid(uuid).build())```

**Comment by alenkacz** [comment]
@zen-dog pulling additional dependency into a library (that should ideally have as little dependencies as possible) feels like too much to me. Just because we don't want to use `toNanos` now...

File: mesos-client/README.md (hunk 5 above)
Comment (jeschkies) [reply mesos-client/README.md:1]: Yeah, I talked to @zen-dog and @meln1k and lost my arguments. We do have to define what should be in a sub-module README and what should be in the docs but this should be not be a debate for this PR.

File: mesos-client/build.gradle (hunk 6 above)
Comment (jeschkies) [reply mesos-client/build.gradle:2]: So the import is wrong? Not sure why this should be different.

File: mesos-client/build.gradle (hunk 7 above)
Comment (jeschkies) [reply mesos-client/build.gradle]: Why would one use `gradle run` on this library?

File: mesos-client/src/main/scala/mesosphere/mesos/examples/UselessFramework.scala (hunk 4 above)
Comment (jeschkies) [reply mesos-client/src/main/scala/mesosphere/mesos/examples/UselessFramework.scala]: I'd not ship it with the module though. how about making it a test target then? The Gradle project should not have a main.

**Review by jeschkies — CHANGES_REQUESTED** [review]
Sorry, I'm still opposed to shipping the `UselessFramework`. We could leave it out here and ship it with the test infrastructure @zen-dog works on.

File: mesos-client/src/main/scala/mesosphere/mesos/client/MesosCalls.scala [hunk 10]
```diff
@@ -0,0 +1,264 @@
+package mesosphere.mesos.client
```
jeschkies: I've forgot. We started using `com.mesosphere`.

jeschkies (approved):

PR #16, opened 2019-01-28
 Added a persistence project module. 

meln1k:
This PR is created on top of on https://github.com/mesosphere/usi/pull/17

JIRA issues: DCOS-47508


persistence/src/main/scala/com/mesosphere/usi/persistence/store/PersistenceStore.scala
```diff
@@ -0,0 +1,114 @@
...
+
+import scala.concurrent.Future
+import scala.util.Try
+
+/**
+  * An interface for a simple persistence store abstraction for the underlying Zookeeper store. Supported are basic
+  * create, read, update, delete and check methods along with transaction operations and Zookeeper specific children
+  * and sync methods.
+  */
+trait PersistenceStore {
```
zen-dog: No, no. This storage interface is a detail of the `persistence-zookeeper` module and has nothing to do here.

zen-dog (changes requested):
See my comment

jeschkies:
> This PR is created on top of on #17
How about using the branch of #17 as a base to make reviewing simpler? Maybe there is another way.

persistence/build.gradle
```diff
@@ -0,0 +1,3 @@
+dependencies {
+    compile project(':core-models')
```
zen-dog, line 2:
```suggestion
    api project(':core-models')
```

persistence/src/main/scala/com/mesosphere/usi/repository/PodRecordRepository.scala
```diff
@@ -0,0 +1,26 @@
+package com.mesosphere.usi.repository
+
+import com.mesosphere.usi.core.models.{PodId, PodRecord}
+
+import scala.concurrent.Future
+
+/**
+  * Repository that stores [[PodRecord]].
```
alenkacz: not sure if such scala docs makes sense but 🤷‍♀️

alenkacz (approved):
LGTM thanks!

persistence/src/main/scala/com/mesosphere/usi/repository/PodRecordRepository.scala
```diff
@@ -0,0 +1,26 @@
...
+  */
+trait PodRecordRepository {
+
+  /**
+    * Stores the provided [[PodRecord]] in the repository.
+    * @param record
+    * @return
+    */
+  def store(record: PodRecord): Future[Unit]
```
zen-dog: How about all four CRUD methods for the time being? And also, how about staying with "default names" e.g. `create`, `read` etc.?

persistence/build.gradle
jeschkies: Neat, I did not know that Gradle could do this.

persistence/src/main/scala/com/mesosphere/usi/repository/PodRecordRepository.scala
jeschkies: Yeah, why not use @zen-dog's storage?

meln1k: @jeschkies  it is the next step, @zen-dog is operating on a lower level

persistence/src/main/scala/com/mesosphere/usi/repository/PodRecordRepository.scala
```diff
@@ -0,0 +1,40 @@
...
+
+/**
+  * Repository for [[PodRecord]].
+  */
+trait PodRecordRepository {
+
+  /**
+    * Stores the provided [[PodRecord]] in the repository if it doesn't exist.
+    * @param record
+    * @return
```
jeschkies: What's returned when? This should not be a blocker now but we'll have to document it.

persistence/src/main/scala/com/mesosphere/usi/repository/PodRecordRepository.scala
```diff
@@ -0,0 +1,40 @@
...
+
+  /**
+    * Retrieves the [[PodRecord]] if it exists.
+    * @param podId
+    * @return Option(PodRecord) if it exists, None otherwise
+    */
+  def read(podId: PodId): Future[Option[PodRecord]]
+
+  /**
+    * Updates the [[PodRecord]] if it exists.
```
jeschkies: And what happens when it does not exist?

jeschkies (approved):
Given you change the dependency to `api` 🙂

zen-dog (approved):
We also need a short README (you can use `metrics` for inspiration). But otherwise lgtm

persistence/build.gradle
meln1k: It's only possible for java

PR #17, opened 2019-01-28
Add core-models module

meln1k:
JIRA issues: DCOS-47508


core-models/src/main/scala/com/mesosphere/usi/core/models/PodRecord.scala
```diff
@@ -0,0 +1,9 @@
+package com.mesosphere.usi.core.models
+
+import java.time.Instant
+
+/**
+  * Snapshot of the Pod Spec state that needs to be persisted and can't be obtained from mesos.
+  * @param podId id of the pod
+  */
+case class PodRecord(podId: PodId)
```
zen-dog, line 8: I guess we can always change how `PodRecord`s and co look like in detail later.

core-models/src/main/scala/com/mesosphere/usi/core/models/PodSpec.scala
```diff
@@ -0,0 +1,3 @@
+package com.mesosphere.usi.core.models
```
zen-dog, line 1: Wdyt about putting:
- `PodSpec` into `core.models.framework`
- `PodRecord` and `PodReservation` into `core.models.scheduler` 
to make it more obvious that they are used on different levels. I know, we're still vague on what `framework` and `scheduler` actually incapsulates but still.

zen-dog (approved):
See my comments but in general lgtm

core-models/src/main/scala/com/mesosphere/usi/core/models/PodSpec.scala
meln1k: Discussed offline, no longer needed.

core-models/build.gradle
```diff
@@ -0,0 +1,2 @@
+dependencies {
```
jeschkies: Is this required? I think we just need a `build.gradle`.

core-models/src/main/scala/com/mesosphere/usi/core/models/Ids.scala
```diff
@@ -0,0 +1,5 @@
+package com.mesosphere.usi.core.models
+
+case class PodId(value: String)
```
jeschkies, line 7: Please add docs 🙂

core-models/src/main/scala/com/mesosphere/usi/core/models/Ids.scala
```diff
@@ -0,0 +1,5 @@
+package com.mesosphere.usi.core.models
```
jeschkies, line 1: Hm, now this is interesting. Should we put the ids in the ids package? I'd say so. This would yield a stuttering api, though: `com.mesosphere.usi.core.models.ids.PodId`. What do you think?

core-models/src/main/scala/com/mesosphere/usi/core/models/PodRecord.scala
```diff
@@ -0,0 +1,9 @@
+package com.mesosphere.usi.core.models
+
+import java.time.Instant
+
+/**
+  * Snapshot of the Pod Spec state that needs to be persisted and can't be obtained from mesos.
```
jeschkies: Please link pod spec:
```suggestion
  * Snapshot of the [[PodSpec]] state that needs to be persisted and can't be obtained from mesos.
```

core-models/src/main/scala/com/mesosphere/usi/core/models/PodSpec.scala
```diff
@@ -0,0 +1,3 @@
...
+
+case class PodSpec(id: PodId)
```
jeschkies, line 7: Please add docs.

jeschkies (commented):
Much cleaner. Thanks. Please add documentation, though.

alenkacz (approved):
LGTM, thanks 👍

core-models/src/main/scala/com/mesosphere/usi/core/models/Ids.scala
meln1k: I don't like too granular packages, I think we can always refactor later if we notice that we have way too many things there.

PR #20, opened 2019-01-29
Compile also mesos client examples

alenkacz:
If we don't compile examples, they will become not up to date with the actual implementation.

### DIFF
--- a/mesos-client/build.gradle
+++ b/mesos-client/build.gradle
@@ -1,3 +1,11 @@
+sourceSets {
+    main {
+        scala {
+            srcDirs = ['src/main/scala', 'examples']
+        }
+    }
+}
+
 dependencies {
     implementation 'org.apache.mesos:mesos:1.7.1'
     implementation 'com.lightbend.akka:akka-stream-alpakka-simple-codecs_2.12:0.20'
--- a/mesos-client/examples/com/mesosphere/mesos/examples/UselessFramework.scala
+++ b/mesos-client/examples/com/mesosphere/mesos/examples/UselessFramework.scala
@@ -1,16 +1,19 @@
 package mesosphere.mesos.examples
 
+import java.util.UUID
+
 import akka.actor.ActorSystem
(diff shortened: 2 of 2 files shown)


zen-dog (approved):
Nice! 🚢

mesos-client/build.gradle
```diff
@@ -1,3 +1,11 @@
+sourceSets {
+    main {
+        scala {
+            srcDirs = ['src/main/scala', 'examples']
```
jeschkies, line 4: I still don't get why this should not be part of `./examples` in root. Just a few days I had to fight for a little more complex folder structure. Now I have to fight against it 😉

jeschkies (commented):
I'm not really convinced that this makes sense.

mesos-client/build.gradle
alenkacz: for others I am including the slack discussion:

just think of this as a separate library. We might want to take this one day and put it into a separate repo for example because we will want to use it in marathon on somewhere else...
this is why I am trying to keep everything belonging to that library part of one module

I understood we are not doing the step of separating it and publishing as a jar dependency now because this way we could iterate on it while working on USI

jeschkies (approved):

mesos-client/build.gradle
meichstedt: I think we need to define what the goal of this repo is, and what isn't.

The goal _is_, to have one implementation that can be used by all our orchestrators/frameworks.

The goal is _not_ to support any of the (related) components _outside_ of this project. I'm aware that implementations like persistence, metrics, leader election, mesos client, etc. are not necessarily part of the scheduler interface. They should be part of the surrounding tooling, and should for now be considered integral parts of this repository.

Please, don't plan with any of these implementations becoming standalone libraries. There is no use for us at this moment to go down that route. If we did, people would assume they're good to use as is, and we don't want to and cannot support someone using e.g. USI with their own persistence implementation, or using our Mesos client with their framework.

PR #21, opened 2019-01-29
Added a note about versioning in USI

meln1k:
Jira issues: DCOS-47505

### DIFF
--- a/README.md
+++ b/README.md
@@ -17,3 +17,12 @@ How to test:
 ```
 $ gradle test
 ```
+
+## Versioning
+
+USI uses [semantic versioning](https://semver.org/). That means 
+we are committed to keep our documented API compatible across releases 
+unless we change the MAJOR version (the first number in the version tuple).
+
+For example version 0.1.0 and 1.0.0 will not have a compatible API, but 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -16,4 +16,13 @@ How to test:
...
+```
+
+## Versioning
+
+USI uses [semantic versioning](https://semver.org/). That means 
+we are committed to keep our documented API compatible across releases 
+unless we change the MAJOR version (the first number in the version tuple).
+
+For example version 0.1.0 and 1.0.0 will not have a compatible API, but 
+0.2.0 and 0.1.0 should be backwards compatible.
```
jeschkies: I think that `0.x` can always break but not `1.x`. See https://semver.org/#spec-item-4
> Major version zero (0.y.z) is for initial development. Anything may change at any time. The public API should not be considered stable.

alenkacz:
Oh nice, so we're actually doing semver and not "pseudo SEMVER" as in Marathon?

jeschkies (approved):

alenkacz (approved):

meichstedt:
👌

PR #24, opened 2019-01-30
Provision Travis CI with Mesos and Zookeeper.

jeschkies:
Summary:
This installs Mesos and Zookeeper before building USI. The services are stopped after the installation.


.travis.yml
```diff
@@ -1,6 +1,9 @@
+sudo: required
 language: scala
 scala:
    - 2.12.2
+before_install:
+  - sudo ./provision.sh
```
jeschkies: We probably want to move this into Gradle and run it from there.

provision.sh
```diff
@@ -0,0 +1,19 @@
...
+set -x -e -o pipefail
+
+apt-key adv --keyserver hkp://keyserver.ubuntu.com:80 --recv E56151BF
+DISTRO=$(lsb_release -is | tr '[:upper:]' '[:lower:]')
+CODENAME=$(lsb_release -cs)
+
+echo "deb http://repos.mesosphere.io/${DISTRO} ${CODENAME} main" | sudo tee /etc/apt/sources.list.d/mesosphere.list
+apt-get -y update
+
+apt-get -y install mesos=1.4.0-2.0.1 zookeeperd
```
jeschkies: The version should come from our Gradle dependency.

build.gradle
```diff
@@ -6,6 +6,13 @@ plugins {
     id 'org.ajoberstar.git-publish' version '2.0.0'
 }
 
+ext {
+    // Scala version for all scala dependencies
+    scalaVersion = '2.12'
+
+    mesosVersion = '1.7.0'
```
jeschkies, line 13: It seems there is no Mesos 1.7.1 for Ubuntu yet.

ci.gradle
```diff
@@ -56,6 +55,26 @@ gitPublish {
 }
 gitPublishCopy.dependsOn(scaladoc)
 
+task provision(type: Exec) {
```
alenkacz, line 58: Just wondering, what do you see as an advantage of having this as a gradle task and not a shell script checked in git? That way it's easier to edit or even e.g. use shellcheck... If we want to have whole lifecycle in gradle, we can call the script from gradle task, I guess... ?

jeschkies: I had it as a script but did not really know where to put it and thought it nicer to have everything in Gradle. I'm happy to move it into a file. Oh, one reason is also the Mesos version.

alenkacz: I find it pretty hard to read and edit, since it's just plain string... if it was up to me, I would create a folder similar to one we have in marathon and place it there. But let's see what others think...

jeschkies: So, `./ci/provision.sh 1.7.0`?

zen-dog (approved):
Agree with Alena on moving the bash code into its own script. Otherwise lgtm 🚢

ci.gradle
kensipe: moving out to `sh` file sounds good.   there is still value in having a gradle task to invoke there right?  a precondition of the task could be the exists of the installation making this part of an incremental build.

mesos-client/examples/com/mesosphere/mesos/examples/UselessFramework.scala
```diff
@@ -1,4 +1,4 @@
-package mesosphere.mesos.examples
+package com.mesosphere.mesos.examples
```
kensipe: this needs to be move out of this example folder

kensipe (changes requested):

mesos-client/examples/com/mesosphere/mesos/examples/UselessFramework.scala
jeschkies: There was a long debate without a solution. Ask @alenkacz about it 😉

kensipe (approved):

PR #25, opened 2019-01-30
Scalafmt applied to a project

alenkacz:
*DISCLAIMER* The plugin currently does not support multimodule projects in a way that you would be able to specify one config shared between all subprojects. I'll fix this ASAP and remove the redundant configs.

Check if there's something super-controversial in the formatting, otherwise we should merge it. I like that the formatting is consistent with every build 😊


core-models/.scalafmt.conf
```diff
@@ -0,0 +1 @@
+maxColumn = 160
```
meln1k: shall we define this on a higher level?

alenkacz: please read the disclaimer :)

meln1k (approved):

mesos-client/src/main/scala/com/mesosphere/mesos/client/MesosCalls.scala
```diff
@@ -25,7 +25,8 @@ class MesosCalls(frameworkId: FrameworkID) {
     * http://mesos.apache.org/documentation/latest/scheduler-http-api/#teardown
     */
   def newTeardown(): Call = {
-    Call.newBuilder()
+    Call
+      .newBuilder()
```
kensipe: I agree with most of the format changes... except this one..  why is this the preference?  can we fix it?   I really don't like it.

mesos-client/src/main/scala/com/mesosphere/mesos/conf/MesosClientSettings.scala
```diff
@@ -8,7 +8,8 @@ case class MesosClientSettings(conf: Config) {
   val master: String = conf.getString("connection.masterUrl")
   val redirectRetries: Int = conf.getInt("connection.redirectRetries")
   // we want FiniteDuration, the conversion is needed to achieve that
-  val idleTimeout: FiniteDuration = Duration.fromNanos(conf.getDuration("connection.idleTimeout").toNanos)
+  val idleTimeout: FiniteDuration =
+    Duration.fromNanos(conf.getDuration("connection.idleTimeout").toNanos)
```
kensipe: is this a line length issue?   160 should cover this right?

metrics-dropwizard/src/main/scala/com/mesosphere/usi/metrics/dropwizard/DropwizardMetrics.scala
```diff
@@ -18,11 +18,16 @@ class DropwizardMetrics(metricsSettings: MetricsSettings, registry: MetricRegist
...
+  private val histogramReservoirHighestTrackableValue =
+    histrogramSettings.reservoirHighestTrackableValue
+  private val histogramReservoirSignificantDigits =
+    histrogramSettings.reservoirSignificantDigits
+  private val histogramReservoirResetPeriodically =
+    histrogramSettings.reservoirResetPeriodically
+  private val histogramReservoirResettingInterval =
+    Duration.ofNanos(histrogramSettings.reservoirResettingInterval.toNanos)
+  private val histogramReservoirResettingChunks =
+    histrogramSettings.reservoirResettingChunks
```
kensipe: ugh

kensipe (changes requested):

mesos-client/src/main/scala/com/mesosphere/mesos/conf/MesosClientSettings.scala
zen-dog: Are 160 characters enough for everybody? ^.^

kensipe:
I don't know where we have this debate.. but I would prefer
* includeCurlyBraceInSelectChains = false (not default)

also... `includeNoParensInSelectChains` which I also think I would prefer breaks scalafmt :(

build.gradle
```diff
@@ -4,12 +4,15 @@ import java.nio.file.Paths
...
     id 'org.ajoberstar.git-publish' version '2.0.0'
+    id 'cz.alenkacz.gradle.scalafmt' version '1.7.0'
 }
 
 allprojects {
 
     version = '0.1'
+
     apply plugin: "scala"
+    apply plugin: 'cz.alenkacz.gradle.scalafmt'
```
jeschkies: This should not be required if we use `plugins { ... }`, right?

core-models/.scalafmt.conf
jeschkies: Really, 160? I thought Github works nice with 120 🙂

mesos-client/src/main/scala/com/mesosphere/mesos/client/MesosCalls.scala
jeschkies: I don't mind but it's not what we had and are used to.

mesos-client/src/main/scala/com/mesosphere/mesos/client/MesosClient.scala
```diff
@@ -283,9 +289,9 @@ object MesosClient extends StrictLogging with StrictLoggingFlow {
     *    available via the materializable-once source, `.mesosSource`, which DOES NOT include the earlier-consumed
     *    SUBSCRIBED event.
     */
-  def apply(conf: MesosClientSettings, frameworkInfo: FrameworkInfo)(
-    implicit
-    system: ActorSystem, materializer: ActorMaterializer): Source[MesosClient, NotUsed] = {
+  def apply(conf: MesosClientSettings, frameworkInfo: FrameworkInfo)(implicit
+                                                                     system: ActorSystem,
```
jeschkies: Oh no. This is an ugly one.

mesos-client/src/main/scala/com/mesosphere/mesos/client/MesosClient.scala
```diff
@@ -320,45 +327,50 @@ object MesosClient extends StrictLogging with StrictLoggingFlow {
...
-    val subscribed: Event.Subscribed,
-    val connectionInfo: MesosClient.ConnectionInfo,
-    /**
-      * Events from Mesos scheduler, sans initial Subscribed event.
-      */
-    val mesosSource: Source[Event, NotUsed])(
-    implicit
-    as: ActorSystem, m: Materializer) extends MesosClient with StrictLoggingFlow {
+class MesosClientImpl(sharedKillSwitch: SharedKillSwitch,
+                      val subscribed: Event.Subscribed,
```
jeschkies: Same here. If I'm not wrong we have a simple tab indention.

mesos-client/src/main/scala/com/mesosphere/mesos/conf/MesosClientSettings.scala
jeschkies: It's too much but I'm probably alone with that. 120 makes reviewing on Github much easier. 160 is ok. Everything above makes it hard to review on MacBooks IMHO.

metrics-dropwizard/src/main/scala/com/mesosphere/usi/metrics/dropwizard/DropwizardMetrics.scala
jeschkies: That's just line length and the same as on the left. We could wrap it with `//format:OFF`, though.

metrics-dropwizard/src/main/scala/com/mesosphere/usi/metrics/dropwizard/reporters/DataDogUDPReporter.scala
```diff
@@ -41,38 +41,46 @@ class DataDogUDPReporter(settings: DataDogUdpReporterSettings, registry: MetricR
...
-    counters: util.SortedMap[String, Counter],
-    histograms: util.SortedMap[String, Histogram],
-    meters: util.SortedMap[String, Meter],
-    timers: util.SortedMap[String, Timer]): Unit = {
+  private def report(socket: ActorRef,
+                     gauges: util.SortedMap[String, Gauge[_]],
+                     counters: util.SortedMap[String, Counter],
+                     histograms: util.SortedMap[String, Histogram],
+                     meters: util.SortedMap[String, Meter],
+                     timers: util.SortedMap[String, Timer]): Unit = {
```
jeschkies: See my comment above. I prefer the Marathon way on the left.

jeschkies (changes requested):
This will make life much easier but we should try to use most of the Marathon specs. I'm also not sure if the application of the Plugin has to be done twice.

core-models/.scalafmt.conf
meichstedt:
```suggestion
maxColumn = 120
```

I use IDEA on a 15" and usually have the project structure opened on the left which leaves me with about 2/3 of my screen for the code view at font size 12. In this setup, I cannot see more than 120 chars.

On 11" or 13" reading diffs at 160 columns is practically impossible. I personally lean towards the legacy ideal of 80, where 120 is an upper limit that should be used for everything that cannot be meaningfully formatted to 80.

core/.scalafmt.conf
```diff
@@ -0,0 +1 @@
+maxColumn = 160
```
meichstedt:
```suggestion
maxColumn = 120
```

mesos-client/.scalafmt.conf
```diff
@@ -0,0 +1 @@
+maxColumn = 160
```
meichstedt:
```suggestion
maxColumn = 120
```

mesos-client/src/main/scala/com/mesosphere/mesos/client/MesosCalls.scala
meichstedt: Ya that's weird

mesos-client/src/main/scala/com/mesosphere/mesos/conf/MesosClientSettings.scala
meichstedt: 120 please 🙏

metrics-dropwizard/.scalafmt.conf
```diff
@@ -0,0 +1 @@
+maxColumn = 160
```
meichstedt:
```suggestion
maxColumn = 120
```

metrics-dropwizard/src/main/scala/com/mesosphere/usi/metrics/dropwizard/reporters/DataDogUDPReporter.scala
meichstedt: +1 😬

persistence/.scalafmt.conf
```diff
@@ -0,0 +1 @@
+maxColumn = 160
```
meichstedt:
```suggestion
maxColumn = 120
```

metrics/.scalafmt.conf
```diff
@@ -0,0 +1 @@
+maxColumn = 160
```
meichstedt:
```suggestion
maxColumn = 120
```

meichstedt (commented):
I hope I wasn't too aggressive with my 120 suggestion 🙈

mesos-client/src/main/scala/com/mesosphere/mesos/client/MesosClient.scala
alenkacz: I don't understand... :/

metrics-dropwizard/src/main/scala/com/mesosphere/usi/metrics/dropwizard/reporters/DataDogUDPReporter.scala
alenkacz: I agree it's different but I find this pretty readable. What's wrong with this?

mesos-client/.scalafmt.conf
```diff
@@ -0,0 +1,3 @@
+maxColumn = 120
+includeCurlyBraceInSelectChains = false
+includeNoParensInSelectChains = false
```
jeschkies: Should this be in all `.scalafmt.conf`? What do you think about symlink all to just one file?

mesos-client/src/main/scala/com/mesosphere/mesos/client/MesosClient.scala
```diff
@@ -150,52 +154,52 @@ object MesosClient extends StrictLogging with StrictLoggingFlow {
...
 
-    val request = HttpRequest(
-      HttpMethods.POST,
-      uri = Uri("/api/v1/scheduler"),
-      entity = HttpEntity(ProtobufMediaType, body),
-      headers = List(headers.Accept(ProtobufMediaType)))
+    val request = HttpRequest(HttpMethods.POST,
+                              uri = Uri("/api/v1/scheduler"),
+                              entity = HttpEntity(ProtobufMediaType, body),
+                              headers = List(headers.Accept(ProtobufMediaType)))
```
jeschkies: Shouldn't this be just and indent as well like on the left?

mesos-client/src/main/scala/com/mesosphere/mesos/client/MesosClient.scala
```diff
@@ -150,52 +154,52 @@ object MesosClient extends StrictLogging with StrictLoggingFlow {
...
 
     val httpConnection = Http().outgoingConnection(url.getHost, url.getPort)
 
-    Source.single(request)
+    Source
+      .single(request)
```
jeschkies: Still not fix 😉

jeschkies (dismissed):
We are getting there.

mesos-client/.scalafmt.conf
alenkacz: I'll add support to the plugin to accept absolute paths and then we'll be able to do just one config. That shouldn't be that much work so this will go

mesos-client/src/main/scala/com/mesosphere/mesos/client/MesosClient.scala
alenkacz: fixed

mesos-client/src/main/scala/com/mesosphere/mesos/client/MesosClient.scala
alenkacz: Scalafmt formats it this way. Sure we can disable but I find this pretty readable. You don't have to indent it like this by yourself, gradle takes care of that so 🤷‍♀️

jeschkies: But I don't like it. I mean we can discuss or just use what we had for Marathon.

mesos-client/.scalafmt.conf
jeschkies: But when is that done? Why not simply do the symlink for now?

metrics-dropwizard/src/main/scala/com/mesosphere/usi/metrics/dropwizard/reporters/DataDogUDPReporter.scala
jeschkies: It kind of wastes real estate. If everyone else is fine let's use this. However, Marathon has a different config. That's confusing.

mesos-client/src/main/scala/com/mesosphere/mesos/client/MesosClient.scala
alenkacz: are you saying it's harder to read (or what's 'not liking it?'). Because to me it's definitely the opposite. It would be harder to write, but you don't have to write it...

mesos-client/.scalafmt.conf
kensipe: @alenkacz I understand that the plugin will be updated soon... and sounds good!   however can we use the same file everywhere.    The top level project has 2 lines... this one has 3 lines.   we should have the same definition in all places.

mesos-client/src/main/scala/com/mesosphere/mesos/client/MesosClient.scala
kensipe: this is one of the formats that I get frustrated with regarding python and pep.   The alignment to the open brace along with line length limits leads to frustrating IMO.   I'm with karsten...  previous indicate is preferred.   it is clear that it is a continuation of the function declaration and allows for verbosity where needed.

mesos-client/src/main/scala/com/mesosphere/mesos/client/MesosClient.scala
kensipe: agree

metrics-dropwizard/.scalafmt.conf
```diff
@@ -0,0 +1,2 @@
+maxColumn = 120
+includeCurlyBraceInSelectChains = false
```
kensipe: lets have the same file in all projects

metrics-dropwizard/src/main/scala/com/mesosphere/usi/metrics/dropwizard/reporters/DataDogAPIReporter.scala
```diff
@@ -51,31 +54,49 @@ class DataDogAPIReporter(settings: DataDogApiReporterSettings, registry: MetricR
...
     val data = buffer
       .insert(0, "{\"series\":[")
       .append("]}")
       .toString
 
     val body = HttpEntity(MediaTypes.`application/json`, data)
-    http.singleRequest(HttpRequest(method = HttpMethods.POST, uri = apiUrl, entity = body)).pipeTo(self)
+    http
+      .singleRequest(HttpRequest(method = HttpMethods.POST, uri = apiUrl, entity = body))
+      .pipeTo(self)
```
kensipe: this is the same issue we had previous with 
```
Call
     . xyz
```
it is frustrating to look at
much prefer

```
http.singleRequest(HttpRequest(method = HttpMethods.POST, uri = apiUrl, entity = body))
       .pipeTo(self)
```

metrics-dropwizard/src/main/scala/com/mesosphere/usi/metrics/dropwizard/reporters/DataDogUDPReporter.scala
kensipe: agree... we can't limit to 120 AND indent 60 :)

kensipe (dismissed):

metrics-dropwizard/src/main/scala/com/mesosphere/usi/metrics/dropwizard/reporters/DataDogAPIReporter.scala
alenkacz: yeah, it does not automatically format this way but it also does not format back. I reverted many places, this one slipped. Thanks

metrics-dropwizard/src/main/scala/com/mesosphere/usi/metrics/dropwizard/reporters/DataDogUDPReporter.scala
alenkacz: Fair point, that I can relate to when it's not "I don't like this". I'll change it, thanks

metrics-dropwizard/src/main/scala/com/mesosphere/usi/metrics/dropwizard/reporters/StatsDReporter.scala
```diff
@@ -41,24 +41,36 @@ class StatsDReporter(settings: StatsdReporterSettings, registry: MetricRegistry)
...
-    counters: util.SortedMap[String, Counter],
-    histograms: util.SortedMap[String, Histogram],
-    meters: util.SortedMap[String, Meter],
-    timers: util.SortedMap[String, Timer]): Unit = {
+  private def report(socket: ActorRef,
+                     gauges: util.SortedMap[String, Gauge[_]],
+                     counters: util.SortedMap[String, Counter],
+                     histograms: util.SortedMap[String, Histogram],
+                     meters: util.SortedMap[String, Meter],
+                     timers: util.SortedMap[String, Timer]): Unit = {
```
jeschkies: Still here. Not sure what to make of it.

metrics-dropwizard/src/main/scala/com/mesosphere/usi/metrics/dropwizard/reporters/DataDogAPIReporter.scala
```diff
@@ -135,10 +171,14 @@ class DataDogAPIReporter(settings: DataDogApiReporterSettings, registry: MetricR
     reportMetered(buffer, name, timer, timestamp)
   }
 
-  private def reportMetric(buffer: StringBuilder, name: String, value: String, timestamp: Long,
-    metricType: String): Unit = {
+  private def reportMetric(buffer: StringBuilder,
+                           name: String,
+                           value: String,
+                           timestamp: Long,
+                           metricType: String): Unit = {
```
jeschkies, line 179: Still here but okay. If some else approves, I'm fine with it.

metrics-dropwizard/src/main/scala/com/mesosphere/usi/metrics/dropwizard/reporters/DataDogAPIReporter.scala
```diff
@@ -86,10 +105,22 @@ class DataDogAPIReporter(settings: DataDogApiReporterSettings, registry: MetricR
     reportMetric(buffer, name, counter.getCount.toString, timestamp, "gauge")
 
   private val histogramSnapshotSuffixes =
-    Seq("min", "average", "median", "75percentile", "95percentile", "98percentile",
-      "99percentile", "999percentile", "max", "stddev")
-  private def reportSnapshot(buffer: StringBuilder, name: String, snapshot: Snapshot, timestamp: Long,
-    scaleMetrics: Boolean): Unit = {
+    Seq("min",
+        "average",
```
timcharper: I tend to agree with @kensipe on indenting further args this way; with really-long-method-names this can cause a lot of our code to be very far to the right.

metrics-dropwizard/src/main/scala/com/mesosphere/usi/metrics/dropwizard/reporters/StatsDReporter.scala
```diff
@@ -69,7 +81,7 @@ class StatsDReporter(settings: StatsdReporterSettings, registry: MetricRegistry)
   private def reportGauge(socket: ActorRef, name: String, gauge: Gauge[_]): Unit = {
     val value: Number = gauge.getValue match {
       case v: Double => if (v.isNaN) 0.0 else v
-      case v: Float => if (v.isNaN) 0.0 else v.toDouble
+      case v: Float  => if (v.isNaN) 0.0 else v.toDouble
```
timcharper: This could be a conflict generating machine? I'd turn alignment off unless if this alignment only happens for one-liners.

metrics-dropwizard/src/main/scala/com/mesosphere/usi/metrics/dropwizard/DropwizardMetrics.scala
```diff
@@ -9,31 +9,32 @@ import com.codahale.metrics
 import com.codahale.metrics.MetricRegistry
 import com.github.rollingmetrics.histogram.{HdrBuilder, OverflowResolver}
 import com.mesosphere.usi.metrics.dropwizard.conf.MetricsSettings
-import com.mesosphere.usi.metrics.{ClosureGauge, Counter, Gauge, Meter, Metrics, SettableGauge, Timer, TimerAdapter, UnitOfMeasurement}
+import com.mesosphere.usi.metrics.{ClosureGauge, Counter, Gauge, Meter, Metrics, SettableGauge, Timer, TimerAdapter, UnitOfMeasurement} //format:OFF
 
-import scala.util.matching.Regex
+  import scala.util.matching.Regex
```
timcharper: What's this space for?

timcharper (approved):
a few comments

PR #26, opened 2019-01-30
Add `test-utils` module

zen-dog:
Summary:
- ported Marathon integration test classes. Most notably `MesosTest`, `ZookeeperServer` and `MesosFacade` and their dependencies
- ported `MesosClientIntegrationTest` into the `mesos-client` project. It depends on the above-mentioned test classes and starts Mesos/Zookeeper to run its tests.

JIRA: DCOS-47483


zen-dog:
Most of the classes here are copy-pasted from Marathon with minimal changes. I removed some unused parts and cleaned up obvious low-hanging fruits. This PR depends on #24 (for provisioning Mesos and Zookeeper)

kensipe (dismissed):
Blocker:  Doesn't work with Java 11

```
WARNING: An illegal reflective access operation has occurred
WARNING: Illegal reflective access by javassist.ClassPool (file:/Users/kensipe/.gradle/caches/modules-2/files-2.1/org.javassist/javassist/3.18.1-GA/d9a09f7732226af26bf99f19e2cffe0ae219db5b/javassist-3.18.1-GA.jar) to method java.lang.ClassLoader.defineClass(java.lang.String,byte[],int,int)
WARNING: Please consider reporting this to the maintainers of javassist.ClassPool
WARNING: Use --illegal-access=warn to enable warnings of further illegal reflective access operations
WARNING: All illegal access operations will be denied in a future release
```

build.gradle
```diff
@@ -37,20 +37,29 @@ allprojects {
...
 
         // Logging
         compile group: 'com.typesafe.scala-logging', name: "scala-logging_$scalaVersion", version: '3.9.2'
+        testRuntime group: 'ch.qos.logback', name: 'logback-classic', version: '1.1.7'
 
         // Akka
         compile group: 'com.typesafe.akka', name: "akka-http_$scalaVersion", version: '10.0.11'
+
+        // Apache Curator. See this [compatibility documentation](http://curator.apache.org/zk-compatibility.html)
+        def withoutZk = { exclude group: 'org.apache.zookeeper', module: 'zookeeper' }
```
alenkacz, line 62: should really all projects depend on curator? Even metrics? I think we should move this to `test-utils/build.gradle` right now

test-utils/src/main/scala/com/mesosphere/utils/http/AkkaHttpResponse.scala
```diff
@@ -0,0 +1,37 @@
...
+import akka.http.scaladsl.Http
+import akka.http.scaladsl.model.{HttpRequest, HttpResponse}
+import akka.stream.Materializer
+import play.api.libs.json.Reads
+
+import scala.concurrent.duration.FiniteDuration
+import scala.concurrent.{ExecutionContext, Future}
+import scala.reflect.ClassTag
+
+object AkkaHttpResponse {
```
alenkacz, line 13: do we need this to test library? There might not even be akka http, right?

build.gradle
```diff
@@ -37,20 +37,29 @@ allprojects {
 
         // Use Scala 2.12 in our library project
         implementation 'org.scala-lang:scala-library:2.12.7'
-        implementation 'com.typesafe.scala-logging:scala-logging_2.12:3.9.2'
+        implementation "com.typesafe.scala-logging:scala-logging_$scalaVersion:3.9.2"
 
         // Use Scalatest for testing our library
         testImplementation 'junit:junit:4.12'
-        testImplementation "org.scalatest:scalatest_$scalaVersion:3.0.5"
+        compile "org.scalatest:scalatest_$scalaVersion:3.0.5"
```
jeschkies: Isn't this `testCompile`?

mesos-client/src/test/scala/com/mesosphere/mesos/client/MesosClientIntegrationTest.scala
```diff
@@ -0,0 +1,147 @@
...
+import scala.annotation.tailrec
+import scala.concurrent.Future
+
+@RunWith(classOf[JUnitRunner])
+class MesosClientIntegrationTest extends AkkaUnitTest
+  with MesosClusterTest
+  with Eventually {
+
+  "Mesos client should successfully subscribe to mesos without framework Id" in withFixture() { f =>
+    When("a framework subscribes")
```
jeschkies: Where is the call?

test-utils/src/main/resources/detect-routable-ip.sh
```diff
@@ -0,0 +1,16 @@
+#!/bin/bash
+
+# Script which returns your primary IP; works on Mac OS X and Linux
+#
+
+case $(uname) in
```
jeschkies: @timcharper what is the reason for this being as script and not Scala code?

test-utils/src/main/scala/com/mesosphere/utils/ProcessOutputToLogStream.scala
```diff
@@ -0,0 +1,12 @@
+package com.mesosphere.utils
+
+import com.typesafe.scalalogging.{Logger, StrictLogging}
+
+import scala.sys.process.ProcessLogger
+
+case class ProcessOutputToLogStream(process: String) extends ProcessLogger with StrictLogging {
```
jeschkies: Could you add some docs?

mesos-client/src/test/scala/com/mesosphere/mesos/client/MesosClientIntegrationTest.scala
```diff
@@ -0,0 +1,147 @@
...
+import org.junit.runner.RunWith
+import org.scalatest.junit.JUnitRunner
+
+import scala.annotation.tailrec
+import scala.concurrent.Future
+
+@RunWith(classOf[JUnitRunner])
+class MesosClientIntegrationTest extends AkkaUnitTest
+  with MesosClusterTest
+  with Eventually {
```
jeschkies: `Eventually` is already part of `AkkaUnitTest`.

test-utils/src/main/scala/com/mesosphere/utils/http/AkkaHttpResponse.scala
alenkacz: oh I see, we use this to talk to mesos API, that was not really obvious at first sight to me :)

alenkacz (approved):
LGTM :) I did not really read it line by line and some parts definitely look like the code could be improved but since it's battle tested and working I am fine with this :)

test-utils/src/main/scala/com/mesosphere/utils/UnitTestLike.scala
```diff
@@ -0,0 +1,76 @@
...
+  with BeforeAndAfterEach
+  with OptionValues
+  with TryValues
+  with AppendedClues
+  with StrictLogging
+  with BeforeAndAfterAll
+  with Eventually
+  with TimeLimitedTests {
+
+  private class LoggingInformer(info: Informer) extends Informer {
```
jeschkies, line 48:
```suggestion
  /**
   * Prepends logs with a position. See [[UnitTestLike.info]].
   */
  private class LoggingInformer(info: Informer) extends Informer {
```

test-utils/src/main/scala/com/mesosphere/utils/UnitTestLike.scala
```diff
@@ -0,0 +1,76 @@
...
+  }
+
+  override val timeLimit = Span(1, Minute)
+
+  override implicit lazy val patienceConfig: PatienceConfig = PatienceConfig(timeout = Span(5, Seconds))
+}
+
+abstract class UnitTest extends WordSpec with UnitTestLike
+
+trait AkkaUnitTestLike extends UnitTestLike {
```
jeschkies, line 74: Please add some docs.

test-utils/src/main/scala/com/mesosphere/utils/http/AkkaHttpResponse.scala
```diff
@@ -0,0 +1,37 @@
...
+
+  def request(request: HttpRequest)(implicit
```
jeschkies: Please add some docs

test-utils/src/main/scala/com/mesosphere/utils/mesos/FaultDomain.scala
```diff
@@ -0,0 +1,11 @@
+package com.mesosphere.utils.mesos
```
jeschkies, line 1: Shouldn't this be part of some models? I don't mean `core-models` but `com.mesosphere.utils.models`.

test-utils/src/main/scala/com/mesosphere/utils/mesos/MesosTest.scala
```diff
@@ -0,0 +1,394 @@
...
+import scala.sys.process.{Process, ProcessBuilder}
+import scala.util.Try
+
+case class MesosConfig(
+                        launcher: String = "posix",
+                        containerizers: String = "mesos",
+                        isolation: Option[String] = None,
+                        imageProviders: Option[String] = None)
+
+case class MesosCluster(
```
jeschkies, line 55: This definitely needs some docs.

test-utils/src/main/scala/com/mesosphere/utils/mesos/MesosTest.scala
```diff
@@ -0,0 +1,394 @@
...
+    agents.foreach(_.start())
+    val masterUri = await(waitForLeader())
+    await(waitForAgents(masterUri))
+    masterUri
+  }
+
+  def waitForLeader(): Future[String] = async {
+    val firstMaster = s"http://${masters.head.ip}:${masters.head.port}"
+    val result = Retry("wait for leader", maxAttempts = Int.MaxValue, maxDuration = waitForMesosTimeout) {
+      Http().singleRequest(Get(firstMaster + "/redirect")).map { result =>
```
jeschkies: What do you think about using the `MesosFacade` instead as we do for agents? We could drop `Retry` then and use `eventually`.

test-utils/src/main/scala/com/mesosphere/utils/mesos/MesosTest.scala
```diff
@@ -0,0 +1,394 @@
...
+        result.discardEntityBytes() // forget about the body
+        if (result.status.isFailure()) {
+          throw new Exception(s"Couldn't determine leader: $result")
+        }
+        result
+      }
+    }
+
+    def maybeFixURI(uri: String): String = {
+      // some versions of mesos issue redirects with broken Location headers; fix them here
```
jeschkies: Is this still the case?

test-utils/src/main/scala/com/mesosphere/utils/mesos/MesosTest.scala
```diff
@@ -0,0 +1,394 @@
...
+    }
+
+    val location = await(result).headers.find(_.lowercaseName() == "location").map(_.value()).getOrElse(firstMaster)
+    maybeFixURI(location)
+  }
+
+  def waitForAgents(masterUri: String): Future[Done] = {
+    val mesosFacade = new MesosFacade(masterUri)
+    Retry.blocking("wait for agents", maxAttempts = Int.MaxValue, maxDuration = waitForMesosTimeout) {
+      val numAgents = mesosFacade.state.value.agents.size
```
jeschkies: See. Here we use it ;). Also, we should just use `eventually` here.

test-utils/src/main/scala/com/mesosphere/utils/mesos/MesosTest.scala
```diff
@@ -0,0 +1,394 @@
...
+  def state = new MesosFacade(Await.result(waitForLeader(), waitForMesosTimeout)).state
+
+  def teardown(): Unit = {
+    val facade = new MesosFacade(Await.result(waitForLeader(), waitForMesosTimeout))
+    val frameworkIds = facade.frameworkIds().value
+
+    // Call mesos/teardown for all framework Ids in the cluster and wait for the teardown to complete
+    frameworkIds.foreach { fId =>
+      facade.teardown(fId)
+      eventually(timeout(1.minutes), interval(2.seconds)) { facade.completedFrameworkIds().value.contains(fId) }
```
jeschkies: See. We use `eventually` here.

jeschkies (changes requested):
Let's get rid off `Retry`.

build.gradle
zen-dog: I did that initially but then I realized that `test-utils`, `persistence-zookeeper`, frameworks and future leader election code will depend on it so I decided to move it up. wdyt?

test-utils/src/main/scala/com/mesosphere/utils/ProcessOutputToLogStream.scala
zen-dog: Hu? To some internal 4-liner test utils class?

build.gradle
zen-dog: `MesosTest` uses `eventually` ¯\_(ツ)_/¯

mesos-client/src/test/scala/com/mesosphere/mesos/client/MesosClientIntegrationTest.scala
zen-dog: Implicitly in `withFixture` but I'll remove it anyway

build.gradle
jeschkies: Sure but why make it a global dependency? Just make it a local compile dependency for `test-utils`.

test-utils/src/main/scala/com/mesosphere/utils/ProcessOutputToLogStream.scala
jeschkies: Yep. This class is not trivial and it would help to understand.

build.gradle
zen-dog: Ah, got ya. Done

test-utils/src/main/scala/com/mesosphere/utils/ProcessOutputToLogStream.scala
zen-dog: I'm open to `suggestion`s

test-utils/src/main/scala/com/mesosphere/utils/mesos/FaultDomain.scala
zen-dog: I'm not starting this discussion _again_. Once we get to the point when they are actually needed somewhere, we can revisit this decision.

test-utils/src/main/scala/com/mesosphere/utils/mesos/MesosTest.scala
zen-dog: Who knows. Let's remove and see what happens

build.gradle
alenkacz: good catch :)

test-utils/src/main/scala/com/mesosphere/utils/mesos/MesosTest.scala
zen-dog: Bam!

test-utils/src/main/resources/detect-routable-ip.sh
timcharper: The motivation at the time I wrote this was I could write this script and just run it on different distros in docker and check if it worked :)

If we can write equivalent Scala code then no problem there.

jeschkies:
I tried to merge master and apply `scalafmtAll` but it did not work. Any ideas why @alenkacz?

test-utils/src/main/resources/detect-routable-ip.sh
jeschkies: Hm, `java.net.InetAddress.getLocalHost.getHostAddress()` returns the same as `detec-routable-ip.sh`. At least on my machine.

zen-dog: Well, if it works on your machine, then we're good to go 😉

jeschkies (approved):

PR #27, opened 2019-01-30
Rename `mesos-client` configuration parameters

zen-dog:
Summary:
- the standard for config files seems to be the kebab-case
- renamed `connection` to `mesos-client` - the former is too generic and who knows what other connections we might end up with
- `MesosSettings` class gets the `mesos-client.*` part of the config tree. This way we can rename parent/s later


kensipe (dismissed):
the example needs to go to the example project... at a min it needs to be a source code space not a random folder

alenkacz (approved):
Thanks

meln1k (approved):

PR #28, opened 2019-01-31
USI Mock implementation

timcharper:
In this PR we implement a basic mocked implementation of the USI. One can provide the mocked mesos flow as seen in the unit test and use this "mock interface" to "launch pods".

JIRA Issues: DCOS-47473


core-models/src/main/scala/com/mesosphere/usi/core/models/Goal.scala
```diff
@@ -0,0 +1,8 @@
+package com.mesosphere.usi.core.models
+
+sealed trait Goal
+
+object Goal {
```
jeschkies, line 8: I guess I'm a little repetitive but please add docs 🙂

core-models/src/main/scala/com/mesosphere/usi/core/models/PodRecord.scala
```diff
@@ -1,8 +1,11 @@
...
+
 
 /**
   * Snapshot of the [[PodSpec]] state that needs to be persisted and can't be obtained from mesos.
+  *
   * @param podId id of the pod
   */
-case class PodRecord(podId: PodId)
\ No newline at end of file
+case class PodRecord(podId: PodId, launchedAt: Instant, agentId: AgentId)
```
jeschkies, line 18: What does `launchedAt` mean? Could you add the `@param` docs?

core-models/src/main/scala/com/mesosphere/usi/core/models/PodSpec.scala
```diff
@@ -4,4 +4,4 @@ package com.mesosphere.usi.core.models
   * Specification of a pod.
   * @param id Id of the specification.
   */
-case class PodSpec(id: PodId)
\ No newline at end of file
+case class PodSpec(id: PodId, goal: Goal, runSpec: String)
```
jeschkies: `@param` docs are missing. Should we type `runSpec`?

core/build.gradle
```diff
@@ -1,3 +1,5 @@
 dependencies {
     compile project(':core-models')
+
+    compile 'com.typesafe.akka:akka-stream_2.12:2.5.19'
```
jeschkies: We have a global `scalaVersion` defined.

core/src/main/scala/com/mesosphere/usi/core/Response.scala
```diff
@@ -0,0 +1,34 @@
+package com.mesosphere.usi.core
+
+import com.mesosphere.usi.models.{Mesos, StatusMessage}
+
+sealed trait Response {
```
jeschkies: What is a response?

core/src/main/scala/com/mesosphere/usi/core/Response.scala
```diff
@@ -0,0 +1,34 @@
...
+  def statusMessages: List[StatusMessage]
+
+  def mesosCalls: List[Mesos.Call]
+}
+
+case class ResponseWithMessages(statusMessages: List[StatusMessage] = Nil, mesosCalls: List[Mesos.Call] = Nil) extends Response
+
+case class ResponseBuilder(reverseStatusMessage: List[StatusMessage] = Nil, reverseMesosCalls: List[Mesos.Call] = Nil) extends Response {
```
jeschkies: Maybe I misunderstand something but should the `ResponseBuilder` have a `build` method?

core/src/main/scala/com/mesosphere/usi/core/Scheduler.scala
```diff
@@ -1,14 +1,27 @@
...
 
-import com.mesosphere.usi.core.models.PodSpec
+import akka.NotUsed
+import akka.stream.BidiShape
+import akka.stream.scaladsl.{BidiFlow, Broadcast, GraphDSL}
+import com.mesosphere.usi.models.{Mesos, SpecMessage, StatusMessage}
 
-import scala.concurrent.Future
+object Scheduler {
+  def graph: BidiFlow[SpecMessage, StatusMessage, Mesos.Event, Mesos.Call, NotUsed] = {
```
jeschkies: Could you add docs to this method? It seems to be of importance but I have no idea what's really happening here?

core/src/main/scala/com/mesosphere/usi/core/SchedulerLogic.scala
```diff
@@ -0,0 +1,251 @@
...
+
+import java.time.Instant
+
+import com.mesosphere.usi.core.models._
+import com.mesosphere.usi.models._
+
+object SchedulerLogic {
+  private def assertValidTransition(oldSpec: PodSpec, newSpec: PodSpec): Unit = {
+    if ((oldSpec.goal == Goal.Terminal) && (newSpec.goal == Goal.Running))
+      throw new IllegalStateException(s"Illegal state transition Terminal to Running for podId ${oldSpec.id}")
```
jeschkies: How are exceptions handled and where? Should this bubble up to the orchestration layer?

core/src/main/scala/com/mesosphere/usi/core/SchedulerLogic.scala
```diff
@@ -0,0 +1,251 @@
...
+          taskIdsFor(podSpec).foreach { taskId =>
+            b = b.withMesosCall(Mesos.Call.Kill(taskId))
+          }
+        }
+    }
+    b
+  }
+
+
+  def matchOffer(podSpecs: Map[PodId, PodSpec])(offer: Mesos.Event.Offer, pendingLaunch: Set[PodId], responseBuilder: ResponseBuilder): ResponseBuilder = {
```
jeschkies: I like that this is a pure function. What is the motivation for using the builder pattern?

core/src/test/scala/com/mesosphere/usi/core/AkkaUnitTest.scala
```diff
@@ -0,0 +1,14 @@
+package com.mesosphere.usi.core
+
+import akka.actor.ActorSystem
+import org.scalatest.concurrent.ScalaFutures
+import org.scalatest.{BeforeAndAfterAll, Suite}
+
+trait AkkaUnitTest extends Suite with BeforeAndAfterAll with ScalaFutures {
```
jeschkies: See `test-utils` from @zen-dog 🙂

jeschkies (commented):
I know this is an early draft so bear with my comments 🙂 

Overall, I had difficulties finding the actual interface. I don't know the entry point into the scheduler.

I was also wondering if we really neat complex graphs. I'm probably missing a lot but I expected something like the Elm/Redux structure where we would fold over a stream of events and the events being Mesos updates or PodSpec changes.

core/src/test/scala/com/mesosphere/usi/core/AkkaUnitTest.scala
alenkacz: how is that related? since this is unit test? 🤔

core/src/main/scala/com/mesosphere/usi/core/Response.scala
alenkacz: yeah I have problems with naming in this part as well... :/ I think we should find better name for this class

core/src/main/scala/com/mesosphere/usi/core/Scheduler.scala
alenkacz: it helped me to draw it down

```    * {{{
    *                 +------+
    *  SpecMessages ~>|      |~> StatusMessages
    *                 | bidi |
    *  Mesos.Call <~  |      |<~ Mesos.Event
    *                 +------+
    * }}}
```

but I still don't totally get the flow of data here. I would expect SpecMessage triggers Mesos.Call and Mesos.Event triggers StatusMessage but this does not seem to be the case right now. @timcharper

alenkacz: also I guess this is not the most highest level interface, right? The one framework will interact... Because then I would expect it to be just `SpecMessages ~> StatusMessages` I think I am still puzzled by the fact that we expose things that are to me on totally different levels in one component - that's the outside USI interface as well as stuff happening internally (mesos calls and events). Could you maybe clarify that?

jeschkies: +1

The visualization helps a lot.

core/src/test/scala/com/mesosphere/usi/core/AkkaUnitTest.scala
jeschkies: Alex imported `AkkaUnitTest` as well. `test-utils` is not just for integration tests.

core/src/main/scala/com/mesosphere/usi/core/Scheduler.scala
alenkacz: I still believe the wrapper handling the lifecycle of mesos client should be part of the USI. We don't have to do it in this PR but I kind of strongly feel that this should be handled by the library.

I really like the separation of concerns and I think we're not gonna lose it if we include the wrapper simplifying this to `SpecMessages ~> StatusMessages`. That to me is the interface I would want to use from a framework perspective when using scheduling library

alenkacz: also please put the ascii art from slack in here as a scaladoc

core/src/main/scala/com/mesosphere/usi/core/Scheduler.scala
```diff
@@ -1,14 +1,27 @@
...
-class Scheduler  {
+          schedulerLogic.out ~> broadcast.in
 
-  def schedule(podSpec: PodSpec): Future[String] = {
-    println(s"scheduling pod ${podSpec.id}")
-    Future.successful("deploymentId")
+          val mesosCalls = broadcast.out(0).mapConcat { response => response.mesosCalls }
+          val statuses = broadcast.out(1).mapConcat { response => response.statusMessages }
+
+          BidiShape.apply(schedulerLogic.in0, statuses.outlet, schedulerLogic.in1, mesosCalls.outlet)
```
alenkacz: can we create variables for in0 and in1 to make it easier to read? something like `val specMessages = schedulerLogic.in0` and then using that in `BidiShape.apply`

core/src/main/scala/com/mesosphere/usi/core/SchedulerLogicGraph.scala
```diff
@@ -0,0 +1,81 @@
...
+  val BUFFER_SIZE = 32
+}
+class SchedulerLogicGraph extends GraphStage[FanInShape2[SpecMessage, Mesos.Event, Response]] {
+  import SchedulerLogicGraph.BUFFER_SIZE
+
+  val mesosEvents = Inlet[Mesos.Event]("mesos-events")
+  val specMessages = Inlet[SpecMessage]("specs")
+  val responses = Outlet[Response]("response")
+  // Define the (sole) output port of this stage
+  val out: Outlet[Int] = Outlet("NumbersSource")
```
alenkacz: I think you did not mean to have the `NumbersSource` here :)

core/src/main/scala/com/mesosphere/usi/core/SchedulerLogicGraph.scala
```diff
@@ -0,0 +1,81 @@
...
+
+object SchedulerLogicGraph {
+  val BUFFER_SIZE = 32
+}
+class SchedulerLogicGraph extends GraphStage[FanInShape2[SpecMessage, Mesos.Event, Response]] {
+  import SchedulerLogicGraph.BUFFER_SIZE
+
+  val mesosEvents = Inlet[Mesos.Event]("mesos-events")
+  val specMessages = Inlet[SpecMessage]("specs")
+  val responses = Outlet[Response]("response")
```
alenkacz: can we find better name for Response? To me even something as simple as `SchedulerLogicOutput` would be better...

core/src/main/scala/com/mesosphere/usi/core/SchedulerLogicGraph.scala
```diff
@@ -0,0 +1,81 @@
...
+      override def preStart(): Unit = {
+        // Start the stream
+        pull(specMessages)
+        pull(mesosEvents)
+      }
+
+      def handleResponse(response: Response): Unit = {
+        if (isAvailable(responses)) {
+          if (pendingResponses.nonEmpty) {
+            throw new IllegalStateException("We should always immediately push on pull if responses are queued")
```
alenkacz: this part I don't understand... why is this state illegal? Especially since on l.66 you do `pendingResponses.enqueue(response)` - how do we ensure that next time we don't end up at this place with pendingResponses containing a value? 🤔

core/src/main/scala/com/mesosphere/usi/core/SchedulerLogicGraph.scala
```diff
@@ -0,0 +1,81 @@
...
+import akka.stream.stage.{GraphStage, GraphStageLogic, InHandler, OutHandler}
+import akka.stream.{Attributes, FanInShape2, Inlet, Outlet}
+import com.mesosphere.usi.models.{Mesos, SpecMessage}
+
+import scala.collection.mutable
+
+object SchedulerLogicGraph {
+  val BUFFER_SIZE = 32
+}
+class SchedulerLogicGraph extends GraphStage[FanInShape2[SpecMessage, Mesos.Event, Response]] {
```
alenkacz: I think this graph deserves ASCII art as well

core/src/main/scala/com/mesosphere/usi/models/Mesos.scala
```diff
@@ -0,0 +1,53 @@
+package com.mesosphere.usi.models
+
+import com.mesosphere.usi.core.models.AgentId
+
+/**
+  * Mock Mesos calls / state
+  */
+object Mesos {
```
alenkacz: I guess this one will go when we put in the mesos client implementation

core/src/main/scala/com/mesosphere/usi/core/SchedulerLogicGraph.scala
alenkacz: so basically this piece of logic is just scheduler processing with added backpressure, correct? Could we put that into docs and maybe even rename the class? One has to figure out now by reading the code :)

core/src/main/scala/com/mesosphere/usi/core/SchedulerLogicGraph.scala
```diff
@@ -0,0 +1,81 @@
+package com.mesosphere.usi.core
+
+import akka.stream.stage.{GraphStage, GraphStageLogic, InHandler, OutHandler}
+import akka.stream.{Attributes, FanInShape2, Inlet, Outlet}
+import com.mesosphere.usi.models.{Mesos, SpecMessage}
+
+import scala.collection.mutable
+
+object SchedulerLogicGraph {
+  val BUFFER_SIZE = 32
```
alenkacz, line 10: should this in the end be exposed as a config value for orchestrator to increase/decrease based on their environment?

core/src/test/scala/com/mesosphere/usi/core/SchedulerTest.scala
```diff
@@ -0,0 +1,65 @@
...
+    implicit val materializer = ActorMaterializer()
+    val podId = PodId("pod-1")
+    val (input, output) = Source.queue[SpecMessage](16, OverflowStrategy.fail)
+      .via(mockedScheduler)
+      .toMat(Sink.queue())(Keep.both)
+      .run
+
+    input.offer(PodSpecChange(podId, Some(PodSpec(podId, Goal.Running, RunSpec(1)))))
+    inside(output.pull().futureValue) {
+      case Some(podRecord: PodRecordChange) =>
```
alenkacz: why not `PodSpecChange`? It's a little bit weird that I put in PodSpec and get PodRecord... I always thought that I should get back PodSpecChange... and records are just implementation detail of the scheduler

core/src/test/scala/com/mesosphere/usi/core/SchedulerTest.scala
```diff
@@ -0,0 +1,65 @@
...
+
+  "It reports a running task when I provide " in {
+    implicit val materializer = ActorMaterializer()
+    val podId = PodId("pod-1")
+    val (input, output) = Source.queue[SpecMessage](16, OverflowStrategy.fail)
+      .via(mockedScheduler)
+      .toMat(Sink.queue())(Keep.both)
+      .run
+
+    input.offer(PodSpecChange(podId, Some(PodSpec(podId, Goal.Running, RunSpec(1)))))
```
alenkacz: shouldn't snapshot be always first? 🤔

core/src/main/scala/com/mesosphere/usi/core/SchedulerLogic.scala
```diff
@@ -0,0 +1,251 @@
...
+      }
+      if (shouldBeLaunched)
+        cachedPendingLaunch += podId
+      else
+        cachedPendingLaunch -= podId
+    }
+  }
+
+  def handleSpecMessage(msg: SpecMessage): Response = {
+    val handleResponse = SchedulerLogic.handleSpecMessage(this.podRecords)(this.podSpecs, msg)
```
alenkacz: why isn't this `(podRecords, podSpecs)(msg)`?

core/src/main/scala/com/mesosphere/usi/core/SchedulerLogic.scala
```diff
@@ -0,0 +1,251 @@
...
+        }
+      }
+    }
+    b
+  }
+
+  private def podIdFor(taskId: TaskId): PodId = PodId(taskId)
+}
+
+class SchedulerLogic {
```
alenkacz: I am still struggling - when do you put a method here and when in `SchedulerLogic` object? In class are only functions that modify the state?

core/src/main/scala/com/mesosphere/usi/core/SchedulerLogic.scala
```diff
@@ -0,0 +1,251 @@
...
+          HandleMesosEventResponse.empty
+        }
+    }
+  }
+
+  /**
+    * We remove a task if it is not reachable and running, and it has no podSpec defined
+    *
+    * Should be called with the effects already applied for the specified podIds
+    * @param podIds podIds changed during the last frame
```
alenkacz: you're refering to a frame here but we don't have frames anywhere. I think we should be consistent with naming...

core/src/main/scala/com/mesosphere/usi/core/SchedulerLogic.scala
```diff
@@ -0,0 +1,251 @@
...
+
+    pendingLaunch.foreach { podId =>
+      b = b.withChangeMessage(PodRecordChange(podId, Some(PodRecord(podId, Instant.now(), offer.agentId))))
+    }
+    b = b.withMesosCall(Mesos.Call.Accept(offer.offerId, operations = operations))
+
+    b
+  }
+
+  def pendingLaunch(podId: PodId, goal: Goal, podRecord: Option[PodRecord]): Boolean = {
```
alenkacz: podId is not used. Also `isPendingLaunch`?

core/src/main/scala/com/mesosphere/usi/core/SchedulerLogic.scala
```diff
@@ -0,0 +1,251 @@
...
+    * Maintains the internal cache of pods pending launch, so a full scan isn't required on every offer
+    *
+    * Must be called after podSpecs are updated, and response effects are applied.
+    *
+    * @param podIds
+    */
+  private def updatePendingLaunch(podIds: Set[PodId]): Unit = {
+    podIds.foreach { podId =>
+      val shouldBeLaunched = podSpecs.get(podId).exists { podSpec =>
+        SchedulerLogic.pendingLaunch(podId, podSpec.goal, podRecords.get(podId))
```
alenkacz: can we do `import SchedulerLogic._` please? It would make stuff easier to read...

core/src/main/scala/com/mesosphere/usi/core/SchedulerLogic.scala
```diff
@@ -0,0 +1,251 @@
...
+    Seq(pod.id.value)
+  }
+
+  def handlePodSpecUpdate(podSpecs: Map[PodId, PodSpec], podRecords: Map[PodId, PodRecord])
+                         (podId: PodId, newState: Option[PodSpec], responseBuilder: ResponseBuilder): ResponseBuilder = {
+    var b = responseBuilder
+
+    newState match {
+      case None =>
+        b = b.withChangeMessage(PodStatusChange(podId, None))
```
alenkacz: I don't understand, why do we need two here? Why is `PodStatusChange` not enough?

core/src/main/scala/com/mesosphere/usi/core/SchedulerLogic.scala
```diff
@@ -0,0 +1,251 @@
...
+        if (podRecords.contains(podId)) {
+          b = b.withChangeMessage(PodRecordChange(podId, None))
+        }
+
+      case Some(podSpec) =>
+
+        podSpecs.get(podId).foreach(assertValidTransition(_, podSpec))
+
+        if (podSpec.goal == Goal.Terminal) {
+          taskIdsFor(podSpec).foreach { taskId =>
```
alenkacz: I hope we're gonna extract the killing logic out of here later. E.g. I don't think we should kill every time you receive update. Do we have JIRA to implement proper killing?

alenkacz (approved):
Uff, I spent several hours going through this and I still don't feel like I am done. What would help me a lot at this stage is if some things were not as deep as they are (e.g. we have killing to mesos implemented as well as revive or pruning tasks. That's just too low level for me at this point when I want to focus on the shape of the overall interface). If this was only graphs + very dummy logic, it would be much easier.

Overall I think we have the biggest problem with naming, we have a lot of methods like "handleXYZ" calling one another and just looking at the method it's not always apparent what is it supposed to do.  E.g. `handlePodSpecUpdate` and `handleSpecMessage` where the second is handling `PodSpecChange`. What's the difference between update and change.

Also I think we definitely need better name for `Response`.

I am on vacation next week so I don't want to block this and at the same time I think it's not set in stone and we can still improve on that so marked it as ✅

core-models/src/main/scala/com/mesosphere/usi/core/models/Goal.scala
```diff
@@ -0,0 +1,19 @@
+package com.mesosphere.usi.core.models
+
+/**
+  * Target state of PodSpec. Scheduler will work towards achieving this goal state.
```
jeschkies:
```suggestion
  * Target state of [[PodSpec]]. Scheduler will work towards achieving this goal state.
```

core-models/src/main/scala/com/mesosphere/usi/core/models/Goal.scala
```diff
@@ -0,0 +1,19 @@
...
+  */
+sealed trait Goal
+
+object Goal {
+
+  /**
+    * PodSpec with this goal should be started if is not running
```
jeschkies:
```suggestion
    * [[PodSpec]] with this goal should be started if is not running
```

core-models/src/main/scala/com/mesosphere/usi/core/models/Goal.scala
```diff
@@ -0,0 +1,19 @@
...
+    */
+  case object Running extends Goal
+
+  /**
+    * PodSpec with this goal should be killed if it is running.
```
jeschkies:
```suggestion
    * [[PodSpec]] with this goal should be killed if it is running.
```

core-models/src/main/scala/com/mesosphere/usi/core/models/PodRecord.scala
```diff
@@ -1,8 +1,13 @@
...
 
+import java.time.Instant
+
 
 /**
   * Snapshot of the [[PodSpec]] state that needs to be persisted and can't be obtained from mesos.
+  *
   * @param podId id of the pod
+  * @param launchedAt time at which we accepted an offer and initiated launch of this pod
+  * @param agentId id of mesos agent from which we accepted offer to launch this pod
```
jeschkies:
```suggestion
  * @param agentId id of Mesos agent from which we accepted offer to launch this pod
```

core/build.gradle
jeschkies:
```suggestion
    compile "com.typesafe.akka:akka-stream_$scalaVersion:2.5.19"
```

core/src/main/scala/com/mesosphere/usi/core/Scheduler.scala
jeschkies: Could we add the flow chart to the docs above. If it uses PlantUML or Mermaid syntax I'm more than happy to render them in our scaladocs.

core/src/main/scala/com/mesosphere/usi/core/SchedulerLogic.scala
```diff
@@ -0,0 +1,251 @@
...
+}
+
+class SchedulerLogic {
+  var podSpecs: Map[PodId, PodSpec] = Map.empty
+  var podRecords: Map[PodId, PodRecord] = Map.empty
+  var podStatuses: Map[PodId, PodStatus] = Map.empty
+
+  var cachedPendingLaunch: Set[PodId] = Set.empty
+
+  private def applyChanges(response: Response): Unit = {
```
jeschkies: This should have some docs. I understand the `response` basically records all changes we want to have. Is that correct?

jeschkies (commented):
I slowly get an idea of the flow. I think all should have a go at this code before we merge.

I'm a little unsure about the response builder. I like it on principles but we might run into issues soon.

In my [refactoring](https://github.com/mesosphere/marathon/pull/6746) I've found that it was not enough to just record changes since some logic further down would require an already update state. Eg in [TaskReplaceActorBehaviour.process](https://github.com/mesosphere/marathon/pull/6746/files#diff-799d40bf745dd8a6faf58e52993998b8R72) we have
```scala
def process(completedPhases: Int, frame: Frame): FrameProcessor.ProcessResult = {
    if (check(frame)) {
      FrameProcessor.Stop
    } else {
      val frameAfterKilling = if (completedPhases == 0) {
        killImmediately(ignitionStrategy.nrToKillImmediately, frame)
      } else {
        killNext(frame)
      }
      val nextFrame = launching(frameAfterKilling)
      FrameProcessor.Continue(nextFrame)
    }
  }
  }
```

`killNext` and `launching` return a new `Frame`, ie a new state. `launching` needs to see the changes from `killNext`. If `killNext` would "just" return the queued responses `launching` would have to apply them first. That was tedious so I changed the [Frame to do that for me](https://github.com/mesosphere/marathon/pull/6746/files#diff-41d67b6cd090ef8b479199b2786fde10R54). I'm not saying my solution is ideal but it made life easier for the core logic. In essence the `Frame` is a builder that also has the intermediate state.

core/build.gradle
zen-dog:
```suggestion
```

zen-dog: `akka-stream` is now defined at the top level

core/src/main/scala/com/mesosphere/usi/core/Response.scala
timcharper: It does not need one

core/src/main/scala/com/mesosphere/usi/core/SchedulerLogic.scala
timcharper: We discussed this in a meeting

core/src/main/scala/com/mesosphere/usi/core/SchedulerLogic.scala
timcharper: Frames, as a concept, still exists

core/src/main/scala/com/mesosphere/usi/core/SchedulerLogic.scala
timcharper: Yes, that is the intention. I can add a comment make that more clear. No specific JIRA, but, then again, were not even doing proper test killing anyways. As an aside, is there really a downside to sending a kill on every terminal update? It's kind of an idempotent thing.

core/src/main/scala/com/mesosphere/usi/models/package.scala
```diff
@@ -0,0 +1,45 @@
...
+
+  /**
+    * Persistent record of agent attributes and other pertinent non-recoverable facts that are inputs for launching pods
+    */
+  case class AgentRecord(id: AgentId, hostname: String)
+
+
+  sealed trait ReservationState
+
+  object ReservationState {
```
kaiwalyajoshi: Does it make sense to introduce another state such as `Stale`, `Obsolete` or `Expired` for cases where `TASK_GONE_BY_OPERATOR` can invalidate a previously valid reservation on a lost agent for example?

core/src/main/scala/com/mesosphere/usi/core/Scheduler.scala
kaiwalyajoshi: Seconded, even better if `schedulerLogic.in0` and `schedulerLogic.in1` can match `MesosEvents` and `FrameworkIn` from the diagram.

core/src/main/scala/com/mesosphere/usi/core/Scheduler.scala
```diff
@@ -1,13 +1,65 @@
...
 
-class Scheduler {
+/*
+ * Provides the scheduler graph component. The component has two inputs, and two outputs:
+ *
+ * Input:
+ * 1) SpecEvents - Used to replicate the specification state from the framework to the USI scheduler
+ * 2) MesosEvents - Events from Mesos; offers, task status updates, etc.
+ *
+ * Output:
```
kaiwalyajoshi, line 17: If I'm not mistaken, `USIStateEvents` and `MesosCalls` collectively make up a `Frame` here. If so it would be nice to depict that in the diagram and in the writeup here.

core/src/main/scala/com/mesosphere/usi/core/SchedulerLogic.scala
timcharper: Ideally we'd crash.

Also, this is a mock implementation.

timcharper:
@alenkacz  I would recommend you just focus on the overall interface. The details aren't important right now. They're place holders and places where logic will go in the future. It's been helpful for me to think through it, I'm sorry that it has been a distraction and source of confusion for you.

core/src/main/scala/com/mesosphere/usi/models/package.scala
timcharper: Yeah, totally, let's go in to detail on this later on.

core/src/main/scala/com/mesosphere/usi/core/Scheduler.scala
timcharper: They make up a FrameResult, or, all of the effects since processing some incoming event (spec updated event or mesos event)

kaiwalyajoshi (approved):

core/src/main/scala/com/mesosphere/usi/core/SchedulerLogicGraph.scala
timcharper: ✅

core/src/main/scala/com/mesosphere/usi/core/SchedulerLogic.scala
timcharper: There's not that much here, I generally dislike splat imports. See if you still feel that way? I've moved a lot of logic out.

core-models/src/main/scala/com/mesosphere/usi/core/models/RunSpec.scala
```diff
@@ -0,0 +1,7 @@
+package com.mesosphere.usi.core.models
+
+/**
+  * Specification used to launch a [[PodSpec]]
+  * @param cpus number of cpus requested by the pod
+  */
+case class RunSpec(cpus: Int)
```
zen-dog:
```suggestion
case class RunSpec(cpus: Double)
```

core/src/main/scala/com/mesosphere/usi/core/Scheduler.scala
```diff
@@ -1,13 +1,65 @@
...
+ * 1) USIStateEvents - Used to replicate the state of pods, agents, and reservations to the framework
```
zen-dog: Nit: Throughout this review, I kept reading `USIStateEvent` as `USIS-tate-events` where `USIS` sounds somewhat like `ISIS` 😱  I know USI is an abbreviation but could we rename it to `UsiStateEvent` or even just `SchedulerStateEVent`? Prefixes usually have value but in this case, we could also omit it altogether.

core/src/main/scala/com/mesosphere/usi/core/Scheduler.scala
```diff
@@ -1,13 +1,65 @@
...
+ * 2) MesosCalls - Actions, such as revive, kill, accept offer, etc., used to realize the specification.
+ *
+ * Fully wired, the graph looks like this at a high-level view:
```
zen-dog, line 21: Nice graphics 👏

core/src/main/scala/com/mesosphere/usi/core/SchedulerLogic.scala
```diff
@@ -0,0 +1,201 @@
...
+import com.mesosphere.usi.models._
+
+/**
+  * Container class responsible for keeping track of the frame and cache.
+  *
+  * As a general rule, business logic does not manipulate the USI state directly, but rather does so by returning
+  * effects.
+  */
+private[core] class SchedulerLogic {
+  private var frame: Frame = Frame(Map.empty, Map.empty, Map.empty)
```
zen-dog: Wdyt about moving this 3x `Map.empty, Map.empty, Map.empty` to the Frame's constructor as default values?

core/src/main/scala/com/mesosphere/usi/core/SchedulerLogic.scala
```diff
@@ -0,0 +1,201 @@
...
+    effects = effects.withMesosCall(Mesos.Call.Accept(offer.offerId, operations = operations))
+
+    effects
+  }
+
+  private[core] def pendingLaunch(goal: Goal, podRecord: Option[PodRecord]): Boolean = {
+    (goal == Goal.Running) && podRecord.isEmpty
+  }
+
+  private[core] def computeNextStateForPods(frame: Frame)(changedPodIds: Set[PodId]): FrameEffects = {
```
zen-dog: Please add documentation (with the focus on state transitions)

core/src/main/scala/com/mesosphere/usi/core/SchedulerLogic.scala
```diff
@@ -0,0 +1,201 @@
...
+    changedPodIds.foldLeft(FrameEffects.empty) { (initialEffects, podId) =>
+      var effects = initialEffects
+
+      frame.podSpecs.get(podId) match {
+        case None =>
+          // TODO - this should be spurious if the podStatus is non-terminal
+          if (frame.podStatuses.get(podId).exists { status =>
```
zen-dog: I'm not sure I understand what case is handled here - changed `podId` isn't part of the `podSpecs` but has a `podStatus` so we handle it as a terminal or unreachable?

core/src/main/scala/com/mesosphere/usi/models/SpecEvent.scala
```diff
@@ -0,0 +1,21 @@
...
+  * @param reservationSpecs
+  */
+case class SpecsSnapshot(podSpecs: Seq[PodSpec], reservationSpecs: Seq[ReservationSpec]) extends SpecEvent
+
+/**
+  * Used to communicate that a specification was updated (added, changed, or deleted)
+  */
+sealed trait SpecUpdated extends SpecEvent
+
+case class PodSpecUpdated(id: PodId, newState: Option[PodSpec]) extends SpecUpdated
```
zen-dog: So if the `newState == None` we infer that the podSpec was deleted? Wdyt about making it a `PodSpecDeleted(id: PodId)`?

core/src/main/scala/com/mesosphere/usi/models/Mesos.scala
zen-dog: It's in now ;)

core/src/main/scala/com/mesosphere/usi/core/SchedulerLogicGraph.scala
zen-dog: Absolutely. We already have an established pattern of using typesafe config (with Settings* classes) - this should be used here too.

core/src/main/scala/com/mesosphere/usi/core/SchedulerLogic.scala
```diff
@@ -0,0 +1,201 @@
...
+    status.taskStatuses.values.forall(status => status == Mesos.TaskStatus.TASK_RUNNING)
+  }
+
+  private def taskIdsFor(pod: PodSpec): Seq[TaskId] = {
+    // TODO - temporary stub implementation
+    Seq(pod.id.value)
+  }
+
+  /**
+    * Process the modification of some podSpec.
```
zen-dog: I believe this comment belongs to the `computeNextStateForPods` method below

core/src/main/scala/com/mesosphere/usi/core/SchedulerLogic.scala
```diff
@@ -0,0 +1,201 @@
...
+    *
+    * - If a podSpec is deleted:
+    *   - Also delete the podRecord.
+    *   - Prune a terminal / unknown task status.
+    * - If a podSpec is marked as terminal, then issue a kill.
+    *
+    * @return The effects of launching this offer
+    */
+  private def matchOffer(offer: Mesos.Event.Offer, pendingLaunchPodSpecs: Seq[PodSpec]): FrameEffects = {
+    var effects = FrameEffects.empty
```
zen-dog: Wdyt about moving this already into a dedicated `OfferMatcher` class with some interface and one (naive) first-come-first-served implementation? This would show where we want to go with it eventually.

core/src/main/scala/com/mesosphere/usi/core/Frame.scala
```diff
@@ -0,0 +1,61 @@
...
+
+case class Frame(podSpecs: Map[PodId, PodSpec], podRecords: Map[PodId, PodRecord], podStatuses: Map[PodId, PodStatus]) {
+
+  def applyStateEffects(effects: FrameEffects): Frame = {
+    var newPodRecords = podRecords
+    var newPodStatuses = podStatuses
+    effects.stateEvents.foreach {
+      case recordChange: PodRecordUpdated =>
+        recordChange.newRecord match {
+          case Some(newRecord) =>
```
zen-dog: It's much more readable to have `PodRecordUpdated` and `PodRecordDelete` than inferring it from the `Option`

core/src/main/scala/com/mesosphere/usi/core/Frame.scala
```diff
@@ -0,0 +1,61 @@
...
+            newPodRecords = newPodRecords.updated(recordChange.id, newRecord)
+          case None =>
+            newPodRecords -= recordChange.id
+        }
+      case statusChange: PodStatusUpdated =>
```
zen-dog: Same here

core-models/src/main/scala/com/mesosphere/usi/core/models/PodSpec.scala
```diff
@@ -1,7 +1,15 @@
 package com.mesosphere.usi.core.models
 
 /**
-  * Specification of a pod.
-  * @param id Id of the specification.
+  * Specification of some Pod that should be launched.
+  *
+  * Pods are launched at-most-once. It is illegal to transition of podSpec from goal terminal to goal running.
```
jeschkies:
```suggestion
  * Pods are launched at-most-once. It is illegal to transition of pod spec from goal terminal to goal running.
```

We should agree on one way to spell it. Camel case is fine with me.

core/src/main/scala/com/mesosphere/usi/core/Frame.scala
```diff
@@ -0,0 +1,61 @@
...
+      case SpecsSnapshot(podSpecSnapshot, reservationSpecSnapshot) =>
+        if (reservationSpecSnapshot.nonEmpty) {
+          // This should make the framework crash
+          throw new NotImplementedError("ReservationSpec support not yet implemented")
+        }
+        val newPodsSpecs: Map[PodId, PodSpec] = podSpecSnapshot.map { pod =>
+          pod.id -> pod
+        }(collection.breakOut)
+
+        val changedPodIds = podSpecs.keySet ++ newPodsSpecs.keySet
```
zen-dog: Is this correct? In `changed = existing + new` why are existing podSpecs part of the changed ones? Or ist because existing ones are considered removed later?

core/src/main/scala/com/mesosphere/usi/core/CachedPendingLaunch.scala
```diff
@@ -0,0 +1,26 @@
...
+case class CachedPendingLaunch(pendingLaunch: Set[PodId] = Set.empty) {
+  /**
+    * Maintains a cache of pods pending launch, so a full scan isn't required on every offer
+    *
+    * Must be called after podSpecs are updated, and response effects are applied.
+    *
+    * @param podIds
+    */
+  def update(frame: Frame, podIds: Set[PodId]): CachedPendingLaunch = {
+    var newPendingLaunch = pendingLaunch
```
jeschkies, line 14: What do you think builder `val builder = pendingLaunch.genericBuilder[PodId]` instead? Maybe it's to early to optimize it.

core/src/main/scala/com/mesosphere/usi/core/Frame.scala
```diff
@@ -0,0 +1,61 @@
+package com.mesosphere.usi.core
+import com.mesosphere.usi.core.models.{PodId, PodRecord, PodSpec}
+import com.mesosphere.usi.models._
+
+case class Frame(podSpecs: Map[PodId, PodSpec], podRecords: Map[PodId, PodRecord], podStatuses: Map[PodId, PodStatus]) {
```
jeschkies: Needs some scaladocs 😉

core/src/main/scala/com/mesosphere/usi/core/FrameEffects.scala
```diff
@@ -0,0 +1,31 @@
...
+      this
+    else if (this == FrameEffects.empty)
+      other
+    else
+      FrameEffects(other.reverseStateEvents ++ reverseStateEvents, other.reverseMesosCalls ++ reverseMesosCalls)
+  }
+
+  def withPodStatus(id: PodId, newStatus: Option[PodStatus]): FrameEffects =
+    withStateUpdated(PodStatusUpdated(id, newStatus))
+  def withPodRecord(id: PodId, newRecord: Option[PodRecord]): FrameEffects =
```
jeschkies: This needs some docs. Eg when `newRecord` is empty it means a deletion, right? Should we have a `withoutPodRecord(id)` as well?

core/src/main/scala/com/mesosphere/usi/core/Scheduler.scala
jeschkies: Yeah, please not too many caps. I once worked on an `EMLVWProcessorImplFactory`. Why not simply `StateEvents`?

core/src/main/scala/com/mesosphere/usi/core/Scheduler.scala
```diff
@@ -1,13 +1,65 @@
...
+ *      /             |  |                  |          |             |                         |\
+ *     /              |  +------------------+          +-------------+                         | \
+ *    /               |                                                                        |  \
+ *   |                +------------------------------------------------------------------------+  |
+ *    \                                                                                           |
+ *     \                               +----------------------+                                  /
+ *      \                              |                      |                                 /
+ *       \-----------------------------<        Mesos         <---------------------------------
+ *                                     |                      |
+ *                                     +----------------------+
```
jeschkies, line 43: This is really cool. Do you have the PlantUML for it?

core/src/main/scala/com/mesosphere/usi/core/SchedulerLogic.scala
jeschkies: Or `Frame.empty` similar to `FrameEffects.empty`?

core/src/main/scala/com/mesosphere/usi/core/SchedulerLogic.scala
```diff
@@ -0,0 +1,201 @@
...
+  * Container class responsible for keeping track of the frame and cache.
+  *
+  * As a general rule, business logic does not manipulate the USI state directly, but rather does so by returning
+  * effects.
+  */
+private[core] class SchedulerLogic {
+  private var frame: Frame = Frame(Map.empty, Map.empty, Map.empty)
+  private var cachedPendingLaunch = CachedPendingLaunch(Set.empty)
+
+  def processSpecEvent(msg: SpecEvent): FrameEffects = {
```
jeschkies: This will be just awesome to test.

core/src/main/scala/com/mesosphere/usi/core/FrameEffects.scala
```diff
@@ -0,0 +1,31 @@
+package com.mesosphere.usi.core
+
+import com.mesosphere.usi.core.models.{PodId, PodRecord}
+import com.mesosphere.usi.models._
+
+case class FrameEffects(reverseStateEvents: List[USIStateEvent] = Nil, reverseMesosCalls: List[Mesos.Call] = Nil) {
+  lazy val stateEvents = reverseStateEvents.reverse
```
zen-dog: Maybe just take a `Queue` instead of `List`s so there is no need to reverse?

core/src/main/scala/com/mesosphere/usi/core/SchedulerLogic.scala
```diff
@@ -0,0 +1,201 @@
...
+    * Process the modification of some podSpec.
+    *
+    * - If a podSpec is deleted:
+    *   - Also delete the podRecord.
+    *   - Prune a terminal / unknown task status.
+    * - If a podSpec is marked as terminal, then issue a kill.
+    *
+    * @return The effects of launching this offer
+    */
+  private def matchOffer(offer: Mesos.Event.Offer, pendingLaunchPodSpecs: Seq[PodSpec]): FrameEffects = {
```
jeschkies: I'm surprised this method is not receiving a frame. It feels the `pendingLaunchPodsSpecs` should be part of the frame.

core/src/main/scala/com/mesosphere/usi/core/SchedulerLogic.scala
```diff
@@ -0,0 +1,201 @@
...
+    val operations = pendingLaunchPodSpecs.iterator.flatMap { pod =>
+      taskIdsFor(pod).iterator.map { taskId =>
+        Mesos.Operation(Mesos.Launch(Mesos.TaskInfo(taskId)))
+      }
+    }.to[Seq]
+
+    pendingLaunchPodSpecs.foreach { pod =>
+      effects = effects.withPodRecord(pod.id, Some(PodRecord(pod.id, Instant.now(), offer.agentId)))
+    }
+    effects = effects.withMesosCall(Mesos.Call.Accept(offer.offerId, operations = operations))
```
jeschkies: Hm, I know you've said we should concentrate on the interfaces but maybe this is part of it.
```scala
pendingLaunchPodSpecs.foreach { pod =>
  effects += UpdatedPodRecord(...)
}
effetcs += MesosCall(...)
```

Was easier to comprehend. It took me two looks to understand why `effects` was not simply overriden. This really feels like a source of possible bugs. Even a mutable `Effects` object feels nicer such as an effects builder

core/src/main/scala/com/mesosphere/usi/core/SchedulerLogic.scala
```diff
@@ -0,0 +1,201 @@
...
+
+  /**
+    * We remove a task if it is not reachable and running, and it has no podSpec defined
+    *
+    * Should be called with the effects already applied for the specified podIds
+    *
+    * @param podIds podIds changed during the last frame
+    * @return
+    */
+  def pruneTaskStatuses(frame: Frame)(podIds: Set[PodId]): FrameEffects = {
```
jeschkies: Hm, it really feels we should return a `Frame` instead.

core/src/main/scala/com/mesosphere/usi/core/CachedPendingLaunch.scala
```diff
@@ -0,0 +1,26 @@
+package com.mesosphere.usi.core
+import com.mesosphere.usi.core.models.PodId
+
+case class CachedPendingLaunch(pendingLaunch: Set[PodId] = Set.empty) {
```
zen-dog, line 4: This looks like premature optimization to me. Is there a benchmark showing that filtering a list of pod specs actually takes long? How many pod specs do you need to see the difference and how big is it? Is it on the order of μs? ms?

core/src/main/scala/com/mesosphere/usi/models/USIStateEvent.scala
```diff
@@ -0,0 +1,24 @@
...
+import com.mesosphere.usi.core.models.{PodId, PodRecord, ReservationId}
+
+/**
+  * General trait for snapshot or updated events, used to replicate USI's evolving view of the state.
+  *
+  * These events are emitted by the USI Scheduler Logic in response to evolution of specification states and Mesos
+  * events. They are emitted so that a framework can replicate the state and react to changes in a way that best suits
+  * the framework's domain.
+  */
+sealed trait USIStateEvent
```
jeschkies:
```suggestion
sealed trait StateEvent
```
I don't think we would have something non USI related in `com.mesosphere.usi.models`.

core/src/main/scala/com/mesosphere/usi/models/package.scala
```diff
@@ -0,0 +1,45 @@
+package com.mesosphere.usi
+
+import com.mesosphere.usi.core.models._
+
+package models {
```
jeschkies: Any reason this is not part of `core-models`?

jeschkies (commented):
I really like it. Some things are not quite clear to me but the `Frame` feels like a good approach.

Any reason some models are not part of `core-models`? It feels like that;s a reason to reject.

core/src/main/scala/com/mesosphere/usi/core/SchedulerLogic.scala
zen-dog: Nicely readable logic 👏

core/src/main/scala/com/mesosphere/usi/core/SchedulerLogic.scala
```diff
@@ -0,0 +1,201 @@
...
+
+  def processSpecEvent(msg: SpecEvent): FrameEffects = {
+    val (frameWithPodSpecChanges, changedPodIds) = frame.applySpecEvent(msg)
+    frame = frameWithPodSpecChanges
+    val frameEffects = SchedulerLogic.computeNextStateForPods(frameWithPodSpecChanges)(changedPodIds)
+    applyAndUpdateCaches(frameEffects, changedPodIds)
+  }
+
+  /**
+    * Given a frame effects, apply the usi state change effects of the frame effects. Then, update the cache and perform
```
zen-dog: I had to read the first sentence three times and I still don't quite understand it 😊 Maybe a diagram would help?

core/src/main/scala/com/mesosphere/usi/core/SchedulerLogic.scala
```diff
@@ -0,0 +1,201 @@
...
+
+/**
+  * Container class responsible for keeping track of the frame and cache.
+  *
+  * As a general rule, business logic does not manipulate the USI state directly, but rather does so by returning
+  * effects.
+  */
+private[core] class SchedulerLogic {
+  private var frame: Frame = Frame(Map.empty, Map.empty, Map.empty)
+  private var cachedPendingLaunch = CachedPendingLaunch(Set.empty)
```
zen-dog, line 80: Why do we need the pending launches cached here like this? Why is not a `lazy val pendingLaunch...` within the above frame since frames are immutable?

core/src/main/scala/com/mesosphere/usi/core/SchedulerLogic.scala
```diff
@@ -0,0 +1,201 @@
...
+    * @param eventFrameEffects The effects from the incoming SpecEvent
+    * @param dirtyPodIds The podSpecs whose cache should be updated or checked for pruning
+    * @return The frame effects with additional housekeeping effects applied
+    */
+  private def applyAndUpdateCaches(eventFrameEffects: FrameEffects, dirtyPodIds: Set[PodId]): FrameEffects = {
+    frame = frame.applyStateEffects(eventFrameEffects)
+    this.cachedPendingLaunch = this.cachedPendingLaunch.update(frame, dirtyPodIds)
+
+    val pruneEffects = SchedulerLogic.pruneTaskStatuses(frame)(dirtyPodIds)
+    frame = frame.applyStateEffects(pruneEffects)
```
zen-dog: This "feels dirty", it's the t**hird time** you update the `private var frame` (starting in the `processSpecEvent ` method call chain) and only by the third time, it's a valid frame (meaning all the effects are there). I would expect to see only one `frame = frame.apply...` call that replaces the current frame with a new one. You can have intermediate frames but I would not update the current one with them. This should make the whole logic more "atomic" and the  `cachedPendingLaunch` could become a `lazy val` within the frame.

core/src/main/scala/com/mesosphere/usi/core/SchedulerLogic.scala
```diff
@@ -0,0 +1,201 @@
...
+  }
+
+  /**
+    * Process a Mesos event and update internal state.
+    *
+    * @param event
+    * @return The events describing state changes as Mesos call intents
+    */
+  def processMesosEvent(event: Mesos.Event): FrameEffects = {
+    val result = SchedulerLogic.handleMesosEvent(frame, cachedPendingLaunch.pendingLaunch)(event)
```
zen-dog: This split into `case class SchedulerLogic` and `object SchedulerLogic` is only necessary (to ease the testing) because `SchedulerLogic` holds the `frame`. If the frame would live outside then both could be united into a set of pure methods: _pass current frame and event -> return new frame and effects_. 
I believe @jeschkies suggested something similar and I support that.

zen-dog (dismissed):
I left a few comments at places which IMHO require some reevaluation (especially around `SchedulerLogic`) so let's talk about it. Overall, this is a great PR! I love the BidiFlow and the diagrams (keep them coming please).

core/src/main/scala/com/mesosphere/usi/core/SchedulerLogic.scala
zen-dog: ...and an even better solution is described in my [comment below](https://github.com/mesosphere/usi/pull/28/files#r253903419)

core/src/main/scala/com/mesosphere/usi/core/SchedulerLogic.scala
timcharper: Why do we need two what here? PodStatus and PodRecord?

Because the records are updated independently. A PodStatus may not have a PodRecord, and vice versa. When we launch a pod, it has no status yet, but we must have a record that we launched it. And, if we discover a status for which there is no record, then its a spurious task.

If we coupled task status with pod record then we would needlessly need to rewrite the podRecord every time a Mesos health check was received or whatnot.

core/src/main/scala/com/mesosphere/usi/core/SchedulerLogicGraph.scala
```diff
@@ -0,0 +1,109 @@
...
+
+  val mesosEventsInlet = Inlet[Mesos.Event]("mesos-events")
+  val specEventsInlet = Inlet[SpecEvent]("specs")
+  val effectsOutlet = Outlet[FrameEffects]("effects")
+  // Define the shape of this stage, which is SourceShape with the port we defined above
+  override val shape: FanInShape2[SpecEvent, Mesos.Event, FrameEffects] = new FanInShape2(specEventsInlet, mesosEventsInlet, effectsOutlet)
+
+  // This is where the actual (possibly stateful) logic will live
+  override def createLogic(inheritedAttributes: Attributes): GraphStageLogic = {
+    val schedulerLogic = new SchedulerLogic
```
zen-dog, line 54: This is where I would put current `Frame` and make `SchedulerLogic` pure

core/src/main/scala/com/mesosphere/usi/core/Scheduler.scala
timcharper: Remnants of comments from before renames :(

core/src/main/scala/com/mesosphere/usi/core/SchedulerLogic.scala
timcharper: sure

timcharper: [ ] - DOING

core/src/main/scala/com/mesosphere/usi/core/SchedulerLogicGraph.scala
zen-dog: ... or not ;)

timcharper: Added FrameWithEffects; we can shuffle things around further, later, if we need.

core/src/main/scala/com/mesosphere/usi/core/SchedulerLogic.scala
timcharper: Added FrameWithEffects abstraction which handles this more elegantly. Preference to have business logic only return effects if possible.

core/src/main/scala/com/mesosphere/usi/core/SchedulerLogic.scala
timcharper: Discussed in meeting.

core/src/main/scala/com/mesosphere/usi/core/SchedulerLogic.scala
timcharper: It should be simpler now with FrameWithEffects

core/src/main/scala/com/mesosphere/usi/models/package.scala
timcharper: Can we move it after?

core/src/main/scala/com/mesosphere/usi/models/USIStateEvent.scala
timcharper: ✅

core/src/main/scala/com/mesosphere/usi/core/SchedulerLogic.scala
timcharper: Preference for FrameEffects. See FrameWithEffects abstraction.

core/src/main/scala/com/mesosphere/usi/core/SchedulerLogic.scala
timcharper: I'll change it to a fold

core/src/main/scala/com/mesosphere/usi/core/FrameEffects.scala
timcharper: `withPodRecord` None handles it for now; I'll add docs

timcharper:
@jeschkies Not sure if I agree this should be rejected because models are in wrong path. This is a mock implementation and we still preserve the ability to change things in the future. If you feel strongly please add a commit and move them.

core/src/main/scala/com/mesosphere/usi/core/Frame.scala
timcharper: disagree

core/src/main/scala/com/mesosphere/usi/core/Frame.scala
timcharper: no longer applicable

core/src/main/scala/com/mesosphere/usi/core/CachedPendingLaunch.scala
timcharper: I prefer the immutable builder

core/src/main/scala/com/mesosphere/usi/core/CachedPendingLaunch.scala
timcharper: Discussed; well aware of premature optimization arguments. It is obvious that we will need caching mechanisms, and I wanted to work out how they will look early in the development of the SchedulerLogic. Recomputing pending launch states for every message is nonsense. I'd like to keep it.

core/src/main/scala/com/mesosphere/usi/core/SchedulerLogic.scala
timcharper: yeah, not opposed to doing that later.

core/src/main/scala/com/mesosphere/usi/models/SpecEvent.scala
timcharper: I prefer the Option

core/src/main/scala/com/mesosphere/usi/core/Scheduler.scala
timcharper: Let's tackle that later; I think this is pretty okay to read and the type system has our back here.

core/src/main/scala/com/mesosphere/usi/core/FakeMesos.scala
```diff
@@ -0,0 +1,23 @@
+package com.mesosphere.usi.core
+import akka.NotUsed
+import akka.stream.scaladsl.Flow
+import com.mesosphere.usi.core.models.AgentId
+import com.mesosphere.usi.models.Mesos
+
+object FakeMesos {
```
jeschkies, line 6: Should we move this to `test-utils`?

core/src/main/scala/com/mesosphere/usi/core/SchedulerLogic.scala
```diff
@@ -0,0 +1,237 @@
...
+import com.mesosphere.usi.core.models._
+import com.mesosphere.usi.models._
+
+case class FrameWithEffects(frame: Frame, effects: FrameEffects, dirtyPodIds: Set[PodId]) {
+  def applyEffects(newEffects: FrameEffects): FrameWithEffects = {
+    // TODO - we need to handle status snapshots and create a mechanism to signal that all cache should be recomputed
+    val newDirty = dirtyPodIds ++ newEffects.reverseStateEvents.iterator.collect {
+      case podEvent: PodStateEvent => podEvent.id
+    }
+    copy(frame = frame.applyStateEffects(effects), dirtyPodIds = newDirty, effects = effects ++ newEffects)
```
jeschkies: Shouldn't the original effects be dropped since they are already applied?

jeschkies (commented):
I like the new `FrameWithEffects` but am unsure about `FrameWithEffects.applyEffects`. Maybe I misunderstand something.

core/src/main/scala/com/mesosphere/usi/models/SpecEvent.scala
zen-dog: But, but... `PodSpecDeleted` is so much clearer and easier to read.

zen-dog: also, think about logging. grepping for all deleted podSpecs is easier with `PodSpecDeleted(...)` then it is with `PodSpecUpdated(???, None)`

core/src/main/scala/com/mesosphere/usi/models/package.scala
zen-dog: I moved it. Don't forget to pull

core/src/main/scala/com/mesosphere/usi/core/FakeMesos.scala
zen-dog: I don't think so.

core/src/main/scala/com/mesosphere/usi/core/SchedulerLogic.scala
zen-dog: Good question. However, of these effects, only `stateEvents` (and not e.g. `mesosCalls`) are applied. So I guess we should keep the later and remove the former?

core/src/main/scala/com/mesosphere/usi/core/SchedulerLogic.scala
```diff
@@ -0,0 +1,235 @@
...
+
+import com.mesosphere.usi.core.models._
+
+case class FrameWithEffects(frame: Frame, effects: FrameEffects, dirtyPodIds: Set[PodId]) {
+  def applyEffects(newEffects: FrameEffects): FrameWithEffects = {
+    // TODO - we need to handle status snapshots and create a mechanism to signal that all cache should be recomputed
+    val newDirty = dirtyPodIds ++ newEffects.reverseStateEvents.iterator.collect {
+      case podEvent: PodStateEvent => podEvent.id
+    }
+    copy(frame = frame.applyStateEffects(effects), dirtyPodIds = newDirty, effects = effects ++ newEffects)
```
zen-dog:
```suggestion
    copy(frame = frame.applyStateEffects(newEffects), dirtyPodIds = newDirty, effects = effects ++ newEffects)
```
This above looks like a bug to me.

zen-dog: Also, I believe we conflate two things in _Effects_:
1) things that change the Frame
2) side effects (like Mesos calls)
Could we call the 2) - `Actions` to make their purpose clearer?

core/src/main/scala/com/mesosphere/usi/core/SchedulerLogic.scala
timcharper: It's correct. We clean up terminal or unreachable statuses if the podSpec is removed, and the status is terminal or unreachable.

core/src/main/scala/com/mesosphere/usi/core/SchedulerLogicGraph.scala
timcharper: Lets do it later. Mock interface.

core/src/main/scala/com/mesosphere/usi/models/Mesos.scala
timcharper: Let's integrate the mesos client in the task specifically designated as such

core/src/main/scala/com/mesosphere/usi/core/SchedulerLogic.scala
timcharper: I could go with intents, perhaps? Let's talk

core/src/main/scala/com/mesosphere/usi/core/SchedulerLogic.scala
```diff
@@ -0,0 +1,235 @@
...
+  private var cachedPendingLaunch = CachedPendingLaunch(Set.empty)
+
+  def processSpecEvent(msg: SpecEvent): FrameEffects = {
+    handleFrame { frameWithEffects =>
+      frameWithEffects
+        .applySpecEvent(msg)
+        .process { f =>
+          SchedulerLogic.computeNextStateForPods(f.frame)(f.dirtyPodIds)
+        }
+    }
```
jeschkies, line 89: Here's my suggestion to flatten it

```scala
val (frameWithSpecEventsApplied, dirityPods) = applySpecEvent(currentFrame, msg)
val FrameIntents(stateUpdates, calls) = SchedulerLogic.computeNextStateForPods(frameWithSpecEventsApplied)(dirtyPodIds)

val finalFrame = applyIntents(stateUpdates)
currentFrame = finalFrame
FrameIntents(stateUpdates, calls)
```
The names are a little verbose.

timcharper: I really prefer the method chaining approach; it gets rid of all of these assignments and precludes us from accidentally returning the wrong version of the frame; further, the chained abstraction separates out the concern of tracking dirty podIds, which I really like. Can you help me understand why you think this is better?

core/src/main/scala/com/mesosphere/usi/core/SchedulerLogic.scala
timcharper: We have to aggregate the events so we can publish them downstream

core/src/main/scala/com/mesosphere/usi/core/SchedulerLogic.scala
timcharper: `pendingLaunchPodsSpecs` is derived state from the frame, and is incrementally recomputed (a cache).

core/src/main/scala/com/mesosphere/usi/core/Scheduler.scala
timcharper: Thanks! No, PlantUML cannot generate graphs like this. I drew it with Emacs artist-mode

PR #29, opened 2019-01-31
Add an example `DummyFramework`

zen-dog:
Summary:
Moved `UselessFramework` class from the `mesos-client` to the `examples` folder and renamed it to `MesosClientExampleFramework`


examples/dummy-framework/README.md
```diff
@@ -0,0 +1,8 @@
+# Dummy Framework
+
+Run a dummy framework that:
+ - uses only the raw mesos-client
+ - successfully subscribes to Mesos master
+ - declines all offers since its well, pretty dummy
+ 
+ Not much, but shows the basic idea. Good to test against local Mesos.
```
meichstedt:
```suggestion
Not much, but shows the basic idea. Good to test against local Mesos.
Note that we neither suggest, nor support, using the mesos-client on its own without the other components within this repository.
```

meichstedt (approved):

examples/dummy-framework/build.gradle
```diff
@@ -0,0 +1,5 @@
+dependencies {
+    compile project(':mesos-client')
+    implementation "org.apache.mesos:mesos:$mesosVersion"
```
jeschkies: Shouldn't this come with `mesos-client`. @kensipe ?

kensipe: agree and Alek resolved it

kensipe (approved):

PR #31, opened 2019-01-31
 Archive JUnit XML reports. 

jeschkies:
Summary:
This will upload the JUnit test reports alongside the HTML reports. The folder structure changes a little but on S3. The human readable reports will be in `<owner>/<repo>/<branch>/<build>/reports/tests` and the JUnit files in `<owner>/<repo>/<branch>/<build>/tests`.

This patch will also migrate the S3 upload task to the AWS SDK 2.3.x and uses an asynchronous upload client.

JIRA issue: DCOS-47520

### DIFF
--- a/buildSrc/src/main/scala/mesosphere/gradle/aws/S3Upload.scala
+++ b/buildSrc/src/main/scala/mesosphere/gradle/aws/S3Upload.scala
@@ -2,6 +2,7 @@ package mesosphere.gradle.aws
 
 import java.util.concurrent.CompletableFuture
 
+import org.apache.commons.io.FilenameUtils
 import com.typesafe.scalalogging.StrictLogging
 import org.gradle.api.DefaultTask
 import org.gradle.api.file.FileTree
@@ -9,7 +10,7 @@ import org.gradle.api.tasks.TaskAction
 import software.amazon.awssdk.core.async.AsyncRequestBody
 import software.amazon.awssdk.regions.Region
 import software.amazon.awssdk.services.s3.S3AsyncClient
-import software.amazon.awssdk.services.s3.model.{PutObjectResponse, PutObjectRequest}
+import software.amazon.awssdk.services.s3.model.{PutObjectRequest, PutObjectResponse}
 
 import scala.beans.BeanProperty
 
@@ -34,7 +35,11 @@ class S3Upload extends DefaultTask with StrictLogging {
       if (!fileDetails.isDirectory) {
         val file = fileDetails.getRelativePath()
         val key = s"$prefix/$file"
-        val request = PutObjectRequest.builder().bucket(bucket).key(key).contentType("test/html").build()
+        val request = PutObjectRequest.builder()
+          .bucket(bucket)
+          .key(key)
+          .contentType(ContentType.of(fileDetails.getName))
+          .build()
         val body = AsyncRequestBody.fromFile(fileDetails.getFile)
         pendingUploads = pendingUploads :+ s3.putObject(request, body)
       }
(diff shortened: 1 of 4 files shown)


jeschkies:
Note: Conent-type needs to be set.

buildSrc/src/main/scala/mesosphere/gradle/aws/S3Upload.scala
```diff
@@ -1,53 +1,50 @@
...
-      logger.info(s"${progress.getPercentTransferred()} % ${upload.getState()}")
-      Thread.sleep(2000)
+    logger.info(s"Uploading files to $bucket with prefix $prefix")
+
+    var pendingUploads = Vector.empty[CompletableFuture[PutObjectResponse]]
+    source.visit { fileDetails =>
+      if (!fileDetails.isDirectory) {
+        val file = fileDetails.getRelativePath()
+        val key = s"$prefix/$file"
+        val request = PutObjectRequest.builder().bucket(bucket).key(key).contentType("test/plain").build()
```
jeschkies: Will infer content type from extension in next commit.

zen-dog:
Should we tackle archiving sandboxes here too @jeschkies ?

jeschkies:
> Should we tackle archiving sandboxes here too @jeschkies ?
Go question. I'd say not in this diff. As long as we'll put them in `./build/sandboxes` we can easily add them. If not we have to do some little design work or define a separate upload task which probably makes more sense.

I'd merge this first and then follow up once the need arises.

buildSrc/src/main/scala/mesosphere/gradle/aws/S3Upload.scala
```diff
@@ -1,53 +1,66 @@
...
 class S3Upload extends DefaultTask with StrictLogging {
 
-  val credentialProviderChain = new DefaultAWSCredentialsProviderChain()
+  lazy val s3: S3AsyncClient = S3AsyncClient.builder().region(Region.of(region)).build()
 
   @BeanProperty var region: String = "us-west-2"
 
   @BeanProperty var bucket: String = ""
 
   @BeanProperty var prefix: String = ""
```
alenkacz, line 44: what kind of prefix is this? The name is very generic and the default value is "" so 🤔 maybe add some docs or rename?

ci.gradle
```diff
@@ -6,7 +6,7 @@ task testReport(type: TestReport) {
     description = "Aggregate all test results from subprojects into one test report."
     group = "CI"
 
-    destinationDir = file("$buildDir/reports/allTests")
+    destinationDir = file("$buildDir/reports/tests")
```
alenkacz: could we use the same path here and on l.27. Here you do `$buildDir` there it's `rootProject.buildDir`. I believe it will end up being the same path but I would still prefer these two to be in sync

ci.gradle
```diff
@@ -24,7 +24,10 @@ task uploadTestReports(type: S3Upload) {
     region = "us-west-2"
     bucket = "usi-builds"
     prefix = "$repoSlug/$branch/$buildNumber"
-    source = testReport.destinationDir
+    source = fileTree(rootProject.buildDir) {
+        include 'reports/tests/**'
+        include 'tests/**'
```
alenkacz: what's `tests/**` 🤔 ? Maybe add some comment?

alenkacz (approved):
I think we should add more comments/docs to some part of the CI build file, but LGTM

buildSrc/src/main/scala/mesosphere/gradle/aws/S3Upload.scala
jeschkies: Good point. The AWS SDK 1.x hat this. I'll add docs.

zen-dog:
Wdyt about calling `tests` in `<owner>/<repo>/<branch>/<build>/tests` something more generic e.g. `artifacts`? I'd like for sandboxes later to be in the same folder with Junit output, logs and everything else that might belong to that particular build run.

zen-dog (approved):
🚢

PR #32, opened 2019-02-01
Scalafmt from one config, more formatting

alenkacz:
The main feature in this diff is:
- now we require only one scalafmt config (!) not one copy-pasted into every module (and yes, we have a lot of these 🙈 )
- as a part of feedback I dropped `apply plugin: 'cz.alenkacz.gradle.scalafmt'` from last PR, which means that all projects did not have that task applied only the root one so lots of fixes in that PR were not applied, the way it is now is correct
- unfortunately `includeNoParensInSelectChains` is scalafmt 2.0 feature and there's only RC now, so we cannot use that. I am so sorry but for now we will have to live with 
```
Call
      .newBuilder()
```

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -4,7 +4,7 @@ import java.nio.file.Paths
 plugins {
     id 'idea'
     id 'org.ajoberstar.git-publish' version '2.0.0'
-    id 'cz.alenkacz.gradle.scalafmt' version '1.7.0'
+    id 'cz.alenkacz.gradle.scalafmt' version '1.7.2'
 }
 
 ext {
@@ -19,6 +19,7 @@ allprojects {
     version = '0.1'
 
     apply plugin: "scala"
+    apply plugin: 'cz.alenkacz.gradle.scalafmt'
 
     sourceCompatibility = 1.8
     targetCompatibility = 1.8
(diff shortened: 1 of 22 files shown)


build.gradle
```diff
@@ -19,6 +19,7 @@ allprojects {
     version = '0.1'
 
     apply plugin: "scala"
+    apply plugin: 'cz.alenkacz.gradle.scalafmt'
```
alenkacz, line 22: this is correct! We have to apply that plugin to every module so the tasks are available in that module

mesos-client/src/main/scala/com/mesosphere/mesos/client/MesosCalls.scala
```diff
@@ -42,7 +43,8 @@ class MesosCalls(frameworkId: FrameworkID) {
     * http://mesos.apache.org/documentation/latest/scheduler-http-api/#accept
     */
   def newAccept(accepts: Accept): Call = {
-    Call.newBuilder()
+    Call
```
alenkacz, line 46: sorry, this will be fixed in scalafmt 2.0 :(

build.gradle
jeschkies: Thanks!

jeschkies (approved):

meln1k (approved):
YOLO

PR #33, opened 2019-02-04
Build with JDK 11 only.

jeschkies:


.travis.yml
```diff
@@ -1,4 +1,7 @@
 sudo: required
+jdk:
+  - openjdk8
+  - openjdk11
```
jeschkies, line 3: @kensipe, is there a way to configure `build.gradle` as well?

kaiwalyajoshi: There's two fields with the Java Plugin:
`sourceCompatibility` and `targetCompatibility` [from here](https://docs.gradle.org/current/userguide/java_plugin.html#other_convention_properties).

It should correspond to [this line](https://github.com/mesosphere/marathon/blob/43cee2540108b67f4cf63946271c90e09bea563e/build.sbt#L100) from Marathon's SBT build.

From our conversations so far, `sourceCompatibility` is 1.8 but `targetCompatibility` can be 1.11.

jeschkies:
Hm, I'm not sure if we really want a matrix build. Reporting errors is not trivial.

zen-dog:
> Hm, I'm not sure if we really want a matrix build. Reporting errors is not trivial.

I second that - Java8 support has almost ended so I guess there is no point in keeping the compatibility.

.travis.yml
kensipe: @jeschkies I'm not sure what you are asking here...   we could make compatibility conditional on task run or ENV.

kensipe:
there are plenty project which provide support back into a version that is not supported... especially in the open source space.    The question is do we want to be one of them... or do we push things forward.   I would strongly support java 11.

.travis.yml
```diff
@@ -1,4 +1,7 @@
...
 language: scala
 scala:
    - 2.12.2
```
kensipe: our scala version is `2.12.7`

jeschkies:
> there are plenty project which provide support back into a version that is not supported... especially in the open source space. The question is do we want to be one of them... or do we push things forward. I would strongly support java 11.

I couldn't agree more. We could have some dependencies that don't support Java 11 but that seems unlikely to me.

kensipe (changes requested):

.travis.yml
jeschkies: Good catch. Should we use `TRAVIS_SCALA_VERSION` in our `build.gradle`?

jeschkies:
@kensipe could you take another look?

build.gradle
```diff
@@ -21,8 +21,8 @@ allprojects {
     apply plugin: 'cz.alenkacz.gradle.scalafmt'
 
     sourceCompatibility = 1.8
```
alenkacz: I think we should have sourceCompatibility set to 1.11 as well since we're aiming for 1.11

alenkacz (approved):
Thanks! And 👍 for targeting 1.11!

kensipe (approved):

build.gradle
```diff
@@ -92,7 +92,7 @@ scaladoc {
 
 idea {
   project {
-    jdkName = '1.8'
+    jdkName = '1.11'
     languageLevel = '1.8'
```
kensipe: are we switching this to `1.11` as well ?

timcharper (approved):
Second what Alena said

PR #36, opened 2019-02-07
Add structured logging to hello-world module and MDC to core module

takirala:
Sample log line:
```
{"ts":"2019-02-07T15:39:22.568","thread":"main","level":"INFO","logger":"com.mesosphere.usi.core.Scheduler$","msg":"Connecting to mesosHostName=asd","caller_method_name":"connect","caller_file_name":"Scheduler.scala","caller_line_number":51,"mesosHostName":"asd","@version":1}
{"ts":"2019-02-07T15:39:23.183","thread":"main","level":"INFO","logger":"com.mesosphere.usi.helloworld.Main$","msg":"hello main","caller_method_name":"main","caller_file_name":"Main.scala","caller_line_number":12,"@version":1}
```

- Supports the usual tcp, udp or ssl appenders - i have tested tcp and console only. we can add more in future as need arises.
- Preferred the `logback.groovy` over the usual `logback.xml` as it involved more customization and groovy syntax made it look readable.
- I have used [Standard Fields](https://github.com/logstash/logstash-logback-encoder#standard-fields), [Caller Info](https://github.com/logstash/logstash-logback-encoder#caller-info-fields) and [Structured Arguments](https://github.com/logstash/logstash-logback-encoder#event-specific-custom-fields) in a sample log line. [MDC](https://github.com/logstash/logstash-logback-encoder#mdc-fields), [Context fields](https://github.com/logstash/logstash-logback-encoder#context-fields) and [Custom fields](https://github.com/logstash/logstash-logback-encoder#custom-fields) are supported too. It may not be trivial to use MDC ([although here is an example](http://code.hootsuite.com/logging-contextual-info-in-an-asynchronous-scala-application/)) but we can improve this in future.

Going forward,
- we can add more appenders (udp, ssl) when required.
- use MDC with `LoggerTakingImplicit` [as documented here](https://github.com/lightbend/scala-logging/tree/9fa1ea0681f72cb5001d378e0c6aaa9e1cf15491#using-scala-logging) when required.

Alternatives considered:
- https://github.com/jacek99/structlog4j : I intended to explore this option in case logstash's logback encoder doesn't fill our requirements. But I found logstash library to be adequate this alternative did not see any commits in past 2 years. So I chose logstash library.


build.gradle
```diff
@@ -41,6 +41,9 @@ allprojects {
         // Use Scala 2.12 in our library project
         implementation 'org.scala-lang:scala-library:2.12.7'
         implementation "com.typesafe.scala-logging:scala-logging_$scalaVersion:3.9.2"
+        implementation group: 'ch.qos.logback', name: 'logback-classic', version: '1.2.3'
```
jeschkies: I'm not really fond of prescribing a logging backend in a library. I feel we should stick to slf4j as much as possible.

core/src/main/scala/com/mesosphere/usi/core/Scheduler.scala
```diff
@@ -42,10 +43,12 @@ import scala.concurrent.Future
  *                                     |                      |
  *                                     +----------------------+
  */
-object Scheduler {
+object Scheduler extends StrictLogging {
+
   case class MesosConnection(mesosHostName: String)
 
   def connect(mesosHostName: String): Future[(MesosConnection, Flow[SpecEvent, StateEvent, NotUsed])] = {
+    logger.info(s"Connecting to {}", kv("mesosHostName", mesosHostName))
```
jeschkies: @alenkacz and I had a brief discussion on the format. Here is the [slack thread](https://mesosphere.slack.com/archives/G7Y5Q786S/p1548922017700400). This is what I really liked
> @kjeschkies yeah, I am open to pretty much anything. What I like about the fluent approach is that you can basically inject some metadata into logger with `logger.with` and then pass that logger into some method (so you don't have to have the whole metadata at hand when you're printing your message)

So this should be something like
```scala
logger.with("meosHostName" -> mesosHostName).info("Connecting")
```

which would output something like
```json
{"class": "Scheduler", "context": {"mesosHostName": "localhost"}, "message": "conntecting"}
```

What would the output look like in your example?

jeschkies (changes requested):
Thanks for taking this over. I think we have to iterate a little 🙂

core/src/main/scala/com/mesosphere/usi/core/Scheduler.scala
alenkacz: yeah sorry, I did not finish writing the doc for that, I thought I'll do it when I come back from my vacation. But most of the ideas are in that thread (plus a little bit in my head). I can write it down since there are several options how to do that... (but I also like the one @jeschkies is describing, that's the one I saw in the hashicorps' golang library for logging).

build.gradle
alenkacz: yeah I am also not 100% sure about that at the same time I don't feel strongly about having this here. Is it possible to override it as a user of a library? At least by providing a different logback config?

But I agree that libraries should be as lightweight in terms of dependencies as possible :)

takirala:
@jeschkies @alenkacz i have done a bit more investigation and here are my observations on two main concerns:
- Using a logging backend in our "library"
	- If we decide to use scalalogging, my understanding is that it requires a logging backend. See its [prerequisites](https://github.com/lightbend/scala-logging#prerequisites). I have tried with `slf4j-simple` instead of logback and it compiled just fine. i think we should be able to swap out the logging backend in the future if we want to.
    - By using gradle's `implementation`, we should be able to separate what is part of our `api` and what is part of our "implementation" only [as shown here](https://docs.gradle.org/current/userguide/java_library_plugin.html#sec:java_library_separation) i still need to verify this but I think one should be able to use USI as a gradle dependency while being able to use whatever logging library they wish to use. Even if they are using logback (perhaps, a different version) it should be able to conflict resolve.

- Logger composition
  - First and foremost, it is incredibly awesome if we can do `logger.with().with()...` so that we can compose metadata with loggers and pass them around. This makes tracing a lot easier.
  - My initial research led me to believe that there is no "one stop shop" to accomplish all of this. One has to make use of `LoggerTakingImplicit` or `MDC` to enrich the context. However all my research led me to believe that we cannot "compose" a logger to pass it around. We can only "compose" logger, well to log statements. This is mainly because, if we are composing the logger with some metadata then that means we are inducing an idea of "state" to the logger. This is not possible (or inefficient if it is possible). Also we cannot instantiate loggers on the fly because it would increase the size of the `Map` they are stored in as scope of their destruction may not always be defined clearly.
   - Still, all the systems that achieve this use some form of MDC with implicit/explicit concurrency control. The easiest way i have attempted to do this is [via this commit that should be cleaned up](https://github.com/mesosphere/usi/pull/36/commits/f20a695a7d2c4dea36e8e426818aaabec71881d9). The basic idea is that we have to maintain the "state" in an ephemeral manner by composing it and passing it around. I _think_ we should be able to clean this up a little by using `macro`s and/or `quasiquotes`, Here are sample log statements from the above commit:
```

{
  "ts": "2019-02-12T00:39:56.526",
  "thread": "main",
  "level": "INFO",
  "logger": "com.mesosphere.usi.core.Scheduler$",
  "msg": "kasdad asd=asd",
  "asd": "asd",
  "@version": 1,
  "tag2": "value1",
  "tag3": "value1"
}
{
  "ts": "2019-02-12T00:39:57.253",
  "thread": "main",
  "level": "INFO",
  "logger": "com.mesosphere.usi.core.Scheduler$",
  "msg": "Connecting to mesosHostName=asd",
  "mesosHostName": "asd",
  "@version": 1,
  "tag2": "value1",
  "tag3": "value1",
  "tag1": "value1"
}
{
  "ts": "2019-02-12T00:39:57.575",
  "thread": "main",
  "level": "INFO",
  "logger": "com.mesosphere.usi.helloworld.Main$",
  "msg": "hello main",
  "@version": 1,
  "hello": "world"
}
{
  "ts": "2019-02-12T00:39:57.575",
  "thread": "main",
  "level": "INFO",
  "logger": "com.mesosphere.usi.helloworld.Main$",
  "msg": "hello test",
  "@version": 1
}
```
we can discuss offline / i can create a google doc if you think it would make the discussion easier to follow.

takirala:
- Moved logging backends to `hello-world` from the library
- Added few abstractions to make logger composition easy. we can add more as our project grows
- Here is a sample log:
  `{"ts":"2019-02-12T21:40:42.403","thread":"main","level":"INFO","logger":"com.mesosphere.usi.core.Scheduler$","msg":"Connecting to localhost","mesosHostName":"localhost","ver":1}
` 
  ( Note that we can put all the `MDC` fields in to a nested json if we want to - for now, i have put them at top level )
- I have been trying to build abstractions so as to NOT redefine/extend the `.info` or `.error` methods (so that the caller context is not interfered)

core/src/main/scala/com/mesosphere/package.scala
```diff
@@ -0,0 +1,51 @@
+package com
+
+import com.typesafe.scalalogging.CanLog
+import com.typesafe.scalalogging.LoggerTakingImplicit
+import org.slf4j.MDC
+import scala.language.implicitConversions
+
+package object mesosphere {
```
jeschkies, line 8: Please add documentation. This is not easy to understand.

core/src/main/scala/com/mesosphere/package.scala
```diff
@@ -0,0 +1,51 @@
+package com
```
jeschkies, line 1: Should this be really part of core? I would not be surprised if the client would use this as well.

core/src/main/scala/com/mesosphere/usi/core/Scheduler.scala
```diff
@@ -42,10 +44,15 @@ import scala.concurrent.Future
...
-object Scheduler {
+object Scheduler extends LazyLogging {
+
+  val log = Logger.takingImplicit[KvArgs](getClass)
+
   case class MesosConnection(mesosHostName: String)
 
   def connect(mesosHostName: String): Future[(MesosConnection, Flow[SpecEvent, StateEvent, NotUsed])] = {
+    val composedLogger = log.ctx(("mesosHostName", mesosHostName))
+    composedLogger.info(s"Connecting to $mesosHostName")(composedLogger.getCtx)
```
jeschkies: I missed some discussion but why is `composedLogger.getCtx` required here? Isn't the context only what we defined above?

core/src/main/scala/com/mesosphere/package.scala
```diff
@@ -0,0 +1,51 @@
...
+    override def afterLog(a: KvArgs): Unit = a.foreach(x => MDC.remove(x._1))
+  }
+
+  // This is needed so that we can call logger methods directly on the tuple.
+  implicit def tupleToLogger(
+    t : (LoggerTakingImplicit[KvArgs], KvArgs)
+  ) : LoggerTakingImplicit[KvArgs] = t._1
+
+  implicit class LoggerImplicitOps(val logger: LoggerTakingImplicit[KvArgs]) extends AnyVal {
+    def ctx(tag: KvArgs): (LoggerTakingImplicit[KvArgs], KvArgs) = (logger, tag)
```
jeschkies:
```suggestion
     def context(tag: KvArgs): (LoggerTakingImplicit[KvArgs], KvArgs) = (logger, tag)
```

We don't save much with `ctx` and `context` is easier to read. What do you think?

core/src/main/scala/com/mesosphere/package.scala
```diff
@@ -0,0 +1,51 @@
...
+  }
+
+  implicit final class LoggerImplicitConverter(
+    val self: (LoggerTakingImplicit[KvArgs], KvArgs)
+  ) extends AnyVal {
+
+    implicit def getCtx : KvArgs = self._2
+
+    def ctx(newContext: KvArgs): (LoggerTakingImplicit[KvArgs], KvArgs) = {
```
jeschkies: I'm a little confused. This returns a tuple but below we have `val composedLogger = log.ctx(("mesosHostName", mesosHostName))`. What am I missing?

jeschkies (commented):
I have a hard time understanding `core/src/main/scala/com/mesosphere/package.scala`. Could you add some docs and write the names out?

core/src/main/scala/com/mesosphere/usi/core/Scheduler.scala
takirala: when calling the [`log.info` method](https://github.com/lightbend/scala-logging/blob/master/src/main/scala/com/typesafe/scalalogging/LoggerTakingImplicit.scala#L44), we have pass an implicit parameter for the context. I tried to do this in couple ways:
1. First, i tried to redefine the `log.info` method, but this was interfering with caller context (Such as the method name and line number from where the logger statement was made). 
2. Then I tried to define the implicit object for context, but did not feel this to be very readable. Also there may be multiple implicits in scope and felt it would not be readable.
After trying above two, i settled for above style of explicitly passing the implicit context.

core/src/main/scala/com/mesosphere/package.scala
takirala: Used `ctx` for the getter and `withCtx` for the "setter" (not really a setter, but still..)

core/src/main/scala/com/mesosphere/package.scala
takirala: Tbh, I have put it here cuz I am not really sure where else to put this. More than happy to accept suggestions on a better home.

core/src/main/scala/com/mesosphere/package.scala
takirala: Added docs to clarify couple details on the abstractions being made. The implicits defined in the object make it easy to do few things:

- `tupleToLogger` helps call the `logger.info`, `logger.error` etc methods directly on the tuple object.
- methods in `LoggerImplicitConverter` and `LoggerImplicitOps` enables easy chaining such as `log.withCtx(..).withCtx(...).withCtx(...)` irrespecitve of `log` being a `LoggerTakingImplicit` or a tuple of `(LoggerTakingImplicit, KvArgs)`

build.gradle
```diff
@@ -52,7 +52,6 @@ allprojects {
         testRuntimeOnly "org.scala-lang.modules:scala-xml_$scalaVersion:1.1.1"
 
         // Logging
-        compile group: 'com.typesafe.scala-logging', name: "scala-logging_$scalaVersion", version: '3.9.2'
```
alenkacz, line 55: how come you removed it but we're still using it in `package.scala`?

core/src/main/scala/com/mesosphere/package.scala
alenkacz: Let's not create another module :) I am fine with this being in core for now :)

core/src/main/scala/com/mesosphere/package.scala
```diff
@@ -0,0 +1,78 @@
...
+
+  type KvArgs = Traversable[(String, Any)]
+
+  /**
+   * By default, we use an empty context for all loggers. This is needed where we have a class/object
+   * level implicit logger but it may not always be needed/required to have an implicit [[KvArgs]] in scope.
+   */
+  implicit val emptyKvArgs : KvArgs = Traversable()
+
+  implicit def tupleToTraversable(kv: (String, Any)): KvArgs = Traversable(kv)
```
alenkacz: why do we need this implicit? 🤔

core/src/main/scala/com/mesosphere/package.scala
```diff
@@ -0,0 +1,78 @@
+package com
+
+import com.typesafe.scalalogging.CanLog
+import com.typesafe.scalalogging.LoggerTakingImplicit
+import org.slf4j.MDC
+import scala.language.implicitConversions
+
+package object mesosphere {
```
alenkacz, line 8: would be nice to have some high level docs desacribing on how to use this logger (basically what's inside `Scheduler`)

core/src/main/scala/com/mesosphere/package.scala
```diff
@@ -0,0 +1,78 @@
...
+        MDC.put(elem._1, elem._2.toString)
+      }
+      originalMsg
+    }
+
+    override def afterLog(a: KvArgs): Unit = a.foreach(x => MDC.remove(x._1))
+  }
+
+  /**
+   * Enable calling logger methods directly on the tuple object.
```
alenkacz: could you show it on a code example in the javadocs? Showing how it would look like without this and with this maybe? Thanks

core/src/main/scala/com/mesosphere/package.scala
```diff
@@ -0,0 +1,78 @@
...
+   */
+  implicit def tupleToLogger(
+    t : (LoggerTakingImplicit[KvArgs], KvArgs)
+  ) : LoggerTakingImplicit[KvArgs] = t._1
+
+  /**
+   * Used to add context parameters to a simple [[LoggerTakingImplicit]] instance
+   */
+  implicit class LoggerImplicitOps(val logger: LoggerTakingImplicit[KvArgs]) extends AnyVal {
+    def withCtx(key: String, value: Any): (LoggerTakingImplicit[KvArgs], KvArgs) = (logger, (key, value))
```
alenkacz: I would vote for full `withContext` but no strong opinion here

core/src/main/scala/com/mesosphere/package.scala
```diff
@@ -0,0 +1,78 @@
...
+    def withCtx(tag: KvArgs): (LoggerTakingImplicit[KvArgs], KvArgs) = (logger, tag)
+  }
+
+  /**
+   * Used to add context parameters to a tuple containing [[LoggerTakingImplicit]] and [[KvArgs]] instance
+   */
+  implicit final class LoggerImplicitConverter(
```
alenkacz: can we please get rid of _1 and _2 in all places and replace with meaningful names? :)

core/src/main/scala/com/mesosphere/usi/core/Scheduler.scala
```diff
@@ -43,9 +44,14 @@ import scala.concurrent.Future
  *                                     +----------------------+
  */
 object Scheduler {
+
+  private val log = Logger.takingImplicit[KvArgs](getClass)
```
alenkacz: do we have to have the `takingImplicit` there? I would go for `withContext` for example?

alenkacz (changes requested):
Very good job, I like it 👏 

I have just few minor polishing things and then I'll be ready to approve!

core/src/main/scala/com/mesosphere/usi/core/Scheduler.scala
```diff
@@ -43,9 +44,14 @@ import scala.concurrent.Future
...
+
   case class MesosConnection(mesosHostName: String)
 
   def connect(mesosHostName: String): Future[(MesosConnection, Flow[SpecEvent, StateEvent, NotUsed])] = {
+    val composedLogger = log.withCtx("mesosHostName", mesosHostName)
+    composedLogger.info(s"Connecting to $mesosHostName")(composedLogger.ctx)
```
jeschkies: Why do we have to pass `composedLogger.ctx`? Shouldn't the logger have its context already?

core/src/main/scala/com/mesosphere/package.scala
```diff
@@ -0,0 +1,78 @@
...
+
+  /**
+   * By default, we use an empty context for all loggers. This is needed where we have a class/object
+   * level implicit logger but it may not always be needed/required to have an implicit [[KvArgs]] in scope.
+   */
+  implicit val emptyKvArgs : KvArgs = Traversable()
+
+  implicit def tupleToTraversable(kv: (String, Any)): KvArgs = Traversable(kv)
+
+  implicit final class KvArgsOps(val self : KvArgs) extends AnyVal {
```
jeschkies: Any reason this is not a case class? Shouldn't `and` return a copy?

core/src/main/scala/com/mesosphere/package.scala
jeschkies: I don't quite understand the return type. In the example we have

```scala
val composedLogger = log.withCtx("mesosHostName", mesosHostName)
composedLogger.info(s"Connecting to $mesosHostName")(composedLogger.ctx)
```

I assume `composedLogger` has the type `(LoggerTakingImplicit[KvArgs], KvArgs)`. How does `composedLogger.info(s"Connecting to $mesosHostName")(composedLogger.ctx)` work then?

jeschkies (changes requested):
I'm sorry. I have big troubles understanding all the implicits in the `mesosphere` package.

build.gradle
takirala: this was a typo, the library is already added (on line 45)

core/src/main/scala/com/mesosphere/package.scala
takirala: This `tupleToTraversable` was used implicitly a couple times in this class (See the overloaded `and` method in `KvArgsOps` class). I can remove the `implicit` however, if that's a concern.

core/src/main/scala/com/mesosphere/usi/core/Scheduler.scala
takirala: `takingImplicit` is from `scalalogging` library.
One can do a 
```
 private val log = Logger.takingImplicit[KvArgs](getClass).withContext("name", "value")
```
however, but `takingImplicit` is mandatory, may be we can create a trait (just like `StrictLogging`) if you think that would be more readable ?!

core/src/main/scala/com/mesosphere/usi/core/Scheduler.scala
jeschkies: That makes sense to me but Why do I need the first line then? I'd expect

```scala
composedLogger.info(s"Connecting to $mesosHostName")(("mesosHostName", mesosHostName))
```

The extra `withContext` does not make much sense to me.

core/src/main/scala/com/mesosphere/package.scala
```diff
@@ -0,0 +1,161 @@
...
+   *    log.info("hello world!")(("key1", "value1").and("key2", "value2").and("key3", "value3"))
+   *
+   *    // Now compose the logger with static meta data.
+   *    // The withContext method here is defined on a value class of LoggerTakingImplicit.
+   *    val composedLogger : LoggerAndContext = log.withContext("constantKey", "constantValue")
+   *
+   *    // Note that the type of composedLogger is just a simple tuple containing the logger as well as context.
+   *    // The state of logger object itself is not mutated in any manner.
+   *    // The logger is not knowledgeable about the context in any manner. The tuple is the ONLY thing that
+   *    // is used correlate between a logger and the context associated with it.
```
jeschkies: What is the motivation of this. It feels we almost get the downside of both worlds. No real wrapper and no real context attached.

Why not 

```scala
class LoggerAndContext(logger: LoggerTakingImplicit[KvArgs], context: KvArgs) {

  def info() = ??? // how is the context passed?
}
```

Which would make the context passing explicit. Right now I still don't understand how

```scala
log.info("hello world!")(("key1", "value1"))
```

works.

jeschkies (changes requested):
Let's talk about this in our sync. I think we have a few too many implicit conversions.

core/src/main/scala/com/mesosphere/usi/core/Scheduler.scala
```diff
@@ -43,9 +44,14 @@ import scala.concurrent.Future
  *                                     +----------------------+
  */
 object Scheduler {
+
+  private val log = Logger.takingImplicit[KvArgs](getClass)
+
   case class MesosConnection(mesosHostName: String)
 
   def connect(mesosHostName: String): Future[(MesosConnection, Flow[SpecEvent, StateEvent, NotUsed])] = {
+    val composedLogger = log.withContext("mesosHostName", mesosHostName)
```
zen-dog: Looking only at this usage example I don't quite understand: if the `composedLogger` already has context with `mesosHostName` - why do we need to implicitly pass it in the next line?

takirala: (Discussed this offline with @jeschkies ) previously, a tuple was being created whenever `withContext` is being called, so the `log` was not actually being associated with a `context` but the abstractions helped to write `log.context`.

Decided that the tuple was not giving any added benefit, we changed from "composing a logger with context" to "composing a context" to be used with logger. Updated the example to reflect the same.

core/src/main/scala/com/mesosphere/package.scala
```diff
@@ -0,0 +1,101 @@
...
+        val (key, value) = elem
+        MDC.put(key, value.toString)
+      }
+      originalMsg
+    }
+
+    override def afterLog(a: KvArgs): Unit = a.foreach { case (key, _) => MDC.remove(key) }
+  }
+
+  implicit val emptyKvArgs: KvArgs = Traversable()
```
jeschkies: Where is this required?

core/src/main/scala/com/mesosphere/package.scala
```diff
@@ -0,0 +1,101 @@
...
+
+  def kv(kv: (String, Any)): KvArgs = Traversable(kv)
+
+  def kv(key: String, value: Any): KvArgs = kv((key, value))
+
+  implicit class KvArgsOps(val self: KvArgs) extends AnyVal {
```
jeschkies: Why not simply define a class?

```scala
case class KvArgs(args: List[(String, Any)]) extends Traversable[(String, Any)] {
  override def foreach() = ???

  def and(kv: (String, Any)) = copy(kv :: args)
  def and(key: String, value: Any) = and((key, value))
}

object KvArg {

  def apply(kv: (String, Any)) = KvArgs(List(kv))
  def apply(key: String, value: Any) = KvArgs(List((key, value)))
}
```

No implicit conversions. This seems much more straight forward to me. What do you think?

core/src/main/scala/com/mesosphere/package.scala
```diff
@@ -0,0 +1,101 @@
...
+
+  /**
+   * Defines `logger` as a value initialized with an underlying `org.slf4j.Logger`
+   * named according to the class into which this trait is mixed.
+   *
+   * Also refer [[com.typesafe.scalalogging.LazyLogging]]
+   */
+  trait ImplicitLazyLogging {
+    @transient
+    protected val logger: LoggerTakingImplicit[KvArgs] =
```
jeschkies: How does it differ from `ImplicitStrictLogging`? When should it be used?

jeschkies (changes requested):
Please address my concern with the implicit conversion. Looks good otherwise 🙂

core/src/main/scala/com/mesosphere/package.scala
takirala: Agreed and updated.

core/src/main/scala/com/mesosphere/package.scala
takirala: This is required when we have a logger that takes implicit and is being used in multiple places in the class. But if we want to use the same logger for logging without any context, this implicit can be used. However, since doing that _may_ be rare, i have moved this to one of the `apply` methods of the case class. We can bring back the implicit if needed.

core/src/main/scala/com/mesosphere/package.scala
takirala: Updated code to correctly mark this as `lazy`. I've done what was being done in `StrictLogging` and `LazyLogging` in [scalalogging's Logging.scala](https://github.com/lightbend/scala-logging/blob/master/src/main/scala/com/typesafe/scalalogging/Logging.scala)

I _think_ we may end up using the lazy version in those cases where we know we may/may not need a logger associated with a short lived object (that has no companion object may be?!). I can remove this if you think it's better to add it if/when need arises.

core/src/main/scala/com/mesosphere/package.scala
```diff
@@ -0,0 +1,104 @@
...
+   *    val moreContext = KvArgs("key1", "value1").and("key2", "value2").and("key3", "value3")
+   *    log.info("hello world!")(moreContext.and("key3", "key4"))
+   *
+   *    // The above statement will log through whatever backend slf4j was tied to statically.
+   *    // All the key value parameters in context are populated in MDC just before the log statement is made and are
+   *    // deleted immediately after. This happens every time the `logger` object is used to generate a log statement.
+   * }}}
+   */
+
+  case class KvArgs(args: Traversable[(String, Any)]) {
```
jeschkies: Why not a `List`? I believe that a list prepend would be faster than `args ++ kv,args`.

core/src/main/scala/com/mesosphere/package.scala
```diff
@@ -0,0 +1,104 @@
...
+    /**
+     * Helper methods to enable fluent composition of context parameter(s).
+     */
+    def and(kv: KvArgs): KvArgs = copy(args ++ kv.args)
+
+    def and(k: String, v: String): KvArgs = and(KvArgs(k, v))
```
jeschkies: Hm, we could improve performance with
```suggestion
    def and(k: String, v: String): KvArgs = copy((k, v) + args)
```

jeschkies (approved):
Thanks! I'm not sure we should use a `Traversable`, though.

core/src/main/scala/com/mesosphere/package.scala
```diff
@@ -0,0 +1,103 @@
...
+    }
+
+    override def afterLog(a: KvArgs): Unit = a.args.foreach { case (key, _) => MDC.remove(key) }
+  }
+}
+
+package mesosphere {
+
+  /**
+   * Abstractions in this class were designed with below interests:
```
alenkacz: nice docs! 👍

core/src/main/scala/com/mesosphere/package.scala
```diff
@@ -0,0 +1,103 @@
...
+   *
+   * {{{
+   *    // import all the implicit defined here in to scope.
+   *    import com.mesosphere._
+   *
+   *    // Mixin one of the traits.
+   *    class MyClass with ImplicitStrictLogging {....
+   *
+   *    // Make a log statement with no (empty) context.
+   *    logger.info("hello world!")(KvArgs())
```
alenkacz: do you need to do it like this or will logger.info("aaa") work as well?

core/src/main/scala/com/mesosphere/package.scala
```diff
@@ -0,0 +1,103 @@
...
+   *    val moreContext = KvArgs("key1", "value1").and("key2", "value2").and("key3", "value3")
+   *    log.info("hello world!")(moreContext.and("key3", "key4"))
+   *
+   *    // The above statement will log through whatever backend slf4j was tied to statically.
+   *    // All the key value parameters in context are populated in MDC just before the log statement is made and are
+   *    // deleted immediately after. This happens every time the `logger` object is used to generate a log statement.
+   * }}}
+   */
+
+  case class KvArgs(args: List[(String, Any)]) {
```
alenkacz: Since we can now compose these and even pass them on between methods, could we do a better name? `LoggerContext` would be for me a better choice because then when I see it as a parameter it's obvious what is it for. WDYT?

alenkacz (commented):
I would like to see both used in code in both hello-world as well as scheduler. I think there's enough code now to be able to do some meaningful logging somewhere

core/src/main/scala/com/mesosphere/package.scala
takirala: as of now, we need to pass an empty `KvArgs()`. If this tends out to be a common pattern, we can bring back the global implicit for empty `KvArgs` https://github.com/mesosphere/usi/pull/36#discussion_r257237992

core/src/main/scala/com/mesosphere/package.scala
takirala: - `LoggerContext` in particular is too bold for what we are trying to accomplish here. There are already `LoggerContext`s in logback, log4j etc.. 
- Pieces of [StructuredArguments](https://github.com/logstash/logstash-logback-encoder/blob/master/src/main/java/net/logstash/logback/argument/StructuredArguments.java) class is what we are trying to emulate with `KvArgs`. the type of args being a `List[(String, Any)]` may become narrow in the future and we _may_ have to introduce some other broader range structured args.
with these two things in mind, i initially had only `kv` in mind instead of `KvArgs` but `LoggerContext` would be too broad for `KvArgs` i think

takirala:
@alenkacz i have added a simple log statement in helloworld integration test. i feel adding more meaningful logs to other modules will increase the scope of this PR.

core/src/main/scala/com/mesosphere/package.scala
takirala: brought the global implicit back :)

core/src/main/scala/com/mesosphere/package.scala
timcharper: @jeschkies I think the data structure we need is a list of tuples

alenkacz (approved):
I would still go with a different name for KvArgs, because this is confusing `processOffer(offer: Offer, pods: Pods, kvArgs: KvArgs)` but I guess we can always change that.

Thanks for the PR, Tarun and addressing all our comments throughout its lifetime :)

PR #37, opened 2019-02-08
Add an `example:simple-hello-world` framework

zen-dog:
Summary:
Add a new example framework that:

- uses only the raw mesos-client and none of the core scheduling logic
- successfully subscribes to Mesos master
- starts one `echo "Hello, world" && sleep N` task
- exits should the task fails (or fails to start)

Good to test against local Mesos and to demonstrate how a framework core-loop might look like

JIRA: DCOS-47479


examples/raw-hello-world/src/main/resources/application.conf
```diff
@@ -0,0 +1,30 @@
+# This is still the default configuration but it's good to see what can be changed
+mesos-client {
+
+  master-url: "127.0.0.1:5050"
```
jeschkies: What do you think about using an environment variable here?

examples/raw-hello-world/src/main/scala/com/mesosphere/mesos/examples/EventProcessor.scala
```diff
@@ -0,0 +1,98 @@
...
+    )
+
+    import TaskState._
+    taskState match {
+      // We handle only "happy" task states by acknowledging them and exit on everything else
+      case TASK_RUNNING | TASK_STAGING | TASK_STARTING =>
+        // Only task status updates that have UUIDs must be acknowledged.
+        // This is always the case for status updates
+        val newTask = task.copy(state = Some(taskState))
+        val msgUuid = uuid.getOrElse(throw new IllegalArgumentException(s"Missing uuid in task status update: $event"))
```
jeschkies: Should this be handled at the top of the method?

jeschkies (approved):
Nice docs! Thanks.

examples/raw-hello-world/src/main/resources/application.conf
zen-dog: Why? It's an example, nobody will use it in production, right? 🤞

examples/raw-hello-world/src/main/scala/com/mesosphere/mesos/examples/EventProcessor.scala
zen-dog: No, some status updates don't have `uuid`s but **task** status updates should. This is more of a sanity check here.

alenkacz (approved):
To me this is more `mesos-client` related than `hello-world` related (which should be USI). To avoid confusion I would prefer if the name of the example folder actually contain `mesos-client` in a form, not `hello-world`. Otherwise 👍

zen-dog:
> To me this is more `mesos-client` related than `hello-world` related (which should be USI)

I would rather rename the current `hello-world` to smth. like `core-hello-world` to give users a feeling fo everything `core` take care of (types, offer matching, event processing, state holding etc.). I think this will be an interesting comparison and will immediately demonstrate the added value of using `core` to do the same thing e.g. `echo Hello, world`

alenkacz:
@zen-dog I still believe that mesos-client examples in the end should not be in `examples` especially since it's not supported way of using the library but that's a different story. There's just so many components that even I am losing track of what is what... :) But I approved, so feel free to merge as is... 'raw-hello-world' is strange though... I would be fine with just one mesos-client example, being it the more complex one. Especially since we don't support mesos-client as is

timcharper (approved):

alenkacz:
@zen-dog new name options: `basic-framework` `simple-framework`... ? :) That means more to me than `raw-hello-world` :)

PR #38, opened 2019-02-11
Add integration tests for  frameworks

zen-dog:
ensuring that the frameworks start and successfully connect to Mesos.

### DIFF
--- a/examples/mesos-client-example/src/test/scala/com/mesosphere/mesos/examples/MesosClientExampleFrameworkTest.scala
+++ b/examples/mesos-client-example/src/test/scala/com/mesosphere/mesos/examples/MesosClientExampleFrameworkTest.scala
@@ -32,8 +32,6 @@ class MesosClientExampleFrameworkTest extends AkkaUnitTest with MesosClusterTest
 
   class Fixture(existingFrameworkId: Option[FrameworkID.Builder] = None) {
     val mesosUrl = new java.net.URI(mesosFacade.url)
-    val mesosHost = mesosUrl.getHost
-    val mesosPort = mesosUrl.getPort
 
     val conf = ConfigFactory.parseString(s"""
                                               |mesos-client.master-url="${mesosUrl.getHost}:${mesosUrl.getPort}"
--- a/examples/mesos-client-example/src/test/scala/com/mesosphere/mesos/examples/MesosClientExampleFrameworkTest.scala
+++ b/examples/mesos-client-example/src/test/scala/com/mesosphere/mesos/examples/MesosClientExampleFrameworkTest.scala
@@ -1,4 +1,5 @@
 package com.mesosphere.mesos.examples
+
 import com.mesosphere.utils.AkkaUnitTest
 import com.mesosphere.utils.mesos.MesosClusterTest
 import com.mesosphere.utils.mesos.MesosFacade.ITFramework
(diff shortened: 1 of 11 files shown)


examples/mesos-client-example/src/test/scala/com/mesosphere/mesos/examples/MesosClientExampleFrameworkTest.scala
```diff
@@ -0,0 +1,42 @@
...
+    try fn(f)
+    finally {
+      f.framework.client.killSwitch.shutdown()
+    }
+  }
+
+  class Fixture(existingFrameworkId: Option[FrameworkID.Builder] = None) {
+    val mesosUrl = new java.net.URI(mesosFacade.url)
+
+    val conf = ConfigFactory.parseString(s"""
```
alenkacz: my preference would be to use something that's not parsing raw string. What about `parseMap`?

examples/mesos-client-example/src/test/scala/com/mesosphere/mesos/examples/MesosClientExampleFrameworkTest.scala
```diff
@@ -0,0 +1,42 @@
...
+                                              |mesos-client.master-url="${mesosUrl.getHost}:${mesosUrl.getPort}"
+                                            """.stripMargin).withFallback(ConfigFactory.load())
+
+    val framework = MesosClientExampleFramework(conf.getConfig("mesos-client"))
```
alenkacz, line 42: why are you getting a subtree of a config here? Why not create the subtree in the first place?

alenkacz (approved):
LGTM. Thanks!

examples/mesos-client-example/src/test/scala/com/mesosphere/mesos/examples/MesosClientExampleFrameworkTest.scala
zen-dog: I agree, and once we have targeted Java11 we could do inline with `Map.of` ;)

examples/mesos-client-example/src/test/scala/com/mesosphere/mesos/examples/MesosClientExampleFrameworkTest.scala
zen-dog: doesn't work otherwise - `withFallback(ConfigFactory.load())` brings the rest of the configuration and it has the `mesos-client` prefix.

examples/mesos-client-example/src/test/scala/com/mesosphere/mesos/examples/MesosClientExampleFrameworkTest.scala
jeschkies: Java 11 just merged 🙂

examples/simple-hello-world/src/test/scala/com/mesosphere/mesos/examples/SimpleHelloWorldFrameworkTest.scala
```diff
@@ -0,0 +1,50 @@
...
+    And("eventually hello-world task should be up and running")
+    eventually {
+      val framework = mesosFacade.frameworks().value.frameworks.head
+      val task = framework.tasks.head
+      task.name shouldBe "hello-world"
+      task.state.get shouldBe "TASK_RUNNING"
+    }
+  }
+
+  def withFixture(frameworkId: Option[FrameworkID.Builder] = None)(fn: Fixture => Unit): Unit = {
```
jeschkies, line 35: I'm not a big fan of this method but also failed to come up with an alternative 😉

jeschkies (approved):

PR #40, opened 2019-02-13
Register with Mesos and launch a real task

timcharper:
JIRA Issues: DCOS-47474


core/src/main/scala/com/mesosphere/usi/core/ProtoBuilders.scala
```diff
@@ -0,0 +1,201 @@
+package com.mesosphere.usi.core
+
+import com.google.protobuf.ByteString
+import org.apache.mesos.v1.{Protos => Mesos}
+
+private[usi] object ProtoBuilders {
```
jeschkies: This seems similar to [ProtosHelper](https://github.com/mesosphere/usi/blob/master/examples/simple-hello-world/src/main/scala/com/mesosphere/mesos/examples/ProtosHelper.scala). Should we consolidate them?

core/src/test/scala/com/mesosphere/usi/core/integration/SchedulerIntegrationTest.scala
```diff
@@ -0,0 +1,46 @@
...
+    val frameworkInfo = Protos.FrameworkInfo.newBuilder().setUser("test").setName("SimpleHelloWorldExample").build()
+
+    val client = Scheduler.connect(settings, frameworkInfo).futureValue
+
+    val podId = PodId("pod-1")
+    val (input, output) = Source
+      .queue[SpecEvent](16, OverflowStrategy.fail)
+      .via(client.schedulerFlow)
+      .toMat(Sink.queue())(Keep.both)
+      .run
```
jeschkies, line 30: This is interesting. @zen-dog and I were wondering how one would use the flow. What do you think of returning `(Source[SpecEvent], Sink[StateEvent])` instead of `Flow[SpecEvent, StateEvent]` as we do now?

core/src/main/scala/com/mesosphere/usi/core/ProtoBuilders.scala
zen-dog: I don't think so. The whole idea of these examples is to show the reader what he/she gets by taking core (e.g. not having to build Protobufs). It's fine to have this duplicated imho

core/src/test/scala/com/mesosphere/usi/core/SchedulerTest.scala
```diff
@@ -3,27 +3,28 @@ package com.mesosphere.usi.core
...
-  PodStatusUpdated,
-  RunSpec,
-  SpecEvent,
-  StateEvent,
-  TaskId
-}
+import com.mesosphere.mesos.client.MesosCalls
+import com.mesosphere.usi.core.launching.SimpleShellCommandInfoGenerator
+import com.mesosphere.usi.core.matching.ScalarResourceRequirement
+import com.mesosphere.usi.core.models._
```
zen-dog, line 10: Oh no, you did not! :P

core/src/main/scala/com/mesosphere/usi/core/SchedulerLogic.scala
```diff
@@ -173,51 +255,67 @@ private[core] object SchedulerLogic {
...
 
-        case Some(podSpec) =>
-          if (podSpec.goal == Goal.Terminal) {
-            taskIdsFor(podSpec).foldLeft(initialIntents) { (effects, taskId) =>
-              effects.withMesosCall(Mesos.Call.Kill(taskId))
+          case Some(podSpec) =>
+            if (podSpec.goal == Goal.Terminal) {
+              taskIdsFor(podSpec).foldLeft(initialIntents) { (effects, taskId) =>
+                effects.withMesosCall(
+                  mesosCallFactory.newKill(taskId.asProto, state.podRecords.get(podSpec.id).map(_.agentId.asProto), None))
```
alenkacz: is it cool as we're initiating kill for any change? E.g. the change is emitted even on a snapshot. Or we're going to tackle this later?

core-models/src/main/scala/com/mesosphere/usi/core/models/PodStatus.scala
```diff
@@ -1,5 +1,7 @@
 package com.mesosphere.usi.core.models
 
+import org.apache.mesos.v1.{Protos => Mesos}
```
jeschkies, line 3: Where is that used?

core-models/src/main/scala/com/mesosphere/usi/core/models/ResourceRequirement.scala
```diff
@@ -0,0 +1,9 @@
+package com.mesosphere.usi.core.models
+
+import org.apache.mesos.v1.{Protos => Mesos}
+
+trait ResourceRequirement {
```
jeschkies, line 8: Could you add docs since this is part of our protocol?

core/src/main/scala/com/mesosphere/usi/core/SchedulerLogic.scala
```diff
@@ -173,51 +255,67 @@ private[core] object SchedulerLogic {
...
+              }(collection.breakOut))
 
-      case Mesos.Event.StatusUpdate(taskId, taskStatus) =>
+              (intents ++ offerMatchIntents, pending -- matchedPodIds)
+          }
+        intents.result
+
+      case UpdateEvent(updateEvent) =>
+        val taskStatus: Mesos.TaskStatus = updateEvent.getStatus
+        val taskId = TaskId(taskStatus.getTaskId.getValue)
```
alenkacz: can we make UpdateEvent a case class and have these (taskId, podId) there as a vals/fields?

core/src/test/scala/com/mesosphere/usi/core/SchedulerTest.scala
timcharper: Organize imports did it I swear

core/src/main/scala/com/mesosphere/usi/core/SchedulerLogic.scala
timcharper: Yeah, i think it is fine for now. We'll want a more robust kill mechanism with retry I think in the future perhaps? We should add a JIRA

core/src/test/scala/com/mesosphere/usi/core/integration/SchedulerIntegrationTest.scala
timcharper: It's something we can explore; there is the thing where the Source and the Sink should share the same materialization as they're backed by the same scheduler logic component; but we could split apart a flow by using Sink.asPublisher etc.

core/src/main/scala/com/mesosphere/usi/core/SchedulerLogic.scala
timcharper: https://jira.mesosphere.com/browse/DCOS-48513

core-models/src/main/scala/com/mesosphere/usi/core/models/ResourceRequirement.scala
timcharper: ✅

core/src/main/scala/com/mesosphere/usi/core/SchedulerLogic.scala
timcharper: Perhaps? I am trying to limit the amount of special containers we use for Protobufs, and instead use helper methods to read them.

core/src/test/scala/com/mesosphere/usi/core/integration/SchedulerIntegrationTest.scala
jeschkies: Ah, that's a good argument. I'd love to have the separation though since I can think of cases when I'd like to have pulling and pushing blocking in a thread on my side when the orchestration code cannot be async.

core/src/main/scala/com/mesosphere/usi/core/SchedulerLogic.scala
timcharper: I misread what you wrote :) Yes this is a good idea.

core/src/test/scala/com/mesosphere/usi/core/integration/SchedulerIntegrationTest.scala
timcharper: Yes; and, you could also have a graph on the other side

core/src/main/scala/com/mesosphere/usi/core/ResourceUtil.scala
```diff
@@ -0,0 +1,218 @@
...
+
+import com.github.ghik.silencer.silent
+import com.typesafe.scalalogging.StrictLogging
+import org.apache.mesos.v1.Protos.Resource.DiskInfo.Source
+import org.apache.mesos.v1.Protos.Resource.{DiskInfo, ReservationInfo}
+import org.apache.mesos.v1.{Protos => Mesos}
+import scala.collection.JavaConverters._
+import scala.util.control.NonFatal
+
+object ResourceUtil extends StrictLogging {
```
alenkacz, line 10: note for reviewer: this code seems to be basically copy&paste from Marathon. I'll follow up with migrating the tests as well

core/src/main/scala/com/mesosphere/usi/core/Scheduler.scala
```diff
@@ -43,44 +50,84 @@ import scala.concurrent.Future
...
 
-  def connect(mesosHostName: String): Future[(MesosConnection, Flow[SpecEvent, StateEvent, NotUsed])] = {
-    val flow = Flow.fromGraph {
-      GraphDSL.create(Scheduler.unconnectedGraph, FakeMesos.flow)((_, _) => NotUsed) { implicit builder =>
-        { (graph, mockMesos) =>
+  type StateOutput = (StateSnapshot, Source[StateEvent, Any])
+
+  case class MesosConnection(mesosHostName: String, mesosCallFactory: MesosCalls, frameworkId: FrameworkID, schedulerFlow: Flow[SpecInput, StateOutput, NotUsed])
+
+  // TODO - provide abstraction for FrameworkID persistence, recovery, etc.; we'll want to exclude the frameworkId
```
alenkacz: should we capture this in JIRA?

alenkacz: also for the comment below

core/src/main/scala/com/mesosphere/usi/core/Scheduler.scala
```diff
@@ -43,44 +50,84 @@ import scala.concurrent.Future
  *                                     +----------------------+
  */
 object Scheduler {
```
alenkacz, line 47: what do you think about PodScheduler?

core/src/main/scala/com/mesosphere/usi/core/Scheduler.scala
```diff
@@ -43,44 +50,84 @@ import scala.concurrent.Future
...
+      case c: StateUpdated => c
+      case _ => ???
+    }
+    (StateSnapshot.empty, stateUpdates)
+  }
+
+  def unconnectedGraph(mesosCallFactory: MesosCalls): BidiFlow[SpecInput, StateOutput, MesosEvent, MesosCall, NotUsed] = {
     BidiFlow.fromGraph {
-      GraphDSL.create(new SchedulerLogicGraph) { implicit builder => (schedulerLogic) =>
+      GraphDSL.create(new SchedulerLogicGraph(mesosCallFactory)) { implicit builder => (schedulerLogic) =>
```
alenkacz, line 99: I would go for `SchedulerGraph`

core/src/main/scala/com/mesosphere/usi/core/logic/MesosEventsLogic.scala
```diff
@@ -0,0 +1,144 @@
...
+            }(collection.breakOut)
+
+            matchPodSpecsTaskRecords(offer, newRemainingResources, result.updated(podSpec.id, taskInfos), rest)
+          case None =>
+            matchPodSpecsTaskRecords(offer, remainingResources, result, rest)
+        }
+    }
+  }
+
+  def matchOffer(offer: Mesos.Offer, specs: Iterable[PodSpec]): (Set[PodId], SchedulerEventsBuilder) = {
```
alenkacz: should be private

core/src/main/scala/com/mesosphere/usi/core/logic/SpecificationLogic.scala
```diff
@@ -0,0 +1,93 @@
...
+import com.mesosphere.usi.core.models._
+import com.typesafe.scalalogging.StrictLogging
+import org.apache.mesos.v1.{Protos => Mesos}
+
+/**
+  * The current home for USI business logic
+  *
+  * TODO separate the Mesos and the specification logic.
+  */
+private[core] class SpecificationLogic(mesosCallFactory: MesosCalls) extends StrictLogging {
```
alenkacz: Specification -> Spec

alenkacz (commented):
After about half-done I have to take a step back and try some stuff in IDEA to see if we can do better in some places. Then I'll come back to continue the review :)

core-models/src/main/scala/com/mesosphere/usi/core/models/RunSpec.scala
```diff
@@ -2,6 +2,5 @@ package com.mesosphere.usi.core.models
 
 /**
   * Specification used to launch a [[PodSpec]]
-  * @param cpus number of cpus requested by the pod
   */
-case class RunSpec(cpus: Double)
+case class RunSpec(resourceRequirements: Seq[ResourceRequirement], commandBuilder: CommandInfoGenerator)
```
zen-dog: I don't know how I feel about `commandBuilder` in the `RunSpec`. I guess it's here because we wanted to keep it flexible for the orchestrators? But on the other hand, orchestrators should be primarily about business logic and not about building Mesos protobufs correctly. Maybe we could _make it extendable later_ similar to how plugins in Marathon _can but not must_ override `TaskInfo` and `TaskGroupInfo`?

core/src/main/scala/com/mesosphere/usi/core/ResourceUtil.scala
```diff
@@ -0,0 +1,218 @@
...
+    * Decrements the scalar resource by amount
+    *
+    */
+  def consumeScalarResource(resource: Mesos.Resource, amount: Double): Option[Mesos.Resource] = {
+    require(resource.getType == Mesos.Value.Type.SCALAR)
+    val isMountDiskResource =
+      resource.hasDisk && resource.getDisk.hasSource &&
+        (resource.getDisk.getSource.getType == Source.Type.MOUNT)
+
+    // TODO(jdef) would be nice to use fixed precision like Mesos does for scalar math
```
zen-dog: `// TODO(jdef)`? 🤣 Is James aware?

core/src/main/scala/com/mesosphere/usi/core/Scheduler.scala
```diff
@@ -43,44 +50,84 @@ import scala.concurrent.Future
...
-  def connect(mesosHostName: String): Future[(MesosConnection, Flow[SpecEvent, StateEvent, NotUsed])] = {
-    val flow = Flow.fromGraph {
-      GraphDSL.create(Scheduler.unconnectedGraph, FakeMesos.flow)((_, _) => NotUsed) { implicit builder =>
-        { (graph, mockMesos) =>
+  type StateOutput = (StateSnapshot, Source[StateEvent, Any])
+
+  case class MesosConnection(mesosHostName: String, mesosCallFactory: MesosCalls, frameworkId: FrameworkID, schedulerFlow: Flow[SpecInput, StateOutput, NotUsed])
+
+  // TODO - provide abstraction for FrameworkID persistence, recovery, etc.; we'll want to exclude the frameworkId
+  // TODO - This layer should provide a cancellation mechanism and an auto-reconnect mechanism
```
zen-dog: Cancellation - yes, auto-reconnect - I don't think so (especially when thinking about HA frameworks).

core/src/main/scala/com/mesosphere/usi/core/Scheduler.scala
zen-dog: how about injecting the complete `MesosClient` here? This way `core` can be easily tested with a Mesos mock

core/src/main/scala/com/mesosphere/usi/core/Scheduler.scala
```diff
@@ -43,44 +50,84 @@ import scala.concurrent.Future
...
 
-  /*
-   */
-  def unconnectedGraph: BidiFlow[SpecEvent, StateEvent, Mesos.Event, Mesos.Call, NotUsed] = {
+  /**
+    * We expess an interface of receiving a snapshot, followed by a series of events, to guide consumers of USI down a
+    * proper implementation path. A SpecSnapshot should be the very first thing that the USI Scheduler receives from the
+    * Framework implementation.
+    *
+    * However, for convenience in working with streams, internally we deal with a single stream of SpecEvents.
```
zen-dog, line 78: I imagined that scheduler would start Mesos task reconciliation after receiving the `SpecSnapshot` since this is a good point in time when all the reconciled tasks should be ideally found in the snapshot. Wdyt?

core/src/main/scala/com/mesosphere/usi/core/Scheduler.scala
```diff
@@ -43,44 +50,84 @@ import scala.concurrent.Future
...
+    rest.prepend(Source.single(snapshot))
+  }
+
+  private val stateOutputBreakoutFlow: Flow[StateEvent, StateOutput, NotUsed] = Flow[StateEvent].prefixAndTail(0).map { case (_, stateEvents) =>
+    // TODO (DCOS-47476) use actual prefixAndTail and expect first event to be a Snapshot
+    val stateUpdates = stateEvents.map {
+      case c: StateUpdated => c
+      case _ => ???
+    }
+    (StateSnapshot.empty, stateUpdates)
```
zen-dog: The other way around (see my comment above), I'm not sure what difference does it make for the orchestration layer to receive `StateSnapshot` as opposed to a stream of `StateUpdated`.

core/src/main/scala/com/mesosphere/usi/core/ResourceUtil.scala
alenkacz: yeah it made me laugh as well :D

alenkacz (approved):
I've submitted few patches in https://github.com/mesosphere/usi/pull/41 with that getting in I am fine with this. There's still a lot to polish but I would rather iterate on that while we evolve core + hello-world side by side.

The only blocker for me is ResourceUtil tests. I think those should be ported here.

core/src/main/scala/com/mesosphere/usi/core/Scheduler.scala
alenkacz: I believe it's just enforcing the order of Snapshot always being first. You need the snapshot to finish initialization of your orchestration layer

core/src/main/scala/com/mesosphere/usi/core/Scheduler.scala
alenkacz: I believe it should be the case as well. But I think reconciliation is out of scope for now

core-models/src/main/scala/com/mesosphere/usi/core/models/RunSpec.scala
alenkacz: We already discussed on slack but just for reference. The usage of this is  `RunSpec(SimpleShellCommandInfoGenerator("echo 'abc'"))`. I would personally prefer something like `RunSpec(ShellCommand("cmd"))` but we can iterate on that

core/src/main/scala/com/mesosphere/usi/core/SchedulerLogicHandler.scala
```diff
@@ -0,0 +1,146 @@
...
+  * - We have built-in guarantees that evolutions to the SchedulerState can be reliably replicated by processing these
+  * events, since they led to the changes in the first place
+  * - We can easily know which portions of the SchedulerState should be persisted during the persistence layer.
+  * - It restricts, via the type system, the portions of the state the business logic is allowed (IE: it
+  * would be illegal for the business logic to update a specification)
+  *
+  * ## Intents and Events
+  *
+  * In the SchedulerLogic code, we'll use the word intents and events. Intents are things not yet applied, and should
+  * be. Events are things that were applied and we're notifying you about.
```
zen-dog, line 34: ...and should be...?

core/src/main/scala/com/mesosphere/usi/core/SchedulerLogicHandler.scala
```diff
@@ -0,0 +1,146 @@
...
+  * ## Concept of a "Frame"
+  *
+  * Visualized:
+  *
+  * [Incoming Event: offer match]
+  *      |
+  *      v
+  * (beginning of frame)
+  *   apply pod launch logic:
+  *     emit Mesos accept offer call for matched pending launch pods
```
zen-dog:
```suggestion
  *     create an intent for Mesos accept offer call for matched pending launch pods
```

core/src/main/scala/com/mesosphere/usi/core/SchedulerLogicHandler.scala
```diff
@@ -0,0 +1,146 @@
...
+  *   process revive / suppress need
+  * (end of frame)
+  *      |
+  *      v
+  * [Emit all state events and Mesos call intents]
+  *
+  * In the SchedulerLogic, each event is processed (either a Specification updated event, or a Mesos event) in what we
+  * call a "frame". During the processing of a single event (such as an offer), the SchedulerLogic will be updated
+  * through the application of several functions. The result of that event will lead to the emission of several
+  * stateEvents and Mesos intents, which will be accumulated and emitted via a data structure we call the FrameResult.
```
zen-dog, line 63: Nice docs 👏

core/src/main/scala/com/mesosphere/usi/core/SpecState.scala
```diff
@@ -0,0 +1,8 @@
+package com.mesosphere.usi.core
+import com.mesosphere.usi.core.models.{PodId, PodSpec}
+
+case class SpecState(podSpecs: Map[PodId, PodSpec])
```
zen-dog, line 4: nit: do we need a case class wrapper? can it be a
```suggestion
type SpecState = Map[PodId, PodSpec]
```

core/src/main/scala/com/mesosphere/usi/core/Scheduler.scala
timcharper: No? We should crash on Mesos connection loss?

core/src/main/scala/com/mesosphere/usi/core/ResourceUtil.scala
timcharper: lol copied from Marathon

core/src/main/scala/com/mesosphere/usi/core/Scheduler.scala
timcharper: Yes, https://jira.mesosphere.com/browse/DCOS-47476

core/src/main/scala/com/mesosphere/usi/core/SpecState.scala
timcharper: We'll have other specs in here, reservation specs

core/src/main/scala/com/mesosphere/usi/core/logic/MesosEventsLogic.scala
timcharper: ✅

core/src/main/scala/com/mesosphere/usi/core/logic/SpecificationLogic.scala
timcharper: ✅

core-models/src/main/scala/com/mesosphere/usi/core/models/RunSpec.scala
timcharper: Yeah, I'm not exactly sure at the moment. 

https://jira.mesosphere.com/browse/DCOS-48503
https://jira.mesosphere.com/browse/DCOS-47481

core/src/main/scala/com/mesosphere/usi/core/Scheduler.scala
zen-dog: Signal that to the orchestration layer and let it decide, I'd say

core/src/main/scala/com/mesosphere/usi/core/Scheduler.scala
zen-dog: `//TODO:  DCOS-48503`

core/src/main/scala/com/mesosphere/usi/core/Scheduler.scala
zen-dog: Discussed

core/src/main/scala/com/mesosphere/usi/core/Scheduler.scala
zen-dog: Discussed

core/src/main/scala/com/mesosphere/usi/core/Scheduler.scala
zen-dog: Discussed

core/src/test/scala/com/mesosphere/usi/core/helpers/FakeMesos.scala
```diff
@@ -0,0 +1,61 @@
...
+
+import akka.NotUsed
+import akka.stream.scaladsl.Flow
+import com.mesosphere.usi.core.models.AgentId
+import com.mesosphere.usi.core.protos.ProtoBuilders
+import org.apache.mesos.v1.{Protos => Mesos}
+import org.apache.mesos.v1.scheduler.Protos.{Call => MesosCall, Event => MesosEvent}
+import scala.collection.JavaConverters._
+
+object FakeMesos {
```
zen-dog: It looks like this class is to stay. Should we call it `MesosMock`?

core/src/test/scala/com/mesosphere/usi/core/helpers/FakeMesos.scala
```diff
@@ -0,0 +1,61 @@
...
+  import com.mesosphere.usi.core.protos.ProtoBuilders._
+  import com.mesosphere.usi.core.protos.ProtoConversions._
+  val fakeAgentId = AgentId("fake-agent")
```
zen-dog: and `mockAgentId`? and `mock-agent`? and `mockMesosLoggingFlow`? and... you get the idea ;)

core/src/test/scala/com/mesosphere/usi/core/integration/SchedulerIntegrationTest.scala
```diff
@@ -0,0 +1,51 @@
...
+class SchedulerIntegrationTest extends AkkaUnitTest with MesosClusterTest with Inside {
+  "launches the thing" in {
+    implicit val materializer = ActorMaterializer()
+
+    val settings = MesosClientSettings(mesosFacade.url)
+    val frameworkInfo = Protos.FrameworkInfo.newBuilder().setUser("test").setName("SimpleHelloWorldExample").build()
+
+    val client = Scheduler.connect(settings, frameworkInfo).futureValue
+
+    val podId = PodId("pod-1")
```
zen-dog: Can we please not use the `pod1` and `app1` names? Please keep them unique and readable (take Marathon ITs as an example)

core/src/main/scala/com/mesosphere/usi/core/Scheduler.scala
timcharper: We need a name for the component that specifically refers to what is currently called SchedulerLogic, then. Scheduler includes persistence component as we discussed yesterday.

core/src/main/scala/com/mesosphere/usi/core/Scheduler.scala
timcharper: ✅  I just rolled this stuff back and made it so the MesosClient is passed in

core-models/src/main/scala/com/mesosphere/usi/core/models/RunSpec.scala
timcharper: Replaced with string shell command for now, with call out to above JIRAs

core/src/test/scala/com/mesosphere/usi/core/helpers/FakeMesos.scala
timcharper: lol maybe later hahaha lets merge this already

core/src/test/scala/com/mesosphere/usi/core/integration/SchedulerIntegrationTest.scala
timcharper: ✅

core/src/main/scala/com/mesosphere/usi/core/ResourceUtil.scala
timcharper: I didn't see this, I already migrated the tests.

core/src/main/scala/com/mesosphere/usi/core/SchedulerLogicHandler.scala
timcharper: Yeah, this seems correct to me. What's unclear?

core/src/test/scala/com/mesosphere/usi/core/helpers/FakeMesos.scala
timcharper: ¯\\\_(ツ)\_/¯ push a change to rename it to whatever you like

zen-dog:
I renamed the `FakeMesos` and pushed it however there is still a problem when building the project:

```
* What went wrong:
Execution failed for task ':core-models:compileScala'.
> class org.codehaus.groovy.runtime.GStringImpl cannot be cast to class java.lang.String (org.codehaus.groovy.runtime.GStringImpl is in unnamed module of loader java.net.URLClassLoader @34ce8af7; java.lang.String is in module java.base of loader 'bootstrap')
```

seems that the introduced silencer plugin exposes a groovy string somewhere and it is not properly converted to `java.lang.String`, however, I didn't have time to dig deeper.

zen-dog (approved):
Lgtm, so we could merge and move on. However, I'd like you to remove the `TODO`s from the scheduler code (there are 14 of those already) - let it fail explicitly for things that aren't implemented and/or create a JIRA ticket to track what needs to be done.

timcharper:
I reverted the silencer plugin. After refactoring the ResourceUtil to specifically target multi-role reservations and allocationInfo, the deprecation warnings are removed.

PR #42, opened 2019-02-15
Introduce a new `persistence-zookeeper` module

kaiwalyajoshi:
This series of patches introduces a new `persistence-zookeeper` module and transplants the following code from Marathon: 
- low-level ZK storage implementation aka `ZooKeeperPersistanceStore` and its helper classes
- a few additional testing classes like `DummyMetrics` and `Mockito`

Keep in mind: this is only the **low-level storage** classes. The concrete repository implementation (of e.g. `PodRecordRepository` and others) will follow in a separate PR.

JIRA Issues: DCOS-48210

Co-authored-by: Aleksey Dukhovniy <alex.dukhovniy@googlemail.com>


metrics/src/main/scala/com/mesosphere/usi/metrics/dummy/DummyMetrics.scala
```diff
@@ -0,0 +1,48 @@
+package com.mesosphere.usi.metrics.dummy
```
zen-dog: Should be part of `:test-utils`

persistence/build.gradle
```diff
@@ -1,3 +1,13 @@
 dependencies {
     compile project(':core-models')
-}
\ No newline at end of file
+    implementation project(':metrics')
+    implementation project(':test-utils')
+
+    compile "org.scala-lang.modules:scala-async_$scalaVersion:0.9.7"
+    compile "com.wix:accord-core_$scalaVersion:0.7.1"
```
zen-dog: Please, not the `wix-accord` :D

persistence/build.gradle
```diff
@@ -1,3 +1,13 @@
...
+
+    testCompile "com.typesafe.akka:akka-testkit_$scalaVersion:2.5.14"
```
zen-dog: testing libs should be defined for the whole project since multiple modules will use them.

zen-dog (changes requested):
A few things:

**modules**:
we decided to keep `:persistence` module about interfaces only. Specific implementation should be in a separate sub-project, in this case, I suggest `:persistence-zookeeper`

**scope**:
I would cut scope - you migrated not only the underlying storage `ZookeeperStorage, AsyncFactory` but also the first repository implementation on top of it e.g. `TemplateRepository` and the caching `PathTrie`. I would make it in a separate PR where we should discuss it.

**some files:**
where I don't think we should simply copy-paste them without revisiting the implementation: 
- `PathId` - parts of it are about relative paths - I don't believe we ever used them in Marathon and I don't think we will use them in the Usi
- `Mockito` - parts like `MockAnswer` or `Stubbed` are never even used in Marathon
-  `UnitTest` - already part of `test-utils`
- `DummyMetrics`- should be part of `test-utils`

**libraries:**
`wix-accord` - we pretty much universally hate this lib. it should not make its way into Usi

metrics/src/main/scala/com/mesosphere/usi/metrics/dummy/DummyMetrics.scala
jeschkies: Where is that used? I'm not sure I understand the relation to this pull request.

persistence/src/main/java/com/mesosphere/usi/storage/PathTrie.java
```diff
@@ -0,0 +1,371 @@
+package com.mesosphere.usi.storage;
```
jeschkies: Hm, I'm not sure this is the right place. Should we pull it into a separate pull request as @zen-dog suggested?

persistence/src/main/scala/com/mesosphere/usi/repository/TemplateRepository.scala
```diff
@@ -0,0 +1,204 @@
...
+  * It can be altered by setting `jute.maxbuffer` environment variable. Packet in this context is an application level
+  * packet containing all the children names. Some experimentation showed that each child element in that packet has ~4bytes
+  * overhead for encoding. So, let's say each child znode *name* e.g. holding Marathon app definition is 50 characters long
+  * (seems like a good guess given 3-4 levels of nesting e.g. `eng_dev_databases_my-favourite-mysql-instance`), we could only
+  * have: 4096 * 1024 / (50 + 4) =  ~75k children nodes until we hit the exception.
+  *
+  * In this class we implicitly rely on the users to *not* put too many children (apps/pods) under one parent. Since that
+  * number can be quite high (~75k) I think we are fine without implementing any guards.
+  */
+class TemplateRepository(val store: ZooKeeperPersistenceStore, val base: String)(implicit val mat: Materializer)
```
jeschkies: I'd drop it for now and concentrate on `ZooKeeperPersistenceStore`

persistence/src/main/scala/com/mesosphere/usi/state/plugin/PathId.scala
```diff
@@ -0,0 +1,23 @@
+package com.mesosphere.usi.state.plugin
+//package plugin
+
+import scala.collection.immutable.Seq
+
+/**
+  * A path based identifier.
+  */
+trait PathId {
```
jeschkies: I don't think we will follow this pattern. Eg the Jenkins plugin will probably use some kind of storage but not a `PathId`.

persistence/src/main/scala/com/mesosphere/usi/storage/zookeeper/AsyncCuratorBuilderFactory.scala
```diff
@@ -0,0 +1,267 @@
...
+  *
+  * @param createOptions a set with create options. By default node parents are created if needed and node data is compressed
+  * @param deleteOptions a set with delete options. By default node children are deleted if the node doesn't exist the delete method will appear to succeed
+  * @param existsOptions a set with exists options. By default no options are set
+  * @param createMode node create mode. By default [[CreateMode.PERSISTENT]] nodes are created
+  * @param acl node ACL. By default an empty list is used
+  * @param nodeVersion node version used to update node data or delete it. By default, -1 is used which matches all versions
+  * @param compressedData true if node data compression is enabled. Compression is enabled by default
+  */
+case class AsyncCuratorBuilderSettings(
```
jeschkies: Hm, should we load them from `reference.conf` and `application.conf` instead? @zen-dog already showed a way.

persistence/src/main/scala/com/mesosphere/usi/storage/zookeeper/ZooKeeperPersistenceStore.scala
```diff
@@ -0,0 +1,245 @@
+package com.mesosphere.usi.storage.zookeeper
```
jeschkies: This should probably be in `persistence-zookeeper` since @meln1k is really keen on creating a `persistence-foundationdb`

persistence/src/main/scala/com/mesosphere/usi/storage/zookeeper/PersistenceStore.scala
```diff
@@ -0,0 +1,115 @@
...
+
+  /**
+    * Main storage unit: it's a key-value like structure, where the key is the path
+    * to the node e.g. `/app1` which can also be nested e.g. `/apps/app1` and the data
+    * is the value (represented as ByteString)
+    *
+    * @param path node's path
+    * @param data node's payload
+    */
+  case class Node(path: String, data: ByteString)
```
jeschkies: I'm not sure this is the interface we want. Checkout `RecordRepository`. The base trait could be templated but I don't know what we gain. We should be fairly specific.

jeschkies (commented):
I have imagined the following steps:

1. Define a simple CRUD interface based on the `core-models`.
2. Implement this with a Zookeeper backend based on `ZooKeeperPersistenceStore`.

Step 1 basically takes the `TemplateRepository` but implements it for `core-models`. It's kind of the opposite of what the storage in Marathon is like. That was very generic on bytes. However, this reduced flexibility. We should not dictate the format.

persistence/build.gradle
timcharper: Yeah, I'm thinking this dep should be in test-utils

persistence/src/main/scala/com/mesosphere/usi/storage/zookeeper/PersistenceStore.scala
zen-dog: This is the **low-level storage interface** where (at least for ZK) we operate with bytes and string paths. 
Not to confuse with the **higher-level persistence interface** where we operate with core-types.

metrics/src/main/scala/com/mesosphere/usi/metrics/dummy/DummyMetrics.scala
zen-dog: Needed for testing of e.g. `ZookeeperPersistenceStore` which uses some metrics.

zen-dog:
I took the liberty of addressing a few feedback comments for this PR:

- I moved low-level ZK **storage** classes to the new `persistence-zookeeper` module
- Removed the `TemplateRepository`, `PathTrie` and other concrete **repository** implementation classes for now. Let's get aligned on how the repository interface looks like first and add a concrete implementation (based on that interface) in a separate PR
- removed unnecessary dependency libs like `wix-accord` lib and moved some testing classes to `test-utils` module

This should unblock us from merging this PR, so we can concentrate on discussing existing [RecordRepository](https://github.com/mesosphere/usi/blob/master/persistence/src/main/scala/com/mesosphere/usi/repository/RecordRepository.scala) class and it's two descendants: [PodRecordRepository](https://github.com/mesosphere/usi/blob/master/persistence/src/main/scala/com/mesosphere/usi/repository/PodRecordRepository.scala) and [ReservationRecordRepository](https://github.com/mesosphere/usi/blob/master/persistence/src/main/scala/com/mesosphere/usi/repository/ReservationRecordRepository.scala)

zen-dog (approved):
🚢 it

persistence/src/main/scala/com/mesosphere/usi/storage/zookeeper/AsyncCuratorBuilderFactory.scala
zen-dog: I thought about it but this is so low-level - I don't expect the users to change this  **ever**. I'd be in favor of leaving it like this for now and add a `conf` later should that ever be necessary.

persistence/src/main/scala/com/mesosphere/usi/state/plugin/PathId.scala
zen-dog: Removed

persistence/src/main/scala/com/mesosphere/usi/repository/TemplateRepository.scala
zen-dog: Done

persistence/src/main/java/com/mesosphere/usi/storage/PathTrie.java
zen-dog: Removed

persistence/build.gradle
zen-dog: Done

zen-dog:
Rebased on master

zen-dog:
> I have imagined the following steps:
> 
> 1. Define a simple CRUD interface based on the `core-models`.

This has already happened (priour to this PR) - see existing [RecordRepository](https://github.com/mesosphere/usi/blob/master/persistence/src/main/scala/com/mesosphere/usi/repository/RecordRepository.scala) class and it's two descendants: [PodRecordRepository](https://github.com/mesosphere/usi/blob/master/persistence/src/main/scala/com/mesosphere/usi/repository/PodRecordRepository.scala) and [ReservationRecordRepository](https://github.com/mesosphere/usi/blob/master/persistence/src/main/scala/com/mesosphere/usi/repository/ReservationRecordRepository.scala)


> 2\. Implement this with a Zookeeper backend based on `ZooKeeperPersistenceStore`.

this will follow in a separate PR

test-utils/build.gradle
```diff
@@ -1,6 +1,12 @@
 dependencies {
+    implementation project(':metrics')
+
     compile group: 'commons-io', name: 'commons-io', version: '2.6'
     compile group: 'de.heikoseeberger', name: "akka-http-play-json_$scalaVersion", version: '1.18.1'
     compile group: 'org.scala-lang.modules', name: "scala-async_$scalaVersion", version: '0.9.7'
     compile "org.scalatest:scalatest_$scalaVersion:3.0.5"
+    compile "org.mockito:mockito-all:1.10.19"
```
meln1k: What do you think about avoiding using mockito and deciding for a more scala-friendly library instead?

meln1k (changes requested):
If we can get rid of mockito it would be ✅

persistence-zookeeper/src/main/scala/com/mesosphere/usi/storage/zookeeper/PersistenceStore.scala
```diff
@@ -0,0 +1,115 @@
+package com.mesosphere.usi.storage.zookeeper
```
jeschkies, line 1: Feels like this is the interface that should be implemented.

test-utils/build.gradle
jeschkies: Fair point. I do have our own little wrapper but wonder if we should not migrate.

jeschkies (commented):
I'm a little confused. I thought this should implement the actual interface. Maybe that's a follow up.

test-utils/src/main/scala/com/mesosphere/utils/metrics/DummyMetrics.scala
```diff
@@ -0,0 +1,52 @@
...
+import com.mesosphere.usi.metrics.{ClosureGauge, Counter, Gauge, Meter, Metrics, SettableGauge, Timer}
+import com.mesosphere.usi.metrics.{UnitOfMeasurement => DropwizardUnitOfMeasurement}
+
+import scala.concurrent.Future
+
+/**
+  * As the name suggest, this is a dummy metrics implementation of the [[Metrics]] interface found in the `metrics`
+  * module. Should be used for testing only.
+  */
+object DummyMetrics extends Metrics {
```
jeschkies: This is not used anywhere,

meln1k (approved):

PR #43, opened 2019-02-19
Propose convenience methods for scalar resources

alenkacz:


core/src/main/scala/com/mesosphere/usi/core/matching/ScalarResource.scala
```diff
@@ -5,7 +5,7 @@ import com.mesosphere.usi.core.models.{ResourceMatchResult, ResourceRequirement,
 import scala.annotation.tailrec
 import org.apache.mesos.v1.{Protos => Mesos}
 
-case class ScalarResourceRequirement(resourceType: ResourceType, amount: Double) extends ResourceRequirement {
+case class ScalarResource(resourceType: ResourceType, amount: Double) extends ResourceRequirement {
```
timcharper, line 8: This is too long to type? :) I think I'd prefer to keep the word "Requirement" in the name as this is a requirement, and not a resource

alenkacz: yes, I think it's too long, for me personally that word is redundant :) it's of type resource 🤷‍♀️ 

When I saw the usage in our test it felt too verbose. `resourceRequirments=List(ScalarResourceRequirement(` vs `resourceRequirments=List(ScalarResource`

I find the second one as expressive :) but easier to grasp

zen-dog (approved):
Lgtm :ship:

timcharper (approved):

takirala (approved):

zen-dog (approved):

PR #44, opened 2019-02-20
 Add an `:examples/core-hello-world` project

zen-dog:
Summary:
Run the hello-world example framework that:
- relies on our `core` module to handle Mesos connection, event handling, offer matching etc. as
  opposed to the `simple-hello-world` example framework which uses only the Mesos client and has to
  implement all of the above-mentioned pieces itself
- starts one `echo "Hello, world" && sleep N` task
- exits when the task fails (or finishes)

JIRA: DCOS-47480


zen-dog:
There are a few issues to resolve before merging:
- we need to ACK Mesos status updates
- there seems to be a bug where we accept offers where there is nothing to launch
- core-hello-world needs an integration test

But otherwise, it looks good

examples/core-hello-world/src/main/resources/logback.xml
```diff
@@ -0,0 +1,18 @@
...
+
+    <!-- Some libraries are just too verbose -->
+    <logger name="akka" level="INFO"/>
+    <logger name="native-zk-connector" level="WARN" />
+    <logger name="org.apache.zookeeper" level="WARN" />
+
+    <root level="DEBUG">
+        <appender-ref ref="stdout"/>
+    </root>
+</configuration>
```
jeschkies, line 18: I think we should add the configuration for structured logging. What do you think, @takirala?

examples/core-hello-world/src/main/scala/com/mesosphere/usi/helloworld/CoreHelloWorld.scala
```diff
@@ -0,0 +1,166 @@
...
+    .newBuilder()
+    .setUser(
+      new SystemProperties()
+        .get("user.name")
+        .getOrElse(throw new IllegalArgumentException("A local user is needed to launch Mesos tasks")))
+    .setName("CoreHelloWorldExample")
+    .setFailoverTimeout(0d)
+    .build()
+
+  val client: MesosClient = Await.result(MesosClient(settings, frameworkInfo).runWith(Sink.head), 10.seconds)
```
jeschkies: Hm, I wonder if we really should provide a constructed client or if we should just inject an implementation. What do you think, @timcharper?

examples/core-hello-world/src/main/scala/com/mesosphere/usi/helloworld/CoreHelloWorld.scala
```diff
@@ -0,0 +1,166 @@
...
+    .prepend(Source.single(specsSnapshot))
+    .map(snapshot => (snapshot, Source.empty))
+    // Here our initial snapshot is going to the scheduler flow
+    .via(schedulerFlow)
+    // We flatten the output of the scheduler flow which is a tuple of an initial snapshot and a source of all
+    // later updates, into one stream where the first element is the snapshot and all later elements are single
+    // state events. This makes the event handling a simple match-case
+    .flatMapConcat {
+      case (snapshot, updates) =>
+        updates.prepend(Source.single(snapshot))
```
jeschkies: I don't really get this. Why isn't the scheduler flow prepending for us, @timcharper  and @alenkacz?

jeschkies (changes requested):
It feels this framework is kind of cheating. It does not tell me how I would send `PodUpdated` events through the scheduler interface. That seems to be the most important thing of our architecture.

examples/core-hello-world/src/main/resources/logback.xml
takirala: 👍 we have one example at https://github.com/mesosphere/usi/blob/master/examples/hello-world/src/main/resources/logback.groovy

alenkacz:
@zen-dog 

> There are a few issues to resolve before merging:
> 
> * we need to ACK Mesos status updates
> * there seems to be a bug where we accept offers where there is nothing to launch
> * core-hello-world needs an integration test
> 
> But otherwise, it looks good

ACKing is done here #50
Decline was implemented in #49

core/src/main/scala/com/mesosphere/usi/core/logic/MesosEventsLogic.scala
```diff
@@ -59,7 +59,8 @@ private[core] class MesosEventsLogic(mesosCallFactory: MesosCalls) extends Impli
             val taskInfos: List[Mesos.TaskInfo] = taskIdsFor(podSpec).map { taskId =>
               newTaskInfo(
                 taskId.asProto,
-                name = "hi",
+                // we use sanitized podId as the task name for now
+                name = podSpec.id.value.replaceAll("[^a-zA-Z0-9-]", ""),
```
alenkacz, line 63: :D

examples/core-hello-world/README.md
```diff
@@ -0,0 +1,8 @@
+# Core Hello World Example Framework
+
+Run the hello-world example framework that:
+- relies on our `core` module to handle Mesos connection, event handling, offer matching etc. as
+  opposed to the `simple-hello-world` example framework which uses only the Mesos client and has to 
+  implement all of the above mentioned pieces itself
+- starts one `echo "Hello, world" && sleep N` task
+- exits should the task fail (or fail to start)
```
alenkacz: I don't understand. Is this sentence correct? are you saying it exits if the task fail?

examples/core-hello-world/src/main/resources/logback.xml
alenkacz: yeah, let's go back to what @takirala did in that link, please :)

examples/core-hello-world/src/main/scala/com/mesosphere/usi/examples/CoreHelloWorldFramework.scala
```diff
@@ -0,0 +1,164 @@
...
+import scala.concurrent.duration._
+import scala.sys.SystemProperties
+
+/**
+  * Run the hello-world example framework that:
+  *  - relies on our `core` module to handle Mesos connection, event handling, offer matching etc. as
+  *    opposed to the `simple-hello-world` framework which uses only the Mesos client and has to implement
+  *    all of the above mentioned pieces itself
+  *  - starts one `echo "Hello, world" && sleep N` task
+  *  - exits should the task fail (or fail to start)
```
alenkacz: if you fix this sentence there, do it here as well :) thanks

examples/core-hello-world/src/main/scala/com/mesosphere/usi/examples/CoreHelloWorldFramework.scala
```diff
@@ -0,0 +1,164 @@
...
+    * +--------------------+         +-----------------------+        +-----------------------+
+    * |                    |         |                       |        |                       |
+    * |     SpecUpdated    |         |                       |        |      StateEvent       |
+    * |                    +--------->  Core SchedulerFlow   +-------->                       |
+    * |   new and updated  |         |                       |        | replicated Pod, Agent |
+    * |       PodSpecs     |         |                       |        | and Reservations state|
+    * +--------------------+         +-----------------------+        +-----------------------+
+    *
+    *
+    * In a nutshell: frameworks sends [[PodSpec]] updates down the flow and receives [[StateEvent]]s when things
```
alenkacz, line 76: nice docs 👍

alenkacz (approved):
No blockers, nice docs 👍  

only it would be nice to pull in the structured logging settings Tarun prepared, please.

examples/core-hello-world/README.md
zen-dog: Yep, it does.

zen-dog:
> only it would be nice to pull in the structured logging settings Tarun prepared, please.

I thought about that and decided not too. While structured logging is good for elk-style log analysis, when playing with those examples locally it is much clearer *not* to have it ;)

jeschkies (approved):

PR #45, opened 2019-02-21
Move design doc to USI

timcharper:
This will be the new home for the USI design doc. As we continue to develop and define the design further, we will update this doc.


kaiwalyajoshi:
Does it make sense to exclude the `*.svg` files from the project? They're artifacts of `plantuml` similar to any `.class` file

alenkacz (approved):
I did not re-read the whole thing but I am all in for moving the design doc to this repo

timcharper:
@kaiwalyajoshi having the files pre-rendered makes it visible from GitHub; I'd say it's convenient.

kaiwalyajoshi (approved):

PR #47, opened 2019-02-21
Decorate base test classes with RunWith

alenkacz:


zen-dog (approved):
Shouldn't we remove the `RunWith` annotation with existing tests then? Otherwise 👍

examples/mesos-client-example/src/test/scala/com/mesosphere/mesos/examples/MesosClientExampleFrameworkTest.scala
```diff
@@ -9,7 +9,6 @@ import org.junit.runner.RunWith
 import org.scalatest.junit.JUnitRunner
 import java.util
 
-@RunWith(classOf[JUnitRunner])
```
takirala, line 12: This means the imports for `RunWith` and `JUnitRunner` are unused and hence can be removed. I thought our scalaconf catches these (or does it ignore cuz these are test files ?!)

takirala (approved):
LGTM!

examples/mesos-client-example/src/test/scala/com/mesosphere/mesos/examples/MesosClientExampleFrameworkTest.scala
alenkacz: 🤔

meln1k (approved):

timcharper (approved):

examples/mesos-client-example/src/test/scala/com/mesosphere/mesos/examples/MesosClientExampleFrameworkTest.scala
alenkacz: btw. the build failed so everything is awesome. It was not failing only because the build was stopped on core tests failing so it did not get to this point (at least that's my understanding)

PR #48, opened 2019-02-22
Fix resourceutil test

alenkacz:
AllocationInfo is required by our `ResourceMatchKey` class. So we should set it in our builders or drop that requirment. This PR chose the first path...

@timcharper are we sure that there cannot be resource coming from mesos without the allocation info?

This test is still not run as part of our build though. That will happen with #47

### DIFF
--- a/core/src/main/scala/com/mesosphere/package.scala
+++ b/core/src/main/scala/com/mesosphere/package.scala
@@ -60,8 +60,8 @@ package mesosphere {
     *    // deleted immediately after. This happens every time the `logger` object is used to generate a log statement.
     * }}}
     */
-  case class LoggingArgs(args:(String, Any)*) {
-    def and(other:(String, Any)*): LoggingArgs = LoggingArgs(args ++ other:_*)
+  case class LoggingArgs(args: (String, Any)*) {
+    def and(other: (String, Any)*): LoggingArgs = LoggingArgs(args ++ other: _*)
   }
 
   /**
--- a/core/src/main/scala/com/mesosphere/usi/core/protos/ProtoBuilders.scala
+++ b/core/src/main/scala/com/mesosphere/usi/core/protos/ProtoBuilders.scala
@@ -91,7 +91,12 @@ private[usi] object ProtoBuilders {
 
     reservations.foreach(b.addReservations)
 
-    if (allocationInfo != null) b.setAllocationInfo(allocationInfo)
+    if (allocationInfo != null) {
+      b.setAllocationInfo(allocationInfo)
+    } else {
+      // allocation info is required property
+      b.setAllocationInfo(Mesos.Resource.AllocationInfo.newBuilder())
+    }
     if (disk != null) b.setDisk(disk)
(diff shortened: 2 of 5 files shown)


core/src/main/scala/com/mesosphere/usi/core/protos/ProtoBuilders.scala
```diff
@@ -91,7 +91,12 @@ private[usi] object ProtoBuilders {
 
     reservations.foreach(b.addReservations)
 
-    if (allocationInfo != null) b.setAllocationInfo(allocationInfo)
+    if (allocationInfo != null) {
```
takirala: `allocationInfo` is also used in `newOffer` method in the same class, may want to change there as well.

core/src/test/scala/com/mesosphere/usi/core/ResourceUtilTest.scala
```diff
@@ -40,15 +40,15 @@ class ResourceUtilTest extends UnitTest {
   }
 
   "ResourceUtil" should {
-    "no base resources" in {
+    "have no leftover for empty resources" in {
```
takirala:
```suggestion
    "have no leftOvers for empty resources" in {
```

core/src/main/scala/com/mesosphere/package.scala
```diff
@@ -60,8 +60,8 @@ package mesosphere {
     *    // deleted immediately after. This happens every time the `logger` object is used to generate a log statement.
     * }}}
     */
-  case class LoggingArgs(args:(String, Any)*) {
-    def and(other:(String, Any)*): LoggingArgs = LoggingArgs(args ++ other:_*)
+  case class LoggingArgs(args: (String, Any)*) {
+    def and(other: (String, Any)*): LoggingArgs = LoggingArgs(args ++ other: _*)
```
takirala, line 64: Thank you, i have somehow missed this when i first added the file. 
we can revisit this to automate in our scalafmt conf - may be when you bump your scalafmt plugin :)

takirala (approved):
LGTM.

core/src/main/scala/com/mesosphere/usi/core/protos/ProtoBuilders.scala
timcharper: If it's required lets just express the requirement in the method.

core/src/main/scala/com/mesosphere/usi/core/protos/ProtoBuilders.scala
```diff
@@ -48,7 +48,12 @@ private[usi] object ProtoBuilders {
     executorIds.foreach(b.addExecutorIds)
     resources.foreach(b.addResources)
 
-    if (allocationInfo != null) b.setAllocationInfo(allocationInfo)
+    if (allocationInfo != null) {
```
timcharper: If it's required lets just express the requirement in the method.

meln1k (approved):

timcharper (approved):

PR #49, opened 2019-02-22
Decline unused/insufficient offers

takirala:
- Added a new FrameworkMock.scala to mock all the objects that are usually created by a framework.
- Added couple debug statements in the code, please comment if the log level/data needs to be changed.

JIRA Issues: DCOS-48847


timcharper (approved):

core/src/main/scala/com/mesosphere/usi/core/logic/MesosEventsLogic.scala
```diff
@@ -87,17 +87,31 @@ private[core] class MesosEventsLogic(mesosCallFactory: MesosCalls) extends Impli
...
+        offer.getId.getValue,
+        if (specs.isEmpty) "as there are no specs to be launched"
+        else s"due to unmet requirement for pods : [${specs.map(_.id.value).mkString(", ")}]")
+      mesosCallFactory.newDecline(Seq(offer.getId))
+    } else {
+      val op = newOfferOperation(
+        Mesos.Offer.Operation.Type.LAUNCH,
+        launch = newOfferOperationLaunch(taskInfos.values.flatten)
+      )
+      logger.debug(s"Launching taskId${if (op.getLaunch.getTaskInfosCount > 1) "s"} : [{}] for offerId {}",
```
alenkacz: what about some structured logging here? :)

core/src/main/scala/com/mesosphere/usi/core/logic/MesosEventsLogic.scala
```diff
@@ -87,17 +87,31 @@ private[core] class MesosEventsLogic(mesosCallFactory: MesosCalls) extends Impli
...
-    val op =
-      newOfferOperation(Mesos.Offer.Operation.Type.LAUNCH, launch = newOfferOperationLaunch(taskInfos.values.flatten))
-
-    val acceptBuilder = MesosCall.Accept
-      .newBuilder()
-      .addOperations(op)
-      .addOfferIds(offer.getId)
-
-    val intents = intentsBuilder.withMesosCall(mesosCallFactory.newAccept(acceptBuilder.build))
+    val offerIntent = if (taskInfos.isEmpty) {
```
alenkacz: we renamed the output parameter to `SchedulerEventsBuilder` so I think we should stop using intents also for names of the variables

core/src/main/scala/com/mesosphere/usi/core/logic/MesosEventsLogic.scala
```diff
@@ -108,9 +122,10 @@ private[core] class MesosEventsLogic(mesosCallFactory: MesosCalls) extends Impli
         val (schedulerEventsBuilder, _) =
           offersList.asScala.foldLeft((SchedulerEventsBuilder.empty, pendingLaunch)) {
             case ((builder, pending), offer) =>
-              val (matchedPodIds, offerMatchSchedulerEvents) = matchOffer(offer, pendingLaunch.flatMap { podId =>
-                specs.podSpecs.get(podId)
-              }(collection.breakOut))
+              val (matchedPodIds, offerMatchSchedulerEvents) = matchOffer(
```
alenkacz, line 133: why did we drop the `(collection.breakOut)` ?

core/src/test/scala/com/mesosphere/usi/core/helpers/FrameworkMock.scala
```diff
@@ -0,0 +1,25 @@
...
+import org.apache.mesos.v1.{Protos => Mesos}
+
+object FrameworkMock {
+
+  val mockPodId: PodId = PodId("mocked-podId")
+  val mockRunSpec: RunSpec = RunSpec(
+    List(ScalarResourceRequirement(ResourceType.CPUS, 1), ScalarResourceRequirement(ResourceType.MEM, 256)),
+    shellCommand = "sleep 3600")
+
+  val mockPodSpec: PodSpec = PodSpec(
```
alenkacz: I am not a big fan of these factory methods for tests, especially if we are still exploring the API.

Also I would say that mock is something, that simulates something else (real). I would say instantiating a PodSpec is not mocking...

My vote would be to move those back to the code itself.

We had pretty huge test builders in Marathon and we're slowly moving out of those and are exercising real code in tests.

Also when using the complete definition it's easier to understand the tests when reading them.

core/src/test/scala/com/mesosphere/usi/core/logic/MesosEventsLogicTest.scala
```diff
@@ -0,0 +1,60 @@
...
+import com.mesosphere.usi.core.models.PodId
+import com.mesosphere.usi.core.models.ResourceType
+import com.mesosphere.utils.UnitTest
+
+class MesosEventsLogicTest extends UnitTest {
+
+  private val mesosEventLogic = new MesosEventsLogic(new MesosCalls(FrameworkMock.mockFrameworkId))
+
+  "MesosEventsLogic" should {
+    "decline an offer when PodSpec's are empty" in {
```
alenkacz: Podspecs... maybe we can do `decline an offer when there is no podspec to launch`? I feel like that's less implementation specific name :)

core/src/test/scala/com/mesosphere/usi/core/logic/MesosEventsLogicTest.scala
```diff
@@ -0,0 +1,60 @@
...
+      )
+      schedulerEventsBuilder.result.mesosCalls.size shouldBe 1
+      val declines = schedulerEventsBuilder.result.mesosCalls.head.getDecline.getOfferIdsList
+      declines.size() shouldBe 1
+      declines.get(0) shouldEqual MesosMock.mockOffer.getId
+    }
+
+    "accept an offer when some PodSpec's resource requirements are met" in {
+      val (matchedPodIds, schedulerEventsBuilder) = mesosEventLogic matchOffer(
+        MesosMock.mockOffer,
```
alenkacz: e.g. here ... so I am reading the test and I am immediately thinking - how do we know that something called mockOffer has enough resources to launch *4* pods. What if this starts to fail with `matchedPodIds.size is 2, expected 4`? What if I want to launch 6 pods and with specific resources? Do I do `mockOfferWith8CpusAnd6Gpus`? :)

alenkacz (approved):
I would vote for not introduce "mock" implementations for tests right now. I think that could easily be a small functions within the test classes, exercising the real code. I think those does not have to be shared as different tests will need different pods.

core/src/test/scala/com/mesosphere/usi/core/logic/MesosEventsLogicTest.scala
takirala: created a method to parametrize different dimensions in an offer. we can add more arguments as we needed. lmk if this is not ideal.

core/src/main/scala/com/mesosphere/usi/core/logic/MesosEventsLogic.scala
takirala: using the breakout we would get a `Seq[PodSpec]` instead of `Set[PodSpec]`. is there any advantage in doing that since we can already pass the `Set` as-is ?

PR #50, opened 2019-02-25
Accept mesos task status update

alenkacz:
We need this to be able to write proper integration test against mesos (and hello world framework).

JIRA: Partially addresses DCOS-48846


core/src/main/scala/com/mesosphere/usi/core/logic/MesosEventsLogic.scala
```diff
@@ -128,7 +131,11 @@ private[core] class MesosEventsLogic(mesosCallFactory: MesosCalls) extends Impli
               PodStatus(podId, Map(taskId -> taskStatus))
           }
 
-          SchedulerEvents(stateEvents = List(PodStatusUpdated(podId, Some(newState))))
+          SchedulerEvents(
```
alenkacz, line 157: we need to test this. But there is another PR from @takirala  introducing test class for this class so I'll not do it now to avoid merge conflicts :)

timcharper:
@alenkacz marking this as WIP since there's the pending test

core/src/main/scala/com/mesosphere/usi/core/logic/MesosEventsLogic.scala
```diff
@@ -119,6 +121,7 @@ private[core] class MesosEventsLogic(mesosCallFactory: MesosCalls) extends Impli
       case UpdateEvent(taskStatus) =>
         val taskId = TaskId(taskStatus.getTaskId.getValue)
         val podId = podIdFor(taskId)
+        logger.info(s"Received task status update from taskId $taskId and podId $podId with status ${taskStatus.getState}", LoggingArgs(("taskId", taskId), ("podId", podId)))
```
takirala: since we are using LoggerTakingImplicit, the LoggingArgs needs to be curried to `info` method [like this](https://github.com/mesosphere/usi/blob/9cb81b020c323faf20e951284db536ee805a815d/core/src/main/scala/com/mesosphere/package.scala#L56):
```
logger.info(s"Received task status update from taskId $taskId and podId $podId with status ${taskStatus.getState}")(LoggingArgs("taskId" -> taskId).and("podId"->podId))
```

jeschkies:
Just a nit. You don't need to label it as `WIP` since Github has `Draft Pull Requests` now. See https://github.com/mesosphere/usi/pull/51.

core/src/main/scala/com/mesosphere/usi/core/logic/MesosEventsLogic.scala
alenkacz: oh right 🙈 oh wow, that's error prone!

alenkacz:
I am not sure I'll be able to finish this until Thursday, when I am out of maintenance so if you have spare cycles, feel free to pick it up...

takirala:
@alenkacz i worked on this today to push this forward to unblock the integration tests.
@timcharper this PR addresses a very small part of DCOS-48846, i will follow up with another PR for the remainder of it and add more test cases as well. 
PTAL

meln1k (approved):

timcharper (approved):
👏 👏 👏

PR #51, opened 2019-02-26
Define Zookeeper storage backend for repositories.

jeschkies:


persistence-zookeeper/src/main/scala/com/mesosphere/usi/storage/zookeeper/PodRecordRepository.scala
```diff
@@ -0,0 +1,48 @@
...
+import scala.util.{Failure, Success}
+
+class PodRecordRepository(val store: PersistenceStore) extends PodRecordRepositoryInterface {
+
+  override def create(record: PodRecord): Future[PodId] = async {
+    val data = record.toString
+    val node = Node(record.podId.value, ByteString(data))
+    await(store.create(node))
+    record.podId
+    // TODO: RecordAlreadyExistsException
```
jeschkies: So yeah. The interface states that the future should fail if the record already exists. How are we going to do that? Should we use a transaction here? The `TemplateRepository` does not have this behaviour as far as I know. It uses the `PathTrie` which synchronizes requests.

persistence-zookeeper/src/main/scala/com/mesosphere/usi/storage/zookeeper/PodRecordRepository.scala
```diff
@@ -0,0 +1,48 @@
...
+  }
+
+  override def read(recordId: PodId): Future[Option[PodRecord]] = async {
+    // We uncheck our match because the only failure case is the [[NoNodeException]]. The future fails in all other cases.
+    (await(store.read(recordId.value)): @unchecked) match {
+      case Success(node) =>
+        val record: PodRecord = ??? // deserialize node
```
jeschkies: What should we use for serialization?

persistence/src/main/scala/com/mesosphere/usi/repository/ReservationRecordRepository.scala
```diff
@@ -5,7 +5,4 @@ import com.mesosphere.usi.core.models.{ReservationId, ReservationRecord}
 /**
   * Repository for [[ReservationRecord]].
   */
-trait ReservationRecordRepository extends RecordRepository {
-  override type Record = ReservationRecord
-  override type RecordId = ReservationId
-}
+trait ReservationRecordRepository extends RecordRepository[ReservationRecord, ReservationId]
```
jeschkies, line 8: @meln1k, what was the reason to use `type` instead of a template?

jeschkies (commented):
This is the very first draft. Tests are missing and some design decisions need to be made.

persistence-zookeeper/src/main/scala/com/mesosphere/usi/storage/zookeeper/PodRecordRepository.scala
```diff
@@ -0,0 +1,48 @@
...
+    val data = record.toString
+    val node = Node(record.podId.value, ByteString(data))
+    await(store.create(node))
+    record.podId
+    // TODO: RecordAlreadyExistsException
+  }
+
+  override def read(recordId: PodId): Future[Option[PodRecord]] = async {
+    // We uncheck our match because the only failure case is the [[NoNodeException]]. The future fails in all other cases.
+    (await(store.read(recordId.value)): @unchecked) match {
```
jeschkies: I'm inclined to change the storage read method. @zen-dog, what was the motivation of `Try` over `Option`?

persistence/src/main/scala/com/mesosphere/usi/repository/ReservationRecordRepository.scala
meln1k: I don't want to carry generic parameters everywhere since they look messy and don't provide real value in this case. Once they're introduced, all code that works with the repository will have to deal with them, and it's going to complicate things without providing any real benefits.

persistence-zookeeper/src/main/scala/com/mesosphere/usi/storage/zookeeper/PodRecordRepository.scala
meln1k: I think we must use a transaction here.

persistence-zookeeper/src/main/scala/com/mesosphere/usi/storage/zookeeper/PodRecordRepository.scala
meln1k: Im my view sticking to protobuf is a solid choice.

persistence-zookeeper/src/main/scala/com/mesosphere/usi/storage/zookeeper/PodRecordRepository.scala
```diff
@@ -0,0 +1,48 @@
...
+
+import scala.async.Async.{async, await}
+import scala.concurrent.ExecutionContext.Implicits.global
+import scala.concurrent.Future
+import scala.util.{Failure, Success}
+
+class PodRecordRepository(val store: PersistenceStore) extends PodRecordRepositoryInterface {
+
+  override def create(record: PodRecord): Future[PodId] = async {
+    val data = record.toString
```
meln1k: We need a proper serialisation here.

meln1k (changes requested):

persistence/src/main/scala/com/mesosphere/usi/repository/ReservationRecordRepository.scala
jeschkies: Not sure. The type alias can cause confusion. Especially since it's could be used in the implementation and one would have two `Record` and `RecordId` types.

persistence-zookeeper/src/main/scala/com/mesosphere/usi/storage/zookeeper/PodRecordRepository.scala
jeschkies: Yup. I was just not sure which one we want.

persistence-zookeeper/src/test/scala/com/mesosphere/usi/repository/InMemoryRepositoryTest.scala
```diff
@@ -0,0 +1,54 @@
...
+import org.junit.runner.RunWith
+import org.scalatest.junit.JUnitRunner
+
+import scala.collection.mutable
+import scala.concurrent.Future
+
+@RunWith(classOf[JUnitRunner])
+class InMemoryRepositoryTest extends UnitTest with RepositoryBehavior with StrictLogging {
+
+  case class InMemoryRepository() extends PodRecordRepository {
```
jeschkies: I wonder where we should put this.

persistence-zookeeper/src/test/scala/com/mesosphere/usi/repository/RepositoryBehavior.scala
```diff
@@ -0,0 +1,131 @@
...
+    "delete an unknown record" in {
+      Given(s"an unknown record id")
+      val repo = newRepo
+      val unknownPodId = PodId("unknown")
+
+      When("the unknown record is  deleted")
+      val result = repo.delete(unknownPodId).failed.futureValue
+
+      Then("an error is returned")
+      result should be(RecordNotFoundException(unknownPodId.value))
```
jeschkies: This is a little strange to me. Why is `read` returning an `Option` but here we throw an exception?

persistence/src/main/scala/com/mesosphere/usi/repository/RecordRepository.scala
```diff
@@ -34,13 +30,13 @@ trait RecordRepository {
...
     * @return id of the deleted record
     */
-  def delete(record: Record): Future[RecordId]
+  def delete(recordId: RecordId): Future[RecordId]
 }
 
-class RecordAlreadyExistsException(id: String) extends RuntimeException(s"record with id $id already exists.")
-class RecordNotFoundException(id: String) extends RuntimeException(s"record with id $id doesn't exist.")
+case class RecordAlreadyExistsException(id: String) extends RuntimeException(s"record with id $id already exists.")
+case class RecordNotFoundException(id: String) extends RuntimeException(s"record with id $id doesn't exist.")
```
jeschkies, line 43: I don't think we should have this.

persistence/src/main/scala/com/mesosphere/usi/repository/ReservationRecordRepository.scala
jeschkies: I don't think there would be code that uses the `Repository` trait directly. I might be wrong though. Of course `repo: Repository[_, _]` would work. It actually gives more control in case we want `repo: Repository[CoreId, _]`

persistence-zookeeper/src/main/scala/com/mesosphere/usi/storage/zookeeper/PodRecordRepository.scala
zen-dog: To distinguish between empty nodes and failed reads?

persistence-zookeeper/src/main/scala/com/mesosphere/usi/storage/zookeeper/PodRecordRepository.scala
zen-dog: > The `TemplateRepository` does not have this behavior as far as I know.

it does. creating a node with an existing id/path is invalid and should fail.

persistence-zookeeper/src/main/scala/com/mesosphere/usi/storage/zookeeper/PodRecordRepository.scala
```diff
@@ -0,0 +1,48 @@
...
+    val node = Node(record.podId.value, ByteString(data))
```
zen-dog: Keep in mind - you want to store more entity types than just `PodRecord`s. You need some prefix for all _PodRecords_ like `/records` so that resulting path looks like `/records/eng/sleep`

persistence-zookeeper/src/main/scala/com/mesosphere/usi/storage/zookeeper/PodRecordRepository.scala
zen-dog: There is only one `create` request here - so what should a transaction do?

persistence-zookeeper/src/main/scala/com/mesosphere/usi/storage/zookeeper/PodRecordRepository.scala
zen-dog: Also: you're using `podId.value` as a node path - does this mean, there is always only 1 `PodRecord` for a given `podId`? What about pod versions? (podId stays the same, but RunSpec changes)

persistence-zookeeper/src/main/scala/com/mesosphere/usi/storage/zookeeper/PodRecordRepository.scala
jeschkies: First. You are on vacation so why should I listen to you 😉 

Second. Show me the code or it did not happen. The `TemplateRepository.create` does not check if the node exists. Maybe this is something in Curator that throws an exception. I'm going to test that today.

persistence-zookeeper/src/main/scala/com/mesosphere/usi/storage/zookeeper/PodRecordRepository.scala
jeschkies: `Failure(NoNodeException)` is a missing node. IMHO this would be a `None`.

persistence-zookeeper/src/main/scala/com/mesosphere/usi/storage/zookeeper/PodRecordRepository.scala
jeschkies: We have to talk about this. I don't think `PodRecord`s will have paths as their id. We can have versioning.

persistence-zookeeper/src/test/scala/com/mesosphere/usi/storage/zookeeper/ZooKeeperPersistenceStoreTest.scala
```diff
@@ -186,7 +186,7 @@ class ZooKeeperPersistenceStoreTest extends UnitTest with ZookeeperServerTest wi
         store.delete(path).futureValue shouldBe path
       }
 
-      "delete a non-existing node does not lead to an exception" in {
+      "delete a non-existing node does not lead to an exception" ignore {
```
jeschkies: So what should it be @zen-dog and @kaiwalyajoshi? Should we fail or not when deleting an unknown record?

persistence-zookeeper/src/main/scala/com/mesosphere/usi/storage/zookeeper/PodRecordRepository.scala
zen-dog: > Maybe this is something in Curator that throws an exception.

Exactly ;) A `NodeExistsException` is thrown if you try to create an existing node. A `NoNodeException` is thrown if you try to update a **non-existing** node. Both makes total sense as far as API semantics go

persistence-zookeeper/src/test/scala/com/mesosphere/usi/storage/zookeeper/ZooKeeperPersistenceStoreTest.scala
zen-dog: We should **not**. The end result is the same - node is not there. `delete a non-existing node does not lead to an exception` is correct.

persistence-zookeeper/src/main/scala/com/mesosphere/usi/storage/zookeeper/PodRecordRepository.scala
zen-dog: I might've misunderstood the question - where is the `Try` you talk about? 

To recapture:
- `Future` success/failure signals whether the node exists or not
- `Option` signals whether the node was empty.

persistence-zookeeper/src/main/scala/com/mesosphere/usi/storage/zookeeper/PodRecordRepository.scala
jeschkies: Found it. Thanks

buildSrc/src/main/scala/mesosphere/gradle/protobuf/ScalaPb.scala
```diff
@@ -0,0 +1,59 @@
...
+
+import scala.beans.BeanProperty
+import collection.JavaConverters._
+
+/**
+  * Gradle task that runs ScalaPb protobuf code generation.
+  *
+  * @see https://scalapb.github.io/scalapbc.html
+  * @see https://github.com/etiennestuder/gradle-jooq-plugin/
+  * @see https://github.com/google/protobuf-gradle-plugin
```
jeschkies: I think we really should model it after the Google Protobuf plugin and go with something like

```groovy
project(':persistence-zookeeper') {
  apply plugin: 'com.mesosphere.protobuf'
}
```

this should then create a `proto` source set.

build.gradle
```diff
@@ -97,4 +99,9 @@ idea {
     languageLevel = '1.11'
     vcs = 'Git'
   }
+  module {
+      def generatedZooKeeper = Paths.get(project(':persistence-zookeeper').protobuf.generatedFilesBaseDir, "main", "java")
+      sourceDirs += file(generatedZooKeeper)
+      generatedSourceDirs+= file(generatedZooKeeper)
```
jeschkies, line 111: IntelliJ code completion does not work even though all docs say this is the way to do it. I'm happy for any help here.

jeschkies (commented):
Handing over to @kaiwalyajoshi.

## Open items
- [ ] Design more behavior in `RepositoryBehavior` and adapt the repository and maybe the ZooKeeper client. I'm not sure eg that we should map from `Try` to `Option`.
- [ ] Implement `ReservationRecordRepository`.
- [ ] Decide on `PodId` structure. Should it be a path?
- [ ] Discuss bucketing with @zen-dog.
- [ ] Add caching to both repository implementations. I don't think we need the `PathTrie` if we don't have paths.

jeschkies:
@kaiwalyajoshi, @takirala, @zen-dog and @timcharper, I'm closing this draft as it is superseded by Tarun's work. This one also grew out of proportion and has many design decisions. It can still be a guideline for future implementations.

PR #52, opened 2019-02-26
Continue even on test failure.

jeschkies:
Right now, if the test fails, we won't get the test report generated and uploaded so we can't see the html test results page as a part of the PR. This PR fixes that.

JIRA issues: DCOS-49176

Co-authored-by: Alena Varkockova <varkockova.a@gmail.com>

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -70,7 +70,7 @@ allprojects {
 
     test {
         reports.html.enabled = false
-        reports.junitXml.destination = file(Paths.get(rootProject.buildDir.toString(), "tests", project.name))
+        ignoreFailures true
     }
     check {}
 }
--- a/build.gradle
+++ b/build.gradle
@@ -70,7 +70,6 @@ allprojects {
 
     test {
         reports.html.enabled = false
-        ignoreFailures true
     }
     check {}
 }
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,7 +5,7 @@ language: scala
 scala:
    - 2.12.7
 script:
-  - sudo ./gradlew provision && ./gradlew ci --info
+  - sudo ./gradlew provision && ./gradlew ci --info --continue
 before_cache:
   - rm -f  $HOME/.gradle/caches/modules-2/modules-2.lock
   - rm -fr $HOME/.gradle/caches/*/plugin-resolution/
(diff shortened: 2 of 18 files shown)


jeschkies:
~This is not quite correct yet. https://stackoverflow.com/questions/16921384/aggregating-gradle-multiproject-test-results-using-testreport~

jeschkies:
[Here](https://android.googlesource.com/platform/tools/build/+/925c0b5cf8730105dd5aa8c851141d5688d07789/gradle/src/main/groovy/com/android/build/gradle/ReportingPlugin.groovy) is the aggregation plugin for Android. It [set the ignore failures flag as well](https://android.googlesource.com/platform/tools/build/+/925c0b5cf8730105dd5aa8c851141d5688d07789/gradle/src/main/groovy/com/android/build/gradle/internal/tasks/AndroidReportTask.groovy#64).

jeschkies:
So it [works](https://s3-us-west-2.amazonaws.com/usi-builds/mesosphere/usi/karsten/enable-failed-test-report/563/reports/tests/index.html). Travis CI and our custom reporter should report a failure, though.

jeschkies:
It seems the Android plugin [re-throws the test failure](https://android.googlesource.com/platform/tools/build/+/925c0b5cf8730105dd5aa8c851141d5688d07789/gradle/src/main/groovy/com/android/build/gradle/internal/tasks/AndroidReportTask.groovy#85).

build.gradle
```diff
@@ -70,7 +70,7 @@ allprojects {
 
     test {
         reports.html.enabled = false
-        reports.junitXml.destination = file(Paths.get(rootProject.buildDir.toString(), "tests", project.name))
+        ignoreFailures true
```
alenkacz: I think you don't want to do this, then if we run test locally, it would not fail

I think we should use `-Dtest.ignoreFailures=true` just for travis

alenkacz:
this might still not work because of 

```
> Task :ci
Skipping task ':ci' as it has no actions.
:ci (Thread[Daemon worker,5,main]) completed. Took 0.0 secs.
Closing Git publish repo: /home/travis/build/mesosphere/usi/build/gitPublish
```

alenkacz (approved):
I think this is an awesome code and it should totally get merged :DD

kensipe (approved):

PR #53, opened 2019-02-26
In CI, fail on incorrect formatting

alenkacz:
As we agreed in Slack, we should fail the build on incorrect formatting. At the same time we want compile to automatically format the code -> we need to run this before the ci pipeline starts.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,7 +5,7 @@ language: scala
 scala:
    - 2.12.7
 script:
-  - sudo ./gradlew provision && ./gradlew ci --info
+  - sudo ./gradlew provision && ./gradlew checkScalaFmtAll ci --info
 before_cache:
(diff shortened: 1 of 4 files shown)


alenkacz:
ok this is passing and it should not be passing (thx. @takirala ) so I am investigating...

.travis.yml
```diff
@@ -5,7 +5,7 @@ language: scala
 scala:
    - 2.12.7
 script:
-  - sudo ./gradlew provision && ./gradlew ci --info
+  - ./gradlew checkScalaFmtAll && sudo ./gradlew provision && ./gradlew ci --info
```
jeschkies, line 12: Nice. I did not know there was a target for it.

jeschkies (approved):

takirala (approved):

PR #54, opened 2019-02-26
Added a first draft of the keepAlive mechanism for a pod.

meln1k:
# Hello world with keep alive example
## What is this
This is an example of a keep alive mechanism with the existing scheduler.
## Purpose of this PR
I want to grab some attention and discuss the current scheduler interface. Currently it's not that elegant, and to integrate with it we need some workarounds such as queues. 

This PR shows the challenges of integration with the current version.

I think we should discuss a better interface and rewrite it to make it simpler.
## How it works
It's similar to @zen-dog PR, but instead of crashing on pod failure we increase the incarnation number and reschedule pod. Underneath, it has a flow that consumes state updates and transforms them into a stream of pod spec changes.
## What to look at
Have a look on keepAliveWatcher flow, as well as Sink/Source queues used for integrating with the scheduler. 

JIRA issues: DCOS-48963


examples/hello-world/src/main/resources/logback.groovy
```diff
@@ -1,127 +0,0 @@
-import net.logstash.logback.appender.LoggingEventAsyncDisruptorAppender
```
alenkacz: why?

examples/keep-alive-framework/README.md
```diff
@@ -0,0 +1,8 @@
+# Keep-Alive Example Framework
+
+Run the keep-alove example framework that:
```
alenkacz: alive

examples/keep-alive-framework/src/main/scala/com/mesosphere/usi/helloworld/KeepAliveFramework.scala
```diff
@@ -0,0 +1,171 @@
...
+
+
+  keepAliveStateSource.via(keepAliveWatcher).to(keepAliveSpecSink).run()
+
+
+  val running = Source.single(specsSnapshot)
+    // A trick to make our stream run, we use a workaround to transform the sink of our keep alive mechanism into
+    // a source. If we change the scheduler interface this will be much more elegant
+    .map(snapshot => (snapshot, Source.unfoldResourceAsync[SpecUpdated, SinkQueueWithCancel[SpecUpdated]](
+      () => Future.successful(keepAliveSpecQueue),
```
jeschkies: Why do we need this? State updates should just be read, shouldn't they?

jeschkies (commented):
First off, many thanks for this example. It is a really great use case. I have to admit that I don't get it though. I don't see where you create a new pod spec. Shouldn't there be a `SpecEvent`? Is it just me or is the interface confusing?

Is there a way we would not leak the Akka stream implementation? I wonder if we could offer both like the Zookeeper client from #51.

meln1k:
I'm closing this PR in favor of #82

PR #55, opened 2019-02-28
Do not expect TASK_RUNNING to be the first task status

alenkacz:
With this PR and #50 the integration test will be finally green!


alenkacz:
btw. this test is *NOT* run in our CI right now. that will be fixed in #47 but I cannot merge that before the tests are green... LOL

timcharper (approved):

kvish (approved):

PR #56, opened 2019-03-04
Add ranged/ports resource matching

alenkacz:
This reuses a lot of code from Marathon. 

It implements following two strategies:
- Match exactly these port numbers
- Take any random X ports

Take N consecutive ports (stretch goal) will be implemented in another PR.

It does not work with reservations and roles right now.

JIRA: DCOS-47481


core/src/main/scala/com/mesosphere/usi/core/matching/PortResource.scala
```diff
@@ -0,0 +1,223 @@
...
+    * The implementation idea:
+    *
+    * * Randomize the order of the offered ranges.
+    * * Now treat the ports contained in the ranges as one long sequence of ports.
+    * * We randomly choose an index where we want to start assigning dynamic ports in that sequence. When
+    *   we hit the last offered port with wrap around and start offering the ports at the beginning
+    *   of the sequence up to (excluding) the port index we started at.
+    * * The next range is determined on demand. That's why an iterator is returned.
+    */
+  private def lazyRandomPortsFromRanges(offeredPortRanges: Seq[PortRange], rand: Random): Iterator[Int] = {
```
alenkacz: this is reused from marathon. It's up for discussion whether we need it. When I discussed this with Tim we just noted that the motivations described in this comment seems to be valid...

core/src/main/scala/com/mesosphere/usi/core/matching/PortResource.scala
```diff
@@ -0,0 +1,223 @@
...
+import com.mesosphere.usi.core.matching.PortResource._
+import com.mesosphere.usi.core.models.{ResourceMatchResult, ResourceRequirement, ResourceType}
+import com.mesosphere.usi.core.protos.ProtoBuilders
+import org.apache.mesos.v1.{Protos}
+
+import scala.annotation.tailrec
+import scala.collection.JavaConverters._
+import scala.util.Random
+
+case class PortResource(ports: Seq[Int], random: Random = Random) extends ResourceRequirement {
```
meln1k: It sounds like `*Resource` shouldn't be a subtype of a `*Requirement`. Shall we use a different naming?

core/src/main/scala/com/mesosphere/usi/core/matching/PortResource.scala
```diff
@@ -0,0 +1,223 @@
...
+  val RandomPort: Int = 0
+
+  /**
+    * Return RangesResources covering all given ports with the given roles.
+    *
+    * Creates as few RangesResources as possible while
+    * preserving the order of the ports.
+    */
+  def createPortsResources(
+      originalResource: Protos.Resource,
```
meln1k: Can we name it differently?

core/src/main/scala/com/mesosphere/usi/core/matching/PortResource.scala
```diff
@@ -0,0 +1,223 @@
...
+      requestedPorts: Seq[Int],
+      resourceType: ResourceType): Seq[Protos.Resource] = {
```
meln1k: Why do we need a `ResourceType` here if the method already says that it will create PortsResources?

jeschkies:
> Take N consecutive ports (stretch goal) will be implemented in another PR.

Is there a JIRA tracking this effort?

core/src/main/scala/com/mesosphere/usi/core/matching/PortResource.scala
jeschkies: Please add docs.

What is the reason not to have this in `core-models`?

core/src/main/scala/com/mesosphere/usi/core/matching/PortResource.scala
```diff
@@ -0,0 +1,223 @@
...
+import scala.annotation.tailrec
+import scala.collection.JavaConverters._
+import scala.util.Random
+
+case class PortResource(ports: Seq[Int], random: Random = Random) extends ResourceRequirement {
+  override def description: String = s"$resourceType:[${ports.mkString(",")}]"
+
+  override def resourceType: ResourceType = ResourceType.PORTS
+
+  @tailrec private def findResourceToMatch(
```
jeschkies: Could you add docs. I have a hard time understand what this method does.

alenkacz:
> > Take N consecutive ports (stretch goal) will be implemented in another PR.
> 
> Is there a JIRA tracking this effort?

Yes, the same JIRA tracking this effort. I just separated it into multiple PRs...

core/src/main/scala/com/mesosphere/usi/core/matching/PortResource.scala
```diff
@@ -0,0 +1,223 @@
...
+          case consumedResources =>
+            Some(
+              ResourceMatchResult(
+                consumedResources,
+                unmatchedResources ++ rest ++ ResourceUtil.consumeResources(Seq(next), consumedResources)))
+        }
+    }
+  }
+
+  private def matchPortResource(resource: Protos.Resource): Seq[Protos.Resource] = {
```
jeschkies: It seems we match against `ports`. What do you think of passing them in and make this a pure method?

core/src/main/scala/com/mesosphere/usi/core/matching/PortResource.scala
```diff
@@ -0,0 +1,223 @@
...
+        None
+      case next :: rest =>
+        matchPortResource(next) match {
+          case Nil =>
+            findResourceToMatch(next :: unmatchedResources, rest)
+          case consumedResources =>
+            Some(
+              ResourceMatchResult(
+                consumedResources,
+                unmatchedResources ++ rest ++ ResourceUtil.consumeResources(Seq(next), consumedResources)))
```
jeschkies: Where is `consumerResources` defined?

jeschkies (changes requested):
Thanks for the port. I I'm not mistaken we would pass in the use ports. Where is that?

Also, what do you think of using a simple loop for `findResourceToMatch` with a mutable state for `unmatchedResources` and `remainingResources`. This should be faster. Maybe we should add the benchmarks as well so that one could verify a follow up refactoring.

alenkacz:
RE: https://github.com/mesosphere/usi/pull/56#discussion_r262462279

ScalarResource is not in models as well. Yeah I was asking the same question, for me models kind of fall apart here because these classes contain so much business logic that having them in models package just feels super weird... But I guess we should move them since we have models :-I

Let's not do it in this PR if we decide to do it though, since more classes would have to be moved.

core/src/main/scala/com/mesosphere/usi/core/matching/PortResource.scala
alenkacz: Renamed and added docs...

core/src/main/scala/com/mesosphere/usi/core/matching/PortResource.scala
alenkacz: in ResourceUtil. That class was already ported over with Tests by Tim in previous PR

core/src/main/scala/com/mesosphere/usi/core/matching/PortResource.scala
alenkacz: @meln1k what would you propose? These classes were named e.g. `ScalarResourceRequirment` - to me that was too long :)

core/src/main/scala/com/mesosphere/usi/core/matching/RangeResource.scala
```diff
@@ -0,0 +1,251 @@
...
+
+  /**
+    * Iterates over given resources keeping in track unmatched resources.
+    * In the end produces ResourceMatchResult if we were able to find a match on the given resources.
+    *
+    * @param unmatchedResources already processed resources that did not contain resources we are trying to match
+    * @param remainingResources resources we still need to process when looking for resource match
+    * @return None if no match was possible, ResourceMatchResult if we were able to consume requested resources
+    */
+  @tailrec private def matchAndConsumeIter(
```
zen-dog, line 40: super nit: the same method in `ScalarResource` is called just `iter` ;)

alenkacz: yeah I know ... I like it more like this :slightly_smiling_face: I think it's easier to understand...

core/src/main/scala/com/mesosphere/usi/core/matching/RangeResource.scala
```diff
@@ -0,0 +1,251 @@
...
+    * @return final list of mesos resources created after consuming requested values, empty if match was not possible
+    */
+  private def tryConsumeValuesFromResource(values: Seq[Int], resource: Protos.Resource): Seq[Protos.Resource] = {
+    val offeredRanges = parseResourceToRanges(resource)
+    if (offeredRanges.isEmpty || values.isEmpty) {
+      return Seq.empty
+    }
+
+    // non-dynamic values
+    val staticRequestedValues = values.collect { case v if v != 0 => v }.toSet
```
zen-dog: maybe just `values.filter(_ != 0)`?

alenkacz: haha yeah, that's probably the Scala you were talking about yesterday :)

zen-dog (approved):
🚢 !

core/src/main/scala/com/mesosphere/usi/core/matching/RangeResource.scala
```diff
@@ -0,0 +1,251 @@
...
+  * Represents requirement for Mesos resource of type Range.
+  * You can either request static values (by providing concrete value) or dynamic values (by providing 0).
+  * Dynamic values are then selected from the offered ranges. If random implementation is passed in, these ranges are randomized.
+  * For more information about randomization see [[lazyRandomValuesFromRanges()]]
+  *
+  * @param requestedValues values pod wants to consume on the given ranges
+  * @param resourceType name of resource (e.g. ports)
+  * @param random when requesting dynamic values, you can provide Random implementation if you want dynamic values to be randomized
+  */
+case class RangeResource(requestedValues: Seq[Int], resourceType: ResourceType, random: Random = Random)
```
jeschkies: I don't quite get the naming scheme. Would I require the whole range? Isn't a range something like `[0, 3]`, `[0, 4)` and so on? Also, what is the benefit of passing a type? What do you think about
```
trait RangeResource {
  val resourceType: ResourceType
  val requestRange
  ...
}

case class Ports extends RangeResource {
  override val resourceType: ResourceType = ResourceType.PORTS
}
```

core/src/main/scala/com/mesosphere/usi/core/matching/RangeResource.scala
```diff
@@ -0,0 +1,251 @@
...
+        tryConsumeValuesFromResource(requestedValues, next) match {
+          case Nil =>
+            matchAndConsumeIter(next :: unmatchedResources, rest)
+          case consumedResources =>
+            Some(
+              ResourceMatchResult(
+                consumedResources,
+                unmatchedResources ++ rest ++ ResourceUtil.consumeResources(Seq(next), consumedResources)))
+        }
+    }
```
jeschkies, line 57: Why not make it a mutable loop? I still think that would speed up things quite a bit.

core/src/main/scala/com/mesosphere/usi/core/matching/RangeResource.scala
```diff
@@ -0,0 +1,251 @@
...
+  }
+
+  /**
+    * Tries to match and consume values from a given resource
+    * @param values values we want to consume
```
jeschkies: `requestedResource` or `requestedValues` is a better fit IMHO.

core/src/main/scala/com/mesosphere/usi/core/matching/RangeResource.scala
```diff
@@ -0,0 +1,251 @@
...
+    * @return final list of mesos resources created after consuming requested values, empty if match was not possible
+    */
+  private def tryConsumeValuesFromResource(values: Seq[Int], resource: Protos.Resource): Seq[Protos.Resource] = {
+    val offeredRanges = parseResourceToRanges(resource)
+    if (offeredRanges.isEmpty || values.isEmpty) {
+      return Seq.empty
+    }
+
+    // non-dynamic values
+    val staticRequestedValues = values.filter(v => v != 0).toSet
```
jeschkies: I don't know if `0` is there proper way to do this. If I'm not mistaken `0` was used because we used it for ports. However, in another context `0` might actually be a concrete value I want to match. What do you think about typing the requested values?

The set can be computed once btw since `values` won't change.

core/src/main/scala/com/mesosphere/usi/core/matching/RangeResource.scala
```diff
@@ -0,0 +1,251 @@
...
+  }
+}
+
+case class MesosRange(minValue: Int, maxValue: Int) {
+  private[this] def range: Range.Inclusive = Range.inclusive(minValue, maxValue)
+  def size: Int = range.size
+
+  def iterator: Iterator[Int] = range.iterator.map(v => v)
+  def drop(n: Int): Iterator[Int] = range.drop(n).iterator.map(v => v)
+  def take(n: Int): Iterator[Int] = range.take(n).iterator.map(v => v)
```
jeschkies: I curious why we have a map here. What is the advantage?

core/src/main/scala/com/mesosphere/usi/core/matching/RangeResource.scala
```diff
@@ -0,0 +1,251 @@
...
+  private def tryConsumeValuesFromResource(values: Seq[Int], resource: Protos.Resource): Seq[Protos.Resource] = {
+    val offeredRanges = parseResourceToRanges(resource)
+    if (offeredRanges.isEmpty || values.isEmpty) {
+      return Seq.empty
+    }
+
+    // non-dynamic values
+    val staticRequestedValues = values.filter(v => v != 0).toSet
+    val availableForDynamicAssignment: Iterator[Int] =
+      lazyRandomValuesFromRanges(offeredRanges, random).filter(v => !staticRequestedValues(v))
```
jeschkies, line 77: If I remember correctly this bothered me back when I refactored [AssignDynamicServiceLogic](https://github.com/mesosphere/marathon/blob/master/src/main/scala/mesosphere/marathon/core/group/impl/AssignDynamicServiceLogic.scala#L9). It seems we could reuse the filtered values but then `lazyRandomValuesFromRanges` is recomputed each time.

core/src/main/scala/com/mesosphere/usi/core/matching/RangeResource.scala
```diff
@@ -0,0 +1,251 @@
...
+    * We want to use consecutive ports to avoid excessive range fragmentation.
+    *
+    * The implementation idea:
+    *
+    * * Randomize the order of the offered ranges.
+    * * Now treat the ports contained in the ranges as one long sequence of ports.
+    * * We randomly choose an index where we want to start assigning dynamic ports in that sequence. When
+    *   we hit the last offered port with wrap around and start offering the ports at the beginning
+    *   of the sequence up to (excluding) the port index we started at.
+    * * The next range is determined on demand. That's why an iterator is returned.
```
jeschkies, line 132: I don't quite get this. Can the ranges be overlapping? If not why not shuffle all ranges in one?

core/src/main/scala/com/mesosphere/usi/core/matching/RangeResource.scala
```diff
@@ -0,0 +1,251 @@
...
+      resourceType: ResourceType): Seq[Protos.Resource] = {
+    /*
+     * Create as few ranges as possible from the given values while preserving the order of the values.
+     *
+     * It does not check if the given values have different roles.
+     */
+    def createRanges(ranges: Seq[Int]): Seq[Range] = {
+      val builder = Seq.newBuilder[Range]
+
+      @tailrec
```
jeschkies: The builder pattern and tail recursion don't mix well IMHO. This should be a simple iteration or fold.

```scala
val builder = Vector.newBuilder[Range]

if (ranges.nonEmpty) {
  var currentRange = ranges.head
  ranges.tail.foreach { v =>
    if (overlaps(currentRange, v)) {
     currentRange = Range(currentRange.first, v)
    } else {
      builder += currentRange
      currentRange = Range(v, v)
    }
  }
  builder += currentRange // append last range.
}
```

```scala
ranges.foldLeft(Vector.empty[Range]) { (acc, next) =>
  // Merge next range and last if they overlap
  if (acc.headOption.exists(curr => overlaps(curr, next))) {
    Range(acc.head.first, next) +: acc.tail
  } else {
    Range(next, next) +: acc
  }
}
```

Anyhow, are the ranges always sorted?

jeschkies (commented):
I don't know. I probably should just withdraw from the review. James made a great effort but it seems we have some quirks and early optimization in it.

core/src/main/scala/com/mesosphere/usi/core/matching/RangeResource.scala
alenkacz: Range resource is type of resource defined http://mesos.apache.org/documentation/attributes-resources/ ... this is way to express you want part of that resource. Similarly as you do it with `ScalarResource` e.g. for CPUs.

There might be similar strategies how you want to take part of that range. Right now I just implemented list of values. That's the list of values you want to consume out of that range.

You need the type because Resource can represent various types (names) of resources. As scalar can be of type mem or cpus, Range is also generic and can be of any type. In the `RangeResource` object there is helper method for ports that will create the resource for you without specifying the type.

alenkacz: maybe it should be actually named `RangeRequirment` instead of `RangeResource`? WDYT?

core/src/main/scala/com/mesosphere/usi/core/matching/RangeResource.scala
alenkacz: 🤷‍♀️ I actually re-used this from `ScalarResource` where Tim implemented it and I trust his judgment. If we change this, we should do it on both places.

core/src/main/scala/com/mesosphere/usi/core/matching/RangeResource.scala
alenkacz: yeah I agree, I actually though about it. Makes sense, will do! 👍

core/src/main/scala/com/mesosphere/usi/core/matching/RangeResource.scala
alenkacz: Reuse for what BTW? For multiple pods? I don't see any other reuse here apart from that... And that's not trivial to do right now... Since the current logic does evaluation pod after pod...

Are you fine with doing such optimizations as a follow up? Could you maybe even create JIRA. I think it would be nice to have some benchmark here as well...

core/src/main/scala/com/mesosphere/usi/core/matching/RangeResource.scala
alenkacz: hmm to be honest, I don't know if they can be overlapping. I would assume not? This seems like a pretty brittle part and I would submit:
- leave it as is and create JIRA to write benchmark and optimize it
- drop the random shuffling and don't support random right now. Do it in a separate JIRA with the benchmark.

I am fine with both. I understand the code is not the best piece of scala we have here but at the same time it's battle tested :D so I just pulled it over.

core/src/main/scala/com/mesosphere/usi/core/matching/RangeResource.scala
alenkacz: how would you do drop and take on a map? that does not sound like an optimal data structure to me

core/src/main/scala/com/mesosphere/usi/core/matching/RangeResource.scala
alenkacz: "while preserving the order of the values."

alenkacz: and agreed, I chose the iterative way an re-implemented it. Thanks!

core/src/main/scala/com/mesosphere/usi/core/matching/RangeResource.scala
jeschkies: Just `drop`? I must be missing something but why do we call `map(v => v)`? What is that part doing?

core/src/main/scala/com/mesosphere/usi/core/matching/RangeResource.scala
```diff
@@ -0,0 +1,260 @@
...
+import com.mesosphere.usi.core.models.{ResourceMatchResult, ResourceRequirement, ResourceType}
+import com.mesosphere.usi.core.protos.ProtoBuilders
+import org.apache.mesos.v1.Protos
+
+import scala.annotation.tailrec
+import scala.collection.JavaConverters._
+import scala.util.Random
+
+/**
+  * Represents requirement for Mesos resource of type Range.
```
jeschkies: Could you link the Apache docs you've mentioned?

core/src/main/scala/com/mesosphere/usi/core/matching/RangeResource.scala
jeschkies: We should just add benchmarks and follow up. I'm happy to do that.

core/src/main/scala/com/mesosphere/usi/core/matching/RangeResource.scala
jeschkies: `tryConsumeValuesFromResource` is called for each `remainingResources`. We probably should move on but I still don't quite get what's happening here.

core/src/main/scala/com/mesosphere/usi/core/matching/RangeResource.scala
jeschkies: I think we can drop `.map(v => v)`.

jeschkies (approved):
Thanks. I have only one nit and the rest could be a follow up.

core/src/main/scala/com/mesosphere/usi/core/matching/RangeResource.scala
alenkacz: lol :) good catch

core/src/main/scala/com/mesosphere/usi/core/matching/RangeResource.scala
alenkacz: yes, but for every remaining resource you need to compute new offered ports because there's different range... I don't see any easy optimization here but maybe I am just missing something. I am open to discussing this further...

PR #57, opened 2019-03-05
Do not ACK statuses without UUID

alenkacz:


zen-dog (approved):

core/src/main/scala/com/mesosphere/usi/core/logic/MesosEventsLogic.scala
```diff
@@ -156,8 +156,12 @@ private[core] class MesosEventsLogic(mesosCallFactory: MesosCalls) extends Impli
 
           SchedulerEvents(
             stateEvents = List(PodStatusUpdated(podId, Some(newStatus))),
-            mesosCalls =
+            mesosCalls = if (taskStatus.hasUuid) {
+              // frameworks should accept only status updates that have UUID set
               List(mesosCallFactory.newAcknowledge(taskStatus.getAgentId, taskStatus.getTaskId, taskStatus.getUuid))
+            } else {
+              List.empty
```
meln1k: Just a nit, but what do you think about `Nil` here? :)

core/src/main/scala/com/mesosphere/usi/core/protos/ProtoBuilders.scala
```diff
@@ -117,6 +118,7 @@ private[usi] object ProtoBuilders {
   def newTaskStatus(
       taskId: Mesos.TaskID,
       state: Mesos.TaskState,
+      uuid: ByteString, // UUID should be present, only e.g. TASK_ERROR does not have UUID
```
meln1k, line 121: Hm what do you think about taking a java.util.UUID class here and doing the serialisation inside the method? 

I'm asking because there are several ways to serialize UUID into bytes, such as `uuid.getBytes`, `uuid.toString.getBytes`, having a ByteString with 2 longs and so on. I'd prefer to do that in this method.

alenkacz: well we are actually not the people creating these. UUID comes from mesos and we just put it into the ACK message. So I think we should ideally not touch it and keep it as close to pure bytes as possible.

zen-dog: Keep in mind: this is the *Protobuf ByteString* holding the UUID and this method can only be used for mock testing (since a framework doesn't generate task statuses)

meln1k: Alright, no questions then :)

meln1k (approved):

PR #58, opened 2019-03-05
Support only MULTI_ROLE frameworks

alenkacz:
MULTI_ROLE check is in MesosClient as well as USI core because one is not tight to our MesosClient implementation and there can be alternative ones. At the same time, USI right now won't work on properly for single role framework because for those `AllocationInfo` is optional.

JIRA: DCOS-47475


core/src/main/scala/com/mesosphere/usi/core/Scheduler.scala
```diff
@@ -50,9 +52,16 @@ object Scheduler {
   type StateOutput = (StateSnapshot, Source[StateEvent, Any])
 
   def fromClient(client: MesosClient): Flow[SpecInput, StateOutput, NotUsed] = {
+    if (!isMultiRoleFramework(client.frameworkInfo)) {
+      throw new IllegalArgumentException(
+        "Mesos client supports only MULTI_ROLE frameworks. Please provide create MesosClient with FrameworkInfo that has capability MULTI_ROLE")
```
zen-dog: It's rather: `USI scheduler provides support for MULTI_ROLE frameworks only....`

mesos-client/src/main/scala/com/mesosphere/mesos/client/MesosClient.scala
```diff
@@ -294,6 +300,11 @@ object MesosClient extends StrictLogging with StrictLoggingFlow {
       system: ActorSystem,
       materializer: ActorMaterializer): Source[MesosClient, NotUsed] = {
 
+    if (!isMultiRoleFramework(frameworkInfo)) {
```
zen-dog: I would not put this check into the client. It's the "limitation" of our core-scheduler.

core/src/main/scala/com/mesosphere/usi/core/Scheduler.scala
```diff
@@ -50,9 +52,16 @@ object Scheduler {
...
+    }
     fromFlow(client.calls, Flow.fromSinkAndSource(client.mesosSink, client.mesosSource))
   }
 
+  private def isMultiRoleFramework(frameworkInfo: FrameworkInfo): Boolean =
+    frameworkInfo.getCapabilitiesList.asScala.exists(c => c.getType == FrameworkInfo.Capability.Type.MULTI_ROLE)
```
jeschkies:
```suggestion
    frameworkInfo.getCapabilitiesList.asScala.exists(_.getType == FrameworkInfo.Capability.Type.MULTI_ROLE)
```

mesos-client/src/main/scala/com/mesosphere/mesos/client/MesosClient.scala
```diff
@@ -25,6 +26,11 @@ trait MesosClient {
     */
   def frameworkId: FrameworkID
 
+  /**
+    * The frameworkInfo as which this client is currently subscribed with.
```
jeschkies:
```suggestion
    * The framework info this client is currently subscribed with.
```

jeschkies (commented):
Do we have a JIRA?

mesos-client/src/main/scala/com/mesosphere/mesos/client/MesosClient.scala
alenkacz: okay, makes sense!

zen-dog (approved):

jeschkies (approved):

PR #61, opened 2019-03-07
Scheduler Adapter interface review

meln1k:
This is a draft of a scheduler adapter. The main purpose of the adapter is to provide a simple interface for writing frameworks using USI.

jira: DCOS-50268


core/src/main/scala/com/mesosphere/usi/core/SchedulerAdapter.scala
```diff
@@ -0,0 +1,199 @@
...
+
+  /**
+    * Represents the scheduler as a Sink and Source.
+    *
+    * This method will materialize the scheduler first, then Sink and Source can be materialized independently.
+    * Please note that the first spec event coming to the Sink must be a SpecsSnapshot otherwise the stream will fail.
+    *
+    * @return Source which produces StateEvents and Sink which accepts SpecEvents
+    */
+  def asSourceAndSink(): (Source[StateEvent, NotUsed], Sink[SpecEvent, NotUsed]) = {
```
jeschkies: What do you think about passing the `SpecsSnapshot`? Or rather why is it not passed here?

core/src/main/scala/com/mesosphere/usi/core/SchedulerAdapter.scala
```diff
@@ -0,0 +1,199 @@
...
+
+  /**
+    * Represents the scheduler as a Sink and Source.
+    *
+    * This method will materialize the scheduler first, then Sink and Source can be materialized independently.
+    *
+    * @param specsSnapshot Snapshot of the current specs
+    * @return Snapshot of the current state, as well as Source which produces StateEvents and Sink which accepts SpecEvents
+    */
+  def asSourceAndSink(specsSnapshot: SpecsSnapshot): (Future[StateSnapshot], Source[StateEvent, Any], Sink[SpecUpdated, Any]) = {
```
zen-dog:
```suggestion
  def asSourceAndSink(specsSnapshot: SpecsSnapshot = SpecsSnapshot.empty): (Future[StateSnapshot], Source[StateEvent, Any], Sink[SpecUpdated, Any]) = {
```
Give it a default parameter and I don't believe we need the same method above

core/src/main/scala/com/mesosphere/usi/core/SchedulerAdapter.scala
meln1k: Check the method below :)

zen-dog: See my comment below ;)

core/src/main/scala/com/mesosphere/usi/core/SchedulerAdapter.scala
```diff
@@ -0,0 +1,199 @@
...
+
+  /**
+    *
+    * Represents the scheduler as a Flow.
+    *
+    * This method will materialize the scheduler first, then Flow can be materialized independently.
+    *
+    * @return
+    */
+  def asFlow(): Flow[SpecEvent, StateEvent, NotUsed] = {
```
zen-dog: imho same goes for this method

zen-dog (commented):
Would you please add a java `examples/java-long-running-task` example, using the java adapter to demonstrate how it works? Otherwise, this is great work! 👏

core/src/main/scala/com/mesosphere/usi/core/javadsl/SchedulerAdapter.scala
```diff
@@ -0,0 +1,84 @@
...
+    *
+    * This method will materialize the scheduler first, then Sink and Source can be materialized independently.
+    *
+    * @param specsSnapshot Snapshot of the current specs
+    * @return Snapshot of the current state, as well as Source which produces StateEvents and Sink which accepts SpecEvents
+    */
+  def asSourceAndSink(specsSnapshot: SpecsSnapshot = SpecsSnapshot.empty)
+    : Tuple3[CompletableFuture[StateSnapshot], Source[StateEvent, Any], Sink[SpecUpdated, Any]] = {
+    val (snap, source, sink) = delegate.asSourceAndSink(specsSnapshot)
+    Tuple3(snap.toJava.toCompletableFuture, source.asJava, sink.asJava)
```
jeschkies: I wonder if the tuple should be a future since one should start pushing `SpecUpdated` and pulling `StateEvent` *after* the `StateSnapshot` arrived.

core/src/main/scala/com/mesosphere/usi/core/javadsl/SchedulerAdapter.scala
```diff
@@ -0,0 +1,84 @@
...
+  }
+
+  /**
+    *
+    * Represents the scheduler as a Flow.
+    *
+    *
```
jeschkies:
```suggestion
```

core/src/main/scala/com/mesosphere/usi/core/javadsl/SchedulerAdapter.scala
```diff
@@ -0,0 +1,84 @@
...
+    * @return Snapshot of the current state, as well as Source which produces StateEvents and Sink which accepts SpecEvents
+    */
+  def asSourceAndSink(specsSnapshot: SpecsSnapshot = SpecsSnapshot.empty)
+    : Tuple3[CompletableFuture[StateSnapshot], Source[StateEvent, Any], Sink[SpecUpdated, Any]] = {
+    val (snap, source, sink) = delegate.asSourceAndSink(specsSnapshot)
+    Tuple3(snap.toJava.toCompletableFuture, source.asJava, sink.asJava)
+  }
+
+  /**
+    *
```
jeschkies:
```suggestion
```

core/src/main/scala/com/mesosphere/usi/core/javadsl/SchedulerAdapter.scala
```diff
@@ -0,0 +1,84 @@
...
+    */
+  def asFlow(specsSnapshot: SpecsSnapshot = SpecsSnapshot.empty)
+    : Pair[CompletableFuture[StateSnapshot], Flow[SpecUpdated, StateEvent, NotUsed]] = {
+    val (snap, flow) = delegate.asFlow(specsSnapshot)
+    Pair(snap.toJava.toCompletableFuture, flow.asJava)
+  }
+
+  /**
+    *
+    * Represents the scheduler as a SourceQueue and a SinkQueue.
```
jeschkies: What is the difference to `Sink` and `Source`? We probably should mention this in the docs.

jeschkies (commented):
I like the idead of `javadsl` and `scaladsl`. Could you add example code to each Scaladoc? This would be amazing!

alenkacz:
Thanks! I like the scala and java separation. Let's see the example usage so I can better see how will the framework code look like.

core/src/main/scala/com/mesosphere/usi/core/javadsl/SchedulerAdapter.scala
timcharper: Hmm, I'd recommend using `NotUsed` over `Any`.

core/src/main/scala/com/mesosphere/usi/core/SchedulerAdapter.scala
jeschkies: I'd much rather have it in the docs or are we lifting the hello world examples to docs?

core/src/main/scala/com/mesosphere/usi/core/javadsl/SchedulerAdapter.scala
meln1k: Well, I guess akka javadoc will tell it better than me :) Do you still want me to put it there?

kensipe (changes requested):
DSL == domain specific language.   I don't see how this qualifies.   Creating a programming language adaptor is extremely valuable.   I don't like the package naming that doesn't align for me.

there are repeated uses of single char variables which are terrible for readability.   Can we please fix that?

core/src/main/scala/com/mesosphere/usi/core/javadsl/SchedulerAdapter.scala
kensipe: I agree with @jeschkies  We should do 2 things here.   
1. provide a link to Akka docs for more details on Sink vs Source
2. more important... help bridge the Sink vs Source of this specific domain for the reader of this code.   An example that is context specific to USI would be valuable IMO

core/src/main/scala/com/mesosphere/usi/core/scaladsl/SchedulerAdapter.scala
```diff
@@ -0,0 +1,167 @@
...
+        killSwitch.shutdown()
+      case Failure(cause) =>
+        killSwitch.abort(cause)
+    }
+
+    Source.maybe.prepend {
+      val events = Source
+        .unfoldResourceAsync[SpecUpdated, SinkQueueWithCancel[SpecUpdated]](
+          create = () => Future.successful(specQueue),
+          read = q => q.pull(),
```
kensipe: please rename `q`

core/src/main/scala/com/mesosphere/usi/core/scaladsl/SchedulerAdapter.scala
```diff
@@ -0,0 +1,167 @@
...
+    val sourceWithKillSwitch = stateSource
+      .watchTermination() {
+        case (m, cancellationSignal) =>
+          cancellationSignal.onComplete {
+            case Success(_) =>
+              killSwitch.shutdown()
+            case Failure(cause) =>
+              killSwitch.abort(cause)
+          }
+          m
```
kensipe: please rename `m`

core/src/main/scala/com/mesosphere/usi/core/scaladsl/SchedulerAdapter.scala
```diff
@@ -0,0 +1,167 @@
...
+    val sinkWithKillSwitch = Flow[SpecUpdated]
+      .watchTermination() {
+        case (m, cancellationSignal) =>
+          cancellationSignal.onComplete {
+            case Success(_) =>
+              killSwitch.shutdown()
+            case Failure(cause) =>
+              killSwitch.abort(cause)
+          }
+          m
```
kensipe: ditto

core/src/test/scala/com/mesosphere/usi/core/scaladsl/SchedulerAdapterTest.scala
```diff
@@ -0,0 +1,307 @@
...
+      "Complete both sink and source when the event stream coming to scheduler is cancelled by the scheduler" in new SchedulerFixture {
+        val (_, source, sink) = adapter.asSourceAndSink()
+        val (_, innerSource) = specInputQueue.pull().futureValue.get
+
+        val stateStreamCompletionPromise = Promise[String]()
+        val specStreamCompletionPromise = Promise[String]()
+
+        Source
+          .repeat(specEvent)
+          .watchTermination()((_, f) => f.map(_ => specStreamCompletionPromise.trySuccess("completed")))
```
kensipe: please rename `f`

core/src/test/scala/com/mesosphere/usi/core/scaladsl/SchedulerAdapterTest.scala
```diff
@@ -0,0 +1,307 @@
...
+                specStreamCompletionPromise.trySuccess("stopped")
+
+              case Failure(ex) =>
+                specStreamCompletionPromise.trySuccess("failed")
+          })
+          .runWith(sink)
+
+        source.map { e =>
+          println(e)
+        }.watchTermination()((_, f) =>
```
kensipe: all the `f`s :)

kensipe (dismissed):

core/src/main/scala/com/mesosphere/usi/core/javaapi/SchedulerAdapter.scala
```diff
@@ -0,0 +1,87 @@
...
+
+  /**
+    * Represents the scheduler as a Sink and Source.
+    *
+    * This method will materialize the scheduler first, then Sink and Source can be materialized independently.
+    *
+    * @param specsSnapshot Snapshot of the current specs
+    * @return Snapshot of the current state, as well as Source which produces StateEvents and Sink which accepts SpecEvents
+    */
+  def asSourceAndSink(specsSnapshot: SpecsSnapshot = SpecsSnapshot.empty)
```
jeschkies: Defaults don't work for Java, you'll have to use method overloading. At least I think so.

core/src/main/scala/com/mesosphere/usi/core/scalaapi/SchedulerAdapter.scala
```diff
@@ -0,0 +1,170 @@
...
+              case Failure(cause) =>
+                killSwitch.abort(cause)
+            }
+        }
+
+      Source.single(specsSnapshot -> events)
+    }.via(killSwitch.flow)
+      .via(schedulerFlow)
+      .flatMapConcat {
+        case (snapshot, updates) =>
```
jeschkies: There are more than twenty lines until we reached this. @timcharper, do we really think that our current USI interface is a good fit? As a consumer I don't really mind but I really wonder who should maintain these interfaces.

core/src/test/scala/com/mesosphere/usi/core/scalaapi/SchedulerAdapterTest.scala
```diff
@@ -0,0 +1,315 @@
...
+import com.mesosphere.usi.core.models._
+import com.mesosphere.utils.AkkaUnitTest
+import org.scalatest._
+
+import scala.concurrent.Promise
+import scala.util.{Failure, Success}
+
+class SchedulerAdapterTest extends AkkaUnitTest with Inside {
+
+  def createMockedScheduler: (
```
jeschkies: Don't we have such a mock in `test-utils`, @alenkacz and @zen-dog?

core/src/test/scala/com/mesosphere/usi/core/scalaapi/SchedulerAdapterTest.scala
```diff
@@ -0,0 +1,315 @@
...
+  "SchedulerAdapter" when {
+    "materialized as Source and Sink" should {
+      "Propagate a snapshot to the underlying scheduler" in new SchedulerFixture {
+        adapter.asSourceAndSink(SpecsSnapshot(List(podSpec), Nil))
+        specInputQueue.pull().futureValue.get._1.podSpecs.head shouldEqual podSpec
+      }
+
+      "If snapshot is not provided push an empty snapshot" in new SchedulerFixture {
+        adapter.asSourceAndSink()
+        specInputQueue.pull().futureValue.get._1 shouldEqual SpecsSnapshot.empty
```
jeschkies:
```suggestion
        specInputQueue.pull().futureValue.get._1 should be('empty)
```

We may have to add `isEmpty` to `SpecsSnapshot`.

Also, shouldn't this be a state snapshot?

core/src/test/scala/com/mesosphere/usi/core/scalaapi/SchedulerAdapterTest.scala
```diff
@@ -0,0 +1,315 @@
...
+
+        snapshotF.futureValue shouldEqual stateSnapshot
+      }
+
+      "Push the SpecUpdatedEvents downstream" in new SchedulerFixture {
+        val (_, _, sink) = adapter.asSourceAndSink()
+
+        Source.repeat(specEvent).runWith(sink)
+
+        specInputQueue.pull().futureValue.get._2.runWith(Sink.head).futureValue shouldEqual specEvent
```
jeschkies: Ah, could you use the following syntax

```scala
"Push the SpecUpdatedEvents downstream" in {
  val f = new SchedulerFixture

  f.specInputQueue.pull().futureValue ...
}
```

it was not obvious to me that you are using a fixture here.

core/src/test/scala/com/mesosphere/usi/core/scalaapi/SchedulerAdapterTest.scala
```diff
@@ -0,0 +1,315 @@
...
+
+    val (scheduler, specInputQueue, stateOutputQueue) = createMockedScheduler
+    val adapter = new SchedulerAdapter(scheduler)
+  }
+
+  "SchedulerAdapter" when {
+    "materialized as Source and Sink" should {
+      "Propagate a snapshot to the underlying scheduler" in new SchedulerFixture {
+        adapter.asSourceAndSink(SpecsSnapshot(List(podSpec), Nil))
+        specInputQueue.pull().futureValue.get._1.podSpecs.head shouldEqual podSpec
```
jeschkies:
```suggestion
        specInputQueue.pull().futureValue.value._1.podSpecs.head shouldEqual podSpec
```

core/src/test/scala/com/mesosphere/usi/core/scalaapi/SchedulerAdapterTest.scala
```diff
@@ -0,0 +1,315 @@
...
+
+        innerSource.runWith(Sink.cancelled) // cancel incoming updates on scheduler size
+
+        specStreamCompletionPromise.future.futureValue shouldEqual "completed"
+        stateStreamCompletionPromise.future.futureValue shouldEqual "completed"
+      }
+
+      "Complete both sink and source when the event stream coming to scheduler is cancelled by the client" in new SchedulerFixture {
+        val (_, source, sink) = adapter.asSourceAndSink()
+        val (_, innerSource) = specInputQueue.pull().futureValue.get
```
jeschkies:
```suggestion
        val (_, innerSource) = specInputQueue.pull().futureValue.value
```

core/src/test/scala/com/mesosphere/usi/core/scalaapi/SchedulerAdapterTest.scala
```diff
@@ -0,0 +1,315 @@
...
+            terminationSignal.map(_ => stateStreamCompletionPromise.trySuccess("completed")))
+          .runWith(Sink.ignore)
+
+        specStreamCompletionPromise.future.futureValue shouldEqual "completed"
+        stateStreamCompletionPromise.future.futureValue shouldEqual "completed"
+      }
+
+      "Complete both sink and source when the event stream outgoing from scheduler is cancelled by the scheduler" in new SchedulerFixture {
+        val (_, source, sink) = adapter.asSourceAndSink()
+        val (_, innerSource) = specInputQueue.pull().futureValue.get
```
jeschkies:
```suggestion
        val (_, innerSource) = specInputQueue.pull().futureValue.value
```

core/src/test/scala/com/mesosphere/usi/core/scalaapi/SchedulerAdapterTest.scala
```diff
@@ -0,0 +1,315 @@
...
+
+        stateOutputQueue.offer(StateSnapshot.empty -> Source.single(stateEvent)).futureValue
+
+        specStreamCompletionPromise.future.futureValue shouldEqual "completed"
+        stateStreamCompletionPromise.future.futureValue shouldEqual "completed"
+      }
+
+      "Complete both sink and source when the event stream outgoing from scheduler is cancelled by the client" in new SchedulerFixture {
+        val (_, source, sink) = adapter.asSourceAndSink()
+        val (_, innerSource) = specInputQueue.pull().futureValue.get
```
jeschkies:
```suggestion
        val (_, innerSource) = specInputQueue.pull().futureValue.value
```

core/src/test/scala/com/mesosphere/usi/core/scalaapi/SchedulerAdapterTest.scala
```diff
@@ -0,0 +1,315 @@
...
+                stateStreamCompletionPromise.trySuccess("failed")
+          })
+          .runWith(Sink.ignore)
+
+        stateStreamCompletionPromise.future.futureValue shouldEqual "failed"
+      }
+
+      "Stop the sink when the failure occurred in the downstream" in new SchedulerFixture {
+        val (_, source, sink) = adapter.asSourceAndSink()
+        val (_, innerSource) = specInputQueue.pull().futureValue.get
```
jeschkies:
```suggestion
        val (_, innerSource) = specInputQueue.pull().futureValue.value
```

core/src/test/scala/com/mesosphere/usi/core/scalaapi/SchedulerAdapterTest.scala
```diff
@@ -0,0 +1,315 @@
...
+          .map(_ => throw new Exception("Boom"))
+          .to(Sink.ignore)
+          .run()
+
+        stateStreamCompletionPromise.future.futureValue shouldEqual "stopped"
+      }
+
+      "Stop both sink and source when the failure occurred in the scheduler input" in new SchedulerFixture {
+        val (_, source, sink) = adapter.asSourceAndSink()
+        val (_, innerSource) = specInputQueue.pull().futureValue.get
```
jeschkies:
```suggestion
        val (_, innerSource) = specInputQueue.pull().futureValue.value
```

core/src/test/scala/com/mesosphere/usi/core/scalaapi/SchedulerAdapterTest.scala
```diff
@@ -0,0 +1,315 @@
...
+          })
+          .runWith(Sink.ignore)
+
+        specStreamCompletionPromise.future.futureValue shouldEqual "stopped"
+        stateStreamCompletionPromise.future.futureValue shouldEqual "stopped"
+      }
+
+      "Stop sink and fail source when the failure occurred in the scheduler output" in new SchedulerFixture {
+        val (_, source, sink) = adapter.asSourceAndSink()
+        val (_, innerSource) = specInputQueue.pull().futureValue.get
```
jeschkies:
```suggestion
        val (_, innerSource) = specInputQueue.pull().futureValue.value
```

jeschkies (commented):
Thanks. Are you planning to provide an example in the docs or a hello world? If you write a hello world, please link it from the docs.

core/src/main/scala/com/mesosphere/usi/core/scalaapi/SchedulerAdapter.scala
```diff
@@ -0,0 +1,170 @@
...
+            completionSignal.onComplete {
+              case Success(_) =>
+                killSwitch.shutdown()
+              case Failure(cause) =>
+                killSwitch.abort(cause)
+            }
+        }
+
+      Source.single(specsSnapshot -> events)
+    }.via(killSwitch.flow)
```
zen-dog: why the `killSwitch.flow` here?

core/src/main/scala/com/mesosphere/usi/core/scalaapi/SchedulerAdapter.scala
```diff
@@ -0,0 +1,170 @@
...
+
+    // We need to handle the case when source is cancelled or failed
+    stateQueue.watchCompletion().onComplete {
+      case Success(_) =>
+        killSwitch.shutdown()
+      case Failure(cause) =>
+        killSwitch.abort(cause)
+    }
+
+    Source.maybe.prepend {
```
zen-dog: So, basically, this is a `Source.fromQueue` method, right? 😊 Should we extract it and call it what it is. I mean, since we have to use this workaround.

core/src/main/scala/com/mesosphere/usi/core/scalaapi/SchedulerAdapter.scala
```diff
@@ -0,0 +1,170 @@
...
+            case (_, cancellationSignal) =>
+              cancellationSignal.onComplete {
+                case Success(_) =>
+                  killSwitch.shutdown()
+                case Failure(cause) =>
+                  killSwitch.abort(cause)
+              }
+          }
+      }
+      .via(killSwitch.flow)
```
zen-dog: similar to my comment above, I don't think this `.via(killSwitch.flow)` is necessary

core/src/main/scala/com/mesosphere/usi/core/javaapi/SchedulerAdapter.scala
alenkacz: shouldn't we write the java adapter in java? That would make sense to me... Also what we produce will be always nice from java :)

alenkacz: also on a different topic - do we need the default value here? I would assume that most of the time, snapshot will have to be provided. Empty snapshot make senses only for totally fresh orchestrator

core/src/main/scala/com/mesosphere/usi/core/scalaapi/SchedulerAdapter.scala
```diff
@@ -0,0 +1,170 @@
...
+        case (materializedValue, cancellationSignal) =>
+          cancellationSignal.onComplete {
+            case Success(_) =>
+              killSwitch.shutdown()
+            case Failure(cause) =>
+              killSwitch.abort(cause)
+          }
+          materializedValue
+      }
+      .via(killSwitch.flow)
```
zen-dog: is this `.via(killSwitch.flow)` necessary?

core/src/main/scala/com/mesosphere/usi/core/scalaapi/SchedulerAdapter.scala
```diff
@@ -0,0 +1,170 @@
...
```
zen-dog: and this?

core/src/main/scala/com/mesosphere/usi/core/scalaapi/SchedulerAdapter.scala
```diff
@@ -0,0 +1,170 @@
...
+    *
+    * This method will materialize the scheduler first, then Sink and Source can be materialized independently, but only once.
+    *
+    * @param specsSnapshot Snapshot of the current specs
+    * @return Snapshot of the current state, as well as Source which produces StateEvents and Sink which accepts SpecEvents
+    */
+  def asSourceAndSink(specsSnapshot: SpecsSnapshot = SpecsSnapshot.empty)
+    : (Future[StateSnapshot], Source[StateEvent, NotUsed], Sink[SpecUpdated, NotUsed]) = {
+
+    val (stateQueue, stateSource) = Source.queue[StateEvent](1, OverflowStrategy.backpressure).preMaterialize()
```
alenkacz: is it OK to hardcode the overflow strategy and buffer size here?

core/src/test/scala/com/mesosphere/usi/core/scalaapi/SchedulerAdapterTest.scala
```diff
@@ -0,0 +1,315 @@
...
+        source
+          .watchTermination()((_, terminationSignal) =>
+            terminationSignal.map(_ => stateStreamCompletionPromise.trySuccess("completed")))
+          .runWith(Sink.ignore)
+
+        specStreamCompletionPromise.future.futureValue shouldEqual "completed"
+        stateStreamCompletionPromise.future.futureValue shouldEqual "completed"
+      }
+
+      "Complete both sink and source when the event stream outgoing from scheduler is cancelled by the scheduler" in new SchedulerFixture {
```
zen-dog:
```suggestion
      "Complete both sink and source when the scheduler event stream is canceled by the scheduler itself" in new SchedulerFixture {
```

core/src/main/scala/com/mesosphere/usi/core/scalaapi/SchedulerAdapter.scala
```diff
@@ -0,0 +1,170 @@
...
+
+    val (specQueue, specSink) = Sink.queue[SpecUpdated]().preMaterialize()
+
+    val stateSnapshotPromise = Promise[StateSnapshot]()
+
+    val killSwitch = KillSwitches.shared("SchedulerAdapter.asSourceAndSink")
+
+    // We need to handle the case when source is cancelled or failed
```
zen-dog: Here and all other occurrences: 
```suggestion
    // We need to handle the case when the source is canceled or failed
```

core/src/test/scala/com/mesosphere/usi/core/scalaapi/SchedulerAdapterTest.scala
```diff
@@ -0,0 +1,315 @@
...
+
+      "Push the events from scheduler to the provided Source" in new SchedulerFixture {
+        val (_, source, _) = adapter.asSourceAndSink()
+
+        stateOutputQueue.offer(StateSnapshot.empty -> Source.repeat(stateEvent))
+
+        source.runWith(Sink.head).futureValue shouldEqual stateEvent
+      }
+
+      "Complete both sink and source when the event stream coming to scheduler is cancelled by the scheduler" in new SchedulerFixture {
```
zen-dog:
```suggestion
      "Complete both sink and source when the event stream coming to scheduler is canceled by the scheduler" in new SchedulerFixture {
```

core/src/test/scala/com/mesosphere/usi/core/scalaapi/SchedulerAdapterTest.scala
```diff
@@ -0,0 +1,315 @@
...
+            terminationSignal.map(_ => stateStreamCompletionPromise.trySuccess("completed")))
+          .runWith(Sink.ignore)
+
+        stateOutputQueue.offer(StateSnapshot.empty -> Source.single(stateEvent)).futureValue
+
+        specStreamCompletionPromise.future.futureValue shouldEqual "completed"
+        stateStreamCompletionPromise.future.futureValue shouldEqual "completed"
+      }
+
+      "Complete both sink and source when the event stream outgoing from scheduler is cancelled by the client" in new SchedulerFixture {
```
zen-dog:
```suggestion
      "Complete both sink and source when the event stream outgoing from scheduler is canceled by the client" in new SchedulerFixture {
```

zen-dog (commented):
Great tests 👏

core/src/main/scala/com/mesosphere/usi/core/scalaapi/SchedulerAdapter.scala
```diff
@@ -0,0 +1,170 @@
...
+
+  /**
+    * Represents the scheduler as a Sink and Source.
+    *
+    * This method will materialize the scheduler first, then Sink and Source can be materialized independently, but only once.
+    *
+    * @param specsSnapshot Snapshot of the current specs
+    * @return Snapshot of the current state, as well as Source which produces StateEvents and Sink which accepts SpecEvents
+    */
+  def asSourceAndSink(specsSnapshot: SpecsSnapshot = SpecsSnapshot.empty)
```
alenkacz: I must say I like this shape of the scheduler interface much more than the default one :) Future, Source and Sink are more straightforward to me :)

core/src/main/scala/com/mesosphere/usi/core/scalaapi/SchedulerAdapter.scala
```diff
@@ -0,0 +1,170 @@
...
+  }
+
+  /**
+    * Represents the scheduler as a Flow.
+    *
+    * This method will materialize the scheduler first, then Flow can be materialized independently.
+    * @param specsSnapshot Snapshot of the current specs
+    * @return
+    */
+  def asFlow(specsSnapshot: SpecsSnapshot = SpecsSnapshot.empty)
```
alenkacz: what would be the case where I would rather use this one? Is there one? If so, can we document it?

Otherwise I would rather have one recommended API than 10. Especially since one can create flow from the sink and source one so easily...

core/src/main/scala/com/mesosphere/usi/core/scalaapi/SchedulerAdapter.scala
```diff
@@ -0,0 +1,170 @@
...
+    * Represents the scheduler as a SourceQueue and a SinkQueue.
+    *
+    * @see https://doc.akka.io/api/akka/current/akka/stream/scaladsl/SourceQueue.html
+    *      https://doc.akka.io/api/akka/current/akka/stream/scaladsl/SinkQueue.html
+    *
+    * This method will materialize the scheduler first, then queues can be used independently.
+    * @param specsSnapshot Snapshot of the current specs
+    * @return
+    */
+  def asAkkaQueues(
```
alenkacz: could we also document what would be the cases when one should choose this one and not the other? I think it would be nice for people integrating with USI

core/src/test/scala/com/mesosphere/usi/core/scalaapi/SchedulerAdapterTest.scala
alenkacz: I think we have one for mesos, not for scheduler as a whole... but I might be mistaken :)

core/src/main/scala/com/mesosphere/usi/core/javadsl/SchedulerAdapter.scala
meln1k: I don't think it's a problem since all events that arrive before `StateSnapshot`will be backpressured.

core/src/main/scala/com/mesosphere/usi/core/javadsl/SchedulerAdapter.scala
meln1k: The usage example will be in the follow-up PR. I think this PR is already too big to include it.

core/src/main/scala/com/mesosphere/usi/core/scaladsl/SchedulerAdapter.scala
meln1k: Usage example will be provided in the follow-up PR.

core/src/main/scala/com/mesosphere/usi/core/javaapi/SchedulerAdapter.scala
meln1k: I don't think it's necessary to write it in Java, Scala classes are totally usable from Java if we follow a few principles and don't overuse scala features.

core/src/main/scala/com/mesosphere/usi/core/scalaapi/SchedulerAdapter.scala
zen-dog: Let's make it configurable 👍

core/src/main/scala/com/mesosphere/usi/core/javaapi/SchedulerAdapter.scala
timcharper: I'd really like to see some Java unit tests that use this

core/src/test/scala/com/mesosphere/usi/core/scalaapi/SchedulerAdapterTest.scala
timcharper: Why this change?

core/src/test/scala/com/mesosphere/usi/core/scalaapi/SchedulerAdapterTest.scala
timcharper: It's actually

```
 specInputQueue.pull().futureValue.get._1 shouldBe empty
```

timcharper: http://www.scalatest.org/user_guide/using_matchers

core/src/test/scala/com/mesosphere/usi/core/scalaapi/SchedulerAdapterTest.scala
timcharper: let's deconstruct

```
Source.repeat(specEvent).runWith(sink)

val done = inside(specInputQueue.pull().futureValue) {
  case (_, events) => events.runWith(Sink.head)
}

done.futureValue shouldEqual specEvent
```

etc

core/src/test/scala/com/mesosphere/usi/core/scalaapi/SchedulerAdapterTest.scala
timcharper: This is one of those dumb things where Americans spell it one way, and European spell it another way

core/src/test/scala/com/mesosphere/usi/core/scalaapi/SchedulerAdapterTest.scala
timcharper:
```
 val (_, Some(innerSource)) = specInputQueue.pull().futureValue
```

timcharper: Also, using _ in deconstruction assignment actually creates a variable (which is seen is unused) with that name. Instead, use Inside

core/src/main/scala/com/mesosphere/usi/core/scalaapi/SchedulerAdapter.scala
meln1k: No, let's not. If those values will be changed it will break the code. I'll add a comment about why it shouldn't be touched.

core/src/main/scala/com/mesosphere/usi/core/scalaapi/SchedulerAdapter.scala
meln1k: If we extract it, this code might pop up in other places :) I can make it private though.

core/src/test/scala/com/mesosphere/usi/core/scalaapi/SchedulerAdapterTest.scala
meln1k: I feel like defining an implicit value that extend `trait Emptiness[-T]` is a bit overkill for this case.

core/src/test/scala/com/mesosphere/usi/core/scalaapi/SchedulerAdapterTest.scala
meln1k: @jeschkies I personally find `f.method_call` pattern a bit verbose. Do you think we should agree to use/not use it? For example, in Marathon, we use `in new Fixture {` pattern in many tests.

core/src/test/scala/com/mesosphere/usi/core/scalaapi/SchedulerAdapterTest.scala
meln1k: Well, I'd be consistent with the `Sink.cancelled` method in akka.

core/src/main/scala/com/mesosphere/usi/core/scalaapi/SchedulerAdapter.scala
meln1k: Yeah, looks like a rare interface. I'll get rid of it!

core/src/main/scala/com/mesosphere/usi/core/scalaapi/SchedulerAdapter.scala
meln1k: I think the best way to document this would be to have an example of usage.

core/src/main/scala/com/mesosphere/usi/core/javaapi/SchedulerAdapter.scala
```diff
@@ -0,0 +1,83 @@
...
+    * @param specsSnapshot Snapshot of the current specs
+    * @return Snapshot of the current state, as well as Source which produces StateEvents and Sink which accepts SpecEvents
+    */
+  def asSourceAndSink(specsSnapshot: SpecsSnapshot)
+    : Tuple3[CompletableFuture[StateSnapshot], Source[StateEvent, NotUsed], Sink[SpecUpdated, NotUsed]] = {
+    val (snap, source, sink) = delegate.asSourceAndSink(specsSnapshot)
+    Tuple3(snap.toJava.toCompletableFuture, source.asJava, sink.asJava)
+  }
+
+  def asSourceAndSink()
```
alenkacz: I still don't understand why we explicitly support providing no snapshot. I think it's pretty dangerous :) I would rather all orchestration code had to explicitly provide the empty one when that's really what's needed...

alenkacz (commented):
Apart from that snapshot point I would still prefer to have a disucssion whether this code should not be directly part of core, with droping the necessity to have `SchedulerAdapter(flow).asSourceAndSink`

plus I don't get why we need the queues as well as sink/source one.

jeschkies (changes requested):
After tinkering a little bit with the Adapter I've found that these two adapters would be the easiest to use

```scala
def simpleFlow(specsSnapshot: SpecsSnapshot, client: MesosClient): Future[Tuple2[StateSnapshot, Flow[SpecUpdated, StateEvent]]] = ???
```

which would allow

```scala
val specsSnapshot = recover()
(state, schedulerFlow) = await(simpleFlow(specsSnapshot, client))

mySpecUpdates.via(schedulerFlow).runWith(Sink.foreach(???))
```

The other interface is

```java
class NonAsyncAdapter {
  // Times out
  public void offer(SpecUpdated update, long timeout, TimeUnit unit) throws InterruptedException
  public StateEvent poll(long timeout, TimeUnit unit) throws InterruptedException

  // Blocking
  public StateEvent take() throws InterruptedException
  public void add(SpecUpdated update) throws InterruptedException
}
```

This is nice for legacy code that cannot handling asynchronous calls well. Please note that the Java adapter has no single methods as the current adapter.

jeschkies:
Please see mesosphere/mesos-plugin#27 how it looks like for Jenkins. It much better.

timcharper:
Sorry to hold this up further; I'm working on a design doc to change the interface for USI to accept commands rather than replicate a set of podSpecs. This will further impact the adapter interface and may render it largely unnecessary.

timcharper:
We can still go ahead with the merge though, and just change the interface as needed

timcharper (approved):

jeschkies (approved):

meln1k:
OMG that was merged!!! 🎉 🎉 🎉

PR #62, opened 2019-03-07
Move resource classes to models

alenkacz:
All Resource classes are moved to models because it's part of the interface. Also, the implementation logic is pulled out of the classes to separate functions inside the core module.

I did not touch any implementation, I just shuffled the code around.


zen-dog (approved):
👏

jeschkies (approved):

PR #64, opened 2019-03-08
Introduce basic `PodSpec` validation

alenkacz:
My proposal is:
- never let anyone create invalid (not-validated) object, thus we have PodSpec trait and private impl (if we agree on this direction I will do the same for RunSpec - thanks Aleksey for that feedback)
- I went with `Either[Seq[ValidationError], PodSpec]` - I think it's pretty much equal to `PodSpec Or Every[ErrorMessage]`
- All validation methods now return `Seq[ValidationError]` thus supporting one or many errors (I don't see reason why we would need to distinguish between one or more errors on this level) ... I find that nicer than `Unit Or Every[ErrorMessage]`
- I've added 5 lines long `validationErrorOrResult` method that does the accumulation

### DIFF
--- a/core-models/src/main/scala/com/mesosphere/usi/core/models/PodSpec.scala
+++ b/core-models/src/main/scala/com/mesosphere/usi/core/models/PodSpec.scala
@@ -1,6 +1,6 @@
 package com.mesosphere.usi.core.models
 
-import com.mesosphere.usi.core.models.resources.{ExactValue, RangeRequirement}
+import com.mesosphere.usi.core.models.resources.{ExactValue, RangeRequirement, ResourceRequirement, ScalarRequirement}
 
 import scala.util.{Failure, Success, Try}
 
@@ -24,7 +24,7 @@ trait PodSpec {
 
 object PodSpec {
 
-  type ErrorMessage = String
+  type ValidationError = String
 
   /**
     * Verifies that every value in range is requested only once. Requesting same value multiple times would yield pod that will never be scheduled.
@@ -32,24 +32,44 @@ object PodSpec {
     * @param runSpec runSpec we are validating
     * @return true if provided range requirements are valid
     */
-  private def validateRunSpec(runSpec: RunSpec): Option[ErrorMessage] = {
+  private def validateStaticRangeRequirementsUnique(runSpec: RunSpec): Seq[ValidationError] = {
     val staticPorts = runSpec.resourceRequirements.collect {
       case RangeRequirement(requestedValues, _, _) => requestedValues
     }.flatten.collect { case ExactValue(value) => value }
 
     if (staticPorts.distinct.length != staticPorts.length) {
-      Some(s"Every value inside RangeResource can be requested only once. Requirement: ${staticPorts.mkString(",")}")
+      Seq(s"Every value inside RangeResource can be requested only once. Requirement: ${staticPorts.mkString(",")}")
+    } else {
+      Seq.empty
     }
-
-    None
   }
 
-  def apply(id: PodId, goal: Goal, runSpec: RunSpec): Try[PodSpec] = {
-    val validationError = validateRunSpec(runSpec)
+  private def validateScalarRequirements(resourceRequirements: Seq[ResourceRequirement]): Seq[ValidationError] = {
+    val invalidScalar = resourceRequirements.collect {
+      case s: ScalarRequirement => s
+    }.filter(s => s.amount < 0)
+    if (invalidScalar.nonEmpty) {
+      Seq(s"Scalar values cannot be smaller than 0. Invalid requirements: ${invalidScalar
+        .map(s => s"${s.resourceType}:${s.amount}")
+        .mkString(",")}")
+    } else {
+      Seq.empty
+    }
+  }
 
-    validationError match {
-      case Some(error) => Failure(new IllegalArgumentException(error))
-      case None => Success(PodSpecImpl(id, goal, runSpec))
+  def apply(id: PodId, goal: Goal, runSpec: RunSpec): Either[Seq[ValidationError], PodSpec] = {
+    val uniqueRangeRequirements = validateStaticRangeRequirementsUnique(runSpec)
+    val scalarRequirements = validateScalarRequirements(runSpec.resourceRequirements)
+
+    validationErrorOrResult(uniqueRangeRequirements, scalarRequirements)(() => PodSpecImpl(id, goal, runSpec))
+  }
+
+  private def validationErrorOrResult(validationResults: Seq[ValidationError]*)(createFunc: () => PodSpec): Either[Seq[ValidationError], PodSpec] = {
+    val validationErrors = validationResults.flatten
+    if (validationErrors.nonEmpty) {
+      Left(validationErrors)
+    } else {
+      Right(createFunc())
     }
   }
 
--- a/core-models/src/main/scala/com/mesosphere/usi/core/models/PodSpec.scala
+++ b/core-models/src/main/scala/com/mesosphere/usi/core/models/PodSpec.scala
@@ -1,5 +1,9 @@
 package com.mesosphere.usi.core.models
 
+import com.mesosphere.usi.core.models.resources.{ExactValue, RangeRequirement}
+
+import scala.util.{Failure, Success, Try}
+
 /**
   * Framework implementation owned specification of some Pod that should be launched.
   *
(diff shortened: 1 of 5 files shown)


core-models/src/main/scala/com/mesosphere/usi/core/models/PodSpec.scala
```diff
@@ -12,4 +16,42 @@ package com.mesosphere.usi.core.models
...
+
+  type ErrorMessage = String
+
+  /**
+    * Verifies that every value in range is requested only once. Requesting same value multiple times would yield pod that will never be scheduled.
+    * E.g. requesting two ports 80 on one machine is just not possible to satisfy
+    * @param runSpec runSpec we are validating
+    * @return true if provided range requirements are valid
+    */
+  private def validateRunSpec(runSpec: RunSpec): Option[ErrorMessage] = {
```
zen-dog: I remember talking with @jeschkies about alternatives to wix-validation for Marathon and we both were fond of [Scalactic's](http://www.scalactic.org/user_guide/OrAndEvery) `Good, Bad, One, Many and Every` types. They are pretty simple, yet powerful abstractions especially when it comes to multiple validation errors and their messages. Wdyt?

alenkacz: I think the validation is pretty basic for now so we could be fine with a simple option. Also as a library, I am always trying to depend on as small amount of dependencies as possible to make it easier for people integrating.

But I am not against that library if other feel like we absolutely need it to maintain readable validation code :)

To sum up: my vote would be for no library :) (if it does not lead to much more readable code...)

zen-dog: I'd like to lay down a good pattern for validation. `Option` only brings you so far. The best example is multiple error messages (in case you have multiple errors) - scalactic already elegantly solves this for you.

jeschkies: I agree with Alex. `Option`, `Either` and `Try` are not that powerful for validations. There's a reason the Scala ecosystem has a ton. I like Scalactic's for its simplicity.

core-models/src/main/scala/com/mesosphere/usi/core/models/PodSpec.scala
```diff
@@ -12,4 +16,42 @@ package com.mesosphere.usi.core.models
...
+    }.flatten.collect { case ExactValue(value) => value }
+
+    if (staticPorts.distinct.length != staticPorts.length) {
+      Some(s"Every value inside RangeResource can be requested only once. Requirement: ${staticPorts.mkString(",")}")
+    }
+
+    None
+  }
+
+  def apply(id: PodId, goal: Goal, runSpec: RunSpec): Try[PodSpec] = {
```
meln1k: I think `Either[ValidationMessage, PodSpec]` might be better here since it explicitly tells you what to expect in case of unsuccessful validation.

core-models/src/main/scala/com/mesosphere/usi/core/models/PodSpec.scala
```diff
@@ -12,4 +16,42 @@ package com.mesosphere.usi.core.models
...
+    val validationError = validateRunSpec(runSpec)
+
+    validationError match {
+      case Some(error) => Failure(new IllegalArgumentException(error))
```
jeschkies: I'd much rather use `require` as that's the simplest thing if we won't go with a proper validation library.

jeschkies (changes requested):

core-models/src/main/scala/com/mesosphere/usi/core/models/PodSpec.scala
zen-dog: I'd rather introduce proper validation since we have to do it at some point anyway ;) But, otherwise, `require` would be my choice too.

core-models/src/main/scala/com/mesosphere/usi/core/models/PodSpec.scala
```diff
@@ -12,4 +16,62 @@ package com.mesosphere.usi.core.models
...
+  }
+
+  def apply(id: PodId, goal: Goal, runSpec: RunSpec): Either[Seq[ValidationError], PodSpec] = {
+    val uniqueRangeRequirements = validateStaticRangeRequirementsUnique(runSpec)
+    val scalarRequirements = validateScalarRequirements(runSpec.resourceRequirements)
+
+    validationErrorOrResult(uniqueRangeRequirements, scalarRequirements)(() => PodSpecImpl(id, goal, runSpec))
+  }
+
+  private def validationErrorOrResult(validationResults: Seq[ValidationError]*)(createFunc: () => PodSpec): Either[Seq[ValidationError], PodSpec] = {
```
jeschkies, line 67: So we are building our own validation library? 😉

alenkacz: no we implemented 5 lines of code instead of pulling in generic utility library exactly because of these 5 lines of code. 

My argument is that Scalactic is giving us like 3 extra types plus this method. We don't need the types because we have all of them in pure scala. And implementing these extra  5 lines is not enough for me as an extra dependency.

Plus the types in scalactic are really ugly :)

We can discuss pulling in REAL validation library (although I wouldn't do that at this point) but scalactic really makes no sense to me.

core-models/src/main/scala/com/mesosphere/usi/core/models/PodSpec.scala
```diff
@@ -12,4 +16,62 @@ package com.mesosphere.usi.core.models
   * @param goal target goal of this pod. See [[Goal]] for more details
   * @param runSpec WIP the thing to run, and resource requirements, etc.
   */
-case class PodSpec(id: PodId, goal: Goal, runSpec: RunSpec)
+trait PodSpec {
```
zen-dog, line 19: Independent of the rest of this PR - could we please change it back to `case class` and give a private constructor?

core-models/src/main/scala/com/mesosphere/usi/core/models/PodSpec.scala
jeschkies: I’m just afraid that we’ll put more stuff on top instead of following some common pattern.

alenkacz: then we should go for a proper validation library at that point. My argument here is that we don't need those types from Scalactic. Maybe at one point full featured validation library would be a way to go although I kind of like the simplicity and the conventions outlined here...

core-models/src/main/scala/com/mesosphere/usi/core/models/PodSpec.scala
alenkacz: I think this is how you do private constructor for case classes in scala. Stackoverflow agrees :) https://stackoverflow.com/questions/20030826/scala-case-class-private-constructor-but-public-apply-method

jeschkies: Really?

```scala
case class Foo private (b: String)
Foo("bar") // passes
new Foo("bar") // fails
```
WTF. I sometimes hate Scala. Anyhow, Java won't profit from any so I propose to use the builder pattern.

jeschkies: Ah, I just talked with @zen-dog. It works like this

```scala
case class Foo private (b: String)
object Foo {
  def apply(b: String) = {
    throw new NotImplementedException("I overrode the default factory method")
  }
}
Foo("bar") // throws exception
new Foo("bar") // fails
```

core/src/test/scala/com/mesosphere/usi/core/logic/MesosEventsLogicTest.scala
```diff
@@ -14,13 +14,14 @@ class MesosEventsLogicTest extends UnitTest {
...
   private val mesosEventLogic = new MesosEventsLogic(new MesosCalls(MesosMock.mockFrameworkId))
 
-  val podWith1Cpu256Mem: PodSpec = PodSpec(
-    PodId("mock-podId"),
-    Goal.Running,
-    RunSpec(
-      List(ScalarRequirement(ResourceType.CPUS, 1), ScalarRequirement(ResourceType.MEM, 256)),
-      shellCommand = "sleep 3600")
-  )
+  def podWith1Cpu256Mem(id: String = "mock-podId"): PodSpec =
```
zen-dog, line 17: here and everybody else: could we please generate some random ids by default? I know, these are not IT but this will 100% copy-pasted later. We don't need the UUID, it can be something alphanumeric too:
```suggestion
  def podWith1Cpu256Mem(id: String = s"mock-podId-${scala.util.Random.alphanumeric.take(10).mkString}"): PodSpec =
```

zen-dog (changes requested):
I would like to see the tests for "unhappy" validation paths. Otherwise lgtm! Thx for your patience with all the discussions around validation!

core/src/test/scala/com/mesosphere/usi/core/logic/MesosEventsLogicTest.scala
timcharper: @zen-dog let's bring that up as a requirement for the fixture story 

https://jira.mesosphere.com/browse/DCOS-50134

core-models/src/main/scala/com/mesosphere/usi/core/models/PodSpec.scala
timcharper: My vote is a firm no for using a validation library.

core/src/test/scala/com/mesosphere/usi/core/logic/MesosEventsLogicTest.scala
timcharper: (I added it to the proposal)

core-models/src/main/scala/com/mesosphere/usi/core/models/PodSpec.scala
timcharper: My vote is that we don't make the constructor private, and we don't force validation. Let's talk about this more tomorrow in a meeting.

timcharper (changes requested):
Let's discuss tomorrow; I don't think (5% of this) is the way we should go

PR #65, opened 2019-03-10
Add support for Mesos fetcher

zen-dog:
Summary:
`RunSpec` gets a `fetch` field which is a sequence of `FetchUri` describing a specification used by the [Mesos fetcher](http://mesos.apache.org/documentation/latest/fetcher/) to download resources into the sandbox directory of a task in preparation of running the task. This will be needed by the soon by the new jenkins-plugin (to fetch the agent.jar)

JIRA: DCOS_OSS-4893


alenkacz (approved):
LGTM! Could you please add some test though? Either fetch used in integration test or at least a unit test for the builder or mesoseventlogic? Thanks

core/src/main/scala/com/mesosphere/usi/core/protos/ProtoBuilders.scala
```diff
@@ -232,17 +232,38 @@ private[usi] object ProtoBuilders {
     b.build()
   }
 
+  def newCommandInfo(shellCommand: String, fetchUri: Seq[FetchUri]): Mesos.CommandInfo = {
+    val b = Mesos.CommandInfo
```
kensipe, line 236: it would be great if we used words instead of letters for object references / vals.     It looks like this habit started before this PR...

kensipe (commented):
Agree that we should have some tests

core/src/main/scala/com/mesosphere/usi/core/protos/ProtoBuilders.scala
zen-dog: Lol, believe it or not but my method called it `builder` at first and I even started to rename other methods to it but then I realized that if *all* methods call it the same, it's a clear pattern and it doesn't matter whether it is `b` or `builder` (and `b` actually saves you some screen space)

timcharper: I'm really fine with `b` for builders. It's pretty standard and it's not difficult to track in this context. (if there were lots of different builders in scope, or different one-letter variables, I would feel differently)

timcharper:
This is a great start. I've started to work on the plan for launching pods with multiple tasks (DCOS-48502) which includes a more detailed design for how CommandInfo generators are defined. This will affect this interface; we still have lots of flexibility to change the API until we release 1.0 so I'm okay with this.

core-models/src/main/scala/com/mesosphere/usi/core/models/FetchUri.scala
```diff
@@ -0,0 +1,36 @@
...
+  *
+  * @param uri artifact URI to download
+  * @param extract if true, the artifact will be treated as an archive an extracted to the sandbox. See the list
+  *                of the supported archive types below
+  * @param executable if true, the fetch result will be changed to be executable (by “chmod”) for every user
+  * @param cache if true, the fetcher cache is to be used for the URI
+  * @param outputFile if set, the fetcher will use that name for the copy stored in the sandbox directory
+  */
+case class FetchUri(
+    uri: URI,
```
meln1k, line 18: why not akka.http.Uri? it's more strict in following the RFC 3986.

zen-dog: because it's Java-friendly. also, where exactly it breaks the RFC?

meln1k (approved):

PR #66, opened 2019-03-11
PodSpec validation with scalactic

alenkacz:


core-models/src/main/scala/com/mesosphere/usi/core/models/PodSpec.scala
```diff
@@ -12,4 +17,63 @@ package com.mesosphere.usi.core.models
...
+    }
+  }
+
+  private def validateScalarRequirements(resourceRequirements: Seq[ResourceRequirement]): Unit Or One[ErrorMessage] = {
+    val invalidScalar = resourceRequirements.collect {
+      case s: ScalarRequirement => s
+    }.filter(s => s.amount < 0)
+
+    invalidScalar match {
+      case Nil => Good(())
```
jeschkies, line 55: I can see why you don't like it. I think this should actually return the resources. `Unit Or One[...]` does not seem right to me. `Or` and `Every` should not be part of a validation method but part of factory methods.

alenkacz: but how do you accumulate then? All accumulating method expect to work with `Good Or Bad`

alenkacz: If you return `Option[ErrorMessage]` here I think you can do that without Scalactic with the same result

core-models/src/main/scala/com/mesosphere/usi/core/models/PodSpec.scala
```diff
@@ -12,4 +17,63 @@ package com.mesosphere.usi.core.models
   * @param goal target goal of this pod. See [[Goal]] for more details
   * @param runSpec WIP the thing to run, and resource requirements, etc.
   */
-case class PodSpec(id: PodId, goal: Goal, runSpec: RunSpec)
```
meln1k: Why the case class was removed?

core-models/src/main/scala/com/mesosphere/usi/core/models/PodSpec.scala
```diff
@@ -12,4 +17,63 @@ package com.mesosphere.usi.core.models
...
+      (_, _) => ())
+  }
+
+  def apply(id: PodId, goal: Goal, runSpec: RunSpec): PodSpec Or Every[ErrorMessage] = {
+    val resourceValidation = validateResources(runSpec.resourceRequirements)
+
+    resourceValidation.map(_ => PodSpecImpl(id, goal, runSpec))
+  }
+
+  private case class PodSpecImpl(id: PodId, goal: Goal, runSpec: RunSpec) extends PodSpec
```
jeschkies, line 78: Not sure I understand the private class. What do you think about a private constructor instead?

jeschkies (changes requested):

core-models/src/main/scala/com/mesosphere/usi/core/models/PodSpec.scala
alenkacz: this is a way to prevent one from instantiating outside of that factory method with validation. How do you do private constructor in case classes? 🤔 I thought this is the way to go.

meln1k (approved):
LGTM

zen-dog:
Closed in favour of #64

PR #67, opened 2019-03-12
Add a section about testing granularity to the coding guidelines

zen-dog:
Additionally fixed some minor markdown errors.

### DIFF
--- a/docs/code-culture.md
+++ b/docs/code-culture.md
@@ -27,7 +27,6 @@ Prefer:
 * Explicit conversion over implicit
 * Explicit passing over implicit
 
-
 For our case, if implicits don't help improve the design of code, we don't use them. We avoid implicit conversions. We also avoid multi-priority implicits. When type classes help improve the design of code, we use implicits for type classes (IE serialization logic).
 
 We prefer implicits to be used for contextual values such as execution contexts or the actor system. However, we prefer to be conservative, and pass the value explicitly for things like validators, authenticators, and normalizers.
@@ -38,7 +37,7 @@ If you are writing some function chain that has multiple levels of inferred func
 
 Prefer:
 
-```
+```scala
 groups.map { group: Group =>
   group.apps
     .groupBy(_.container.image)
@@ -50,7 +49,7 @@ groups.map { group: Group =>
 
 Over:
 
-```
+```scala
 groups.map { group =>
   group.apps.groupBy(_.container.image).map { case(k, v) =>
     k -> v.length
@@ -70,7 +69,7 @@ We prefer, almost always, that imports go at the top of the file. Additionally, 
 
 Prefer:
 
-```
+```scala
 import org.company.{
   Stringifier,
   Demuxer
@@ -79,7 +78,7 @@ import org.company.{
 
 Over:
 
-```
+```scala
 import org.company._
 ```
 
@@ -114,7 +113,7 @@ It is better to do nothing, than to do the wrong thing.
 
 One of my favorite non-examples of this is found in PHP 5:
 
-```
+```bash
 $ echo '<? print(strftime("%Y-%m-%d",strtotime("lol"))) ?>' | php
 1970-01-01
 ```
@@ -145,9 +144,10 @@ Debugging distributed system is hard. To make this at least a bit easier we favo
 For introduction to structured logging read [this article](https://stackify.com/what-is-structured-logging-and-why-developers-need-it/).
 
 All relevant ids should be part of the logging message e.g.:
-- pod ID ("podId")
-- mesos task ID ("taskId")
-- mesos offer ID ("offerId")
+
+* pod ID ("podId")
+* mesos task ID ("taskId")
+* mesos offer ID ("offerId")
 
 It's important to keep the naming of keys consistent so that one can use them for log pre-selection.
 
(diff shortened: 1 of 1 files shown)


docs/code-culture.md
```diff
@@ -164,3 +164,66 @@ Example of good framework log messages (a little bit simplified) can look like e
   "message": "Task $taskId of Pod $podId was killed"
 }
 ```
+
+## Testing
+
+Tests are executable documentation and should be **written, read and maintained** as such. A test tells it's reader a short story about how some part of the system should behave given a certain input. Hundreeds of excelent books and articles are dedicated to writing proper tests so we'll try to avoid repeating them here.
```
alenkacz: Hundreds

docs/code-culture.md
```diff
@@ -164,3 +164,66 @@ Example of good framework log messages (a little bit simplified) can look like e
...
+### On Libraries and Styles
+
+We heavily utilize [ScalaTest](http://www.scalatest.org/) as our primary test library. When it comes to the test style we prefer [WordSpec](http://www.scalatest.org/at_a_glance/WordSpec)
+
+```scala
+  // Describe a scope for a subject, in this case: "A Set"
+  "A Set" can { // All tests within these curly braces are about "A Set"
+
+    // Can describe nested scopes that "narrow" its outer scopes
+    "empty" should { // All tests within these curly braces are about "A Set (when empty)"
```
alenkacz: do we really need two levels of nesting? I think so far one was enough. I never had to use the "can" level...

docs/code-culture.md
```diff
@@ -164,3 +164,66 @@ Example of good framework log messages (a little bit simplified) can look like e
...
+
+      "have size 0" in {    // Here, 'it' refers to "A Set (when empty)". The full name
+        Set.empty.size shouldBe 0 // of this test is: "A Set (when empty) should have size 0"
```
alenkacz: maybe we could come up with a different example? Because for this case, I think `shouldBe empty` is a better approach :)

docs/code-culture.md
```diff
@@ -164,3 +164,66 @@ Example of good framework log messages (a little bit simplified) can look like e
...
+
+### On Testing Granularity
+
+In USI we define following granularity levels:
+
+* **Unit tests**: the lowest test level possible. Unit test are cheap, can be run fast and give the developer an immediate feedback on when something is obviously broken
+* **Integration tests**: an integration test starts a minimal Mesos cluster consisting of Mesos master, Mesos agent, in-memory Zookeeper server and a framework. These are more expensive tests that typically require more resources and time to run. Tests that require back and forth communication with Mesos are best placed here because we don't want to mock Mesos responses.
+* **System integraiton test**: here we start a full-blown DC/OS cluster testing all aspects of framework interacting with the DC/OS ecosystem. These are the most expensive tests (in terms of time and money) that would typically reqire a DC/OS cluster running on e.g. AWS utilizing multiple EC2 nodes, volumes, ELB etc. A system integration test would typically cover some coarse-grained USI feature or a feature that relies on other DC/OS components like the secret store.
```
alenkacz, line 220: do we have those? I think these don't make that much sense for USI

alenkacz (approved):
Nice writeup, I have only nits.

I don't know how to use suggestions so you have my feedback only as comments, sorry :)

docs/code-culture.md
zen-dog: Agreed. Also added a comment about number of nested sections

docs/code-culture.md
zen-dog: I know! But I also failed to come up with a better example and left it like this for now

docs/code-culture.md
zen-dog: We don't have them now but I think we will need them. And below, I argue why this is the case.

docs/code-culture.md
timcharper:
```suggestion
Tests are executable documentation and should be **written, read and maintained** as such. A test tells its reader a short story about how some part of the system should behave given a certain input. Hundreds of excellent books and articles are dedicated to writing proper tests, so we'll try to avoid repeating them here.
```

timcharper: it's vs its is the worst. Every time I read "it's" I mentally read aloud "it is".

docs/code-culture.md
```diff
@@ -164,3 +164,66 @@ Example of good framework log messages (a little bit simplified) can look like e
...
 }
 ```
+
+## Testing
+
+Tests are executable documentation and should be **written, read and maintained** as such. A test tells it's reader a short story about how some part of the system should behave given a certain input. Hundreeds of excelent books and articles are dedicated to writing proper tests so we'll try to avoid repeating them here.
+
+### On Libraries and Styles
+
+We heavily utilize [ScalaTest](http://www.scalatest.org/) as our primary test library. When it comes to the test style we prefer [WordSpec](http://www.scalatest.org/at_a_glance/WordSpec)
```
timcharper:
```suggestion
We heavily utilize [ScalaTest](http://www.scalatest.org/) as our primary test library. All tests should implement the UnitTest class in test-utils, which uses [WordSpec](http://www.scalatest.org/at_a_glance/WordSpec).
```

docs/code-culture.md
```diff
@@ -164,3 +164,66 @@ Example of good framework log messages (a little bit simplified) can look like e
...
+* **System integraiton test**: here we start a full-blown DC/OS cluster testing all aspects of framework interacting with the DC/OS ecosystem. These are the most expensive tests (in terms of time and money) that would typically reqire a DC/OS cluster running on e.g. AWS utilizing multiple EC2 nodes, volumes, ELB etc. A system integration test would typically cover some coarse-grained USI feature or a feature that relies on other DC/OS components like the secret store.
+
+**As a rule of thumb a broken behavior should fail at the lowest possible level**. Most features will be covered on more than one level but the coverage is different. Let's consider support for [Mesos fetcher](http://mesos.apache.org/documentation/latest/fetcher/) as an example. The Mesos fetcher is a mechanism to download resources into the sandbox directory of a task in preparation of running the task. So how do the tests look like?
```
timcharper:
```suggestion
**As a rule of thumb, a broken behavior should fail at the lowest possible level**. Most features will be covered on more than one level but the coverage is different. Let's consider support for [Mesos fetcher](http://mesos.apache.org/documentation/latest/fetcher/) as an example. In preparation to run a task, the Mesos fetcher downloads resources into the task's sandbox directory. So how do the tests look like?
```

docs/code-culture.md
```diff
@@ -164,3 +164,66 @@ Example of good framework log messages (a little bit simplified) can look like e
...
+
+* **Unit test**: makes sure that given a `PodSpec` with a defined fetch URI it is converted to a Mesos protobuf message, where correspoding [URI](https://github.com/apache/mesos/blob/master/include/mesos/mesos.proto#L658) fields are initialized properly
+* **Integration test**: makes sure that when a task from a `PodSpec` is started, the fetched artifact is actually part of it's sandbox
```
timcharper:
```suggestion
* **Integration test**: makes sure that when a task from a `PodSpec` is started, the fetched artifact is actually part of its sandbox
```

docs/code-culture.md
```diff
@@ -164,3 +164,66 @@ Example of good framework log messages (a little bit simplified) can look like e
...
+* **System integration test**: this is a tricky one - do we need full DC/OS cluster to test this? How is this different from the integration test from the fetchers perspective? The simple answer is - it's not and an integration test might be sufficient enough
```
timcharper:
```suggestion
* **System integration test**: this is a tricky one; do we need full DC/OS cluster to test this? How is this different from the integration test from the fetchers perspective? The simple answer is that it's not, and an integration test might be sufficient enough.
```

docs/code-culture.md
```diff
@@ -164,3 +164,66 @@ Example of good framework log messages (a little bit simplified) can look like e
...
+
+However consider following aspects (which are all taken from prior experience building Mesos frameworks):
```
timcharper:
```suggestion
However consider the following aspects (which are all taken from prior experience building Mesos frameworks):
```

docs/code-culture.md
```diff
@@ -164,3 +164,66 @@ Example of good framework log messages (a little bit simplified) can look like e
...
+
+* while an integration test usually starts a some stable Mesos version that USI officially depends on, DC/OS frequently integrates latest Mesos changes into its master branch. Testing a framework against the latest DC/OS master might expose a bug sooner rather then later
```
timcharper:
```suggestion
* While an integration test usually starts some stable Mesos version on which USI officially depends, DC/OS frequently integrates the latest Mesos changes into its master branch. Testing a framework against the latest DC/OS master might expose a bug sooner rather than later.
```

docs/code-culture.md
```diff
@@ -164,3 +164,66 @@ Example of good framework log messages (a little bit simplified) can look like e
...
+* an integration test runs against a local Mesos cluster where communication is typically fast and reliable. A system integration test runs against a cluster somewhere in the cloud and has different communications profile. In the past we saw bugs that would only manifest themselves in the later case but not in the former. Admited, it seems unlikely that such a bug is triggered in the Mesos fetcher but we've seen unlikely things happen before
```
timcharper:
```suggestion
* An integration test runs against a local Mesos cluster where communication is typically fast and reliable. A system integration test runs against a cluster somewhere in the cloud, and has a different communications profile. In the past, we've seen bugs that would only manifest themselves in the latter case, but not in the former. Admittedly, it seems unlikely that such a bug is triggered in the Mesos fetcher; however, we've seen unlikely things happen before.
```

docs/code-culture.md
```diff
@@ -164,3 +164,66 @@ Example of good framework log messages (a little bit simplified) can look like e
...
+* even a simple request to the framework running on a DC/OS cluster touches many components on it's way e.g. ELB, Admin Router, Monitoring service which themselves rely on other components (e.g. Admin Router relies on the CockroachDB) so there is always potential for things to go wrong
```
timcharper:
```suggestion
* Even a simple request to the framework running on a DC/OS cluster touches many components on its way (e.g. ELB, Admin Router, Monitoring service which themselves rely on other components, such as Admin Router relying on CockroachDB), so there is always potential for things to go wrong.
```

docs/code-culture.md
```diff
@@ -164,3 +164,66 @@ Example of good framework log messages (a little bit simplified) can look like e
...
+
+It makes sense to have a system integration test for every user facing coarse-grained API feature including the example above.
```
timcharper:
```suggestion
It makes sense to have a system integration test for every user-facing coarse-grained API feature, including the example above.
```

timcharper (changes requested):
Many suggestions (grammar etc). Also, I'd like the following expressed:

* Choose the lightest testing mechanism that suits the need; if something can be tested well at a unit level, then test there first. IE: An integration test might check that a validation error is properly returned via the API, but a unit test should check the 20 different ways a validation fails.
* Avoid mocks/stubbing. Mocks are often a code smell that happens when effects are coupled with business logic, and the tests often look too much like the implementation. If we separate our effects then we don't need mocks. Preference for fake implementations of interfaces (IE MockMesos flow) over "method X should be invoked once with param Y and return Z".

docs/code-culture.md
zen-dog: It's (!) my kryptonite 😞

docs/code-culture.md
zen-dog: Funny, this sentence about Mesos fetcher is taken directly from [Mesos documentation](http://mesos.apache.org/documentation/latest/fetcher/#what-is-the-mesos-fetcher)

timcharper (approved):

PR #68, opened 2019-03-12
Add multi-role suppress/revive

alenkacz:
This does not implement debounce. That will be a follow-up.

JIRA: DCOS-47475


core/src/main/scala/com/mesosphere/usi/core/SchedulerLogicHandler.scala
```diff
@@ -118,15 +118,33 @@ private[core] class SchedulerLogicHandler(mesosCallFactory: MesosCalls) {
...
+  def generateSuppressCalls(pendingLaunch: Set[PodId], newLaunched: Set[PodId]): List[Call] = {
+    val alreadyLaunchedRoles = newLaunched
+      .map(id => specs.podSpecs.get(id))
+      .collect { case Some(p) => p.runSpec.role }
+
+    val rolesBeingLaunched = pendingLaunch
+      .map(id => specs.podSpecs.get(id))
+      .collect { case Some(p) => p.runSpec.role }
+
+    alreadyLaunchedRoles.diff(rolesBeingLaunched).map(r => mesosCallFactory.newSuppress(Some(r))).toList
```
timcharper: Let's avoid creating interim collections unnecessarily

```
alreadyLaunchedRoles.iterator
  .filterNot(rolesBeingLaunched)
  .map { r =>
    mesosCallFactory.newSuppress(Some(r))
  }
  .toList
```

core/src/main/scala/com/mesosphere/usi/core/SchedulerLogicHandler.scala
```diff
@@ -118,15 +118,33 @@ private[core] class SchedulerLogicHandler(mesosCallFactory: MesosCalls) {
...
       state: SchedulerState,
       dirtyPodIds: Set[PodId]): SchedulerEvents = {
-    this.cachedPendingLaunch = this.cachedPendingLaunch.update(specs, state, dirtyPodIds)
-    if (cachedPendingLaunch.pendingLaunch.nonEmpty)
-      SchedulerEvents(mesosCalls = List(mesosCallFactory.newRevive(None)))
-    else
-      SchedulerEvents.empty
+    val updateResult = this.cachedPendingLaunch.update(specs, state, dirtyPodIds)
+    this.cachedPendingLaunch = updateResult.newCachedPendingLaunch
+    val reviveCalls = updateResult.newToBeLaunched
```
timcharper: Let's avoid interim collections

core/src/test/scala/com/mesosphere/usi/core/SchedulerTest.scala
```diff
@@ -56,7 +56,8 @@ class SchedulerTest extends AkkaUnitTest with Inside {
             Goal.Running,
             RunSpec(
               resourceRequirements = List(ScalarRequirement.cpus(1), ScalarRequirement.memory(256)),
-              shellCommand = "sleep 3600")
+              shellCommand = "sleep 3600",
+              "*")
```
timcharper: You should never launch a task as "*", Mesos will never allocate resources to that role. I know this just a test fixture, can we please change this to an actual role to avoid setting the wrong precedent?

core/src/test/scala/com/mesosphere/usi/core/logic/MesosEventsLogicTest.scala
```diff
@@ -19,7 +19,8 @@ class MesosEventsLogicTest extends UnitTest {
     Goal.Running,
     RunSpec(
       List(ScalarRequirement(ResourceType.CPUS, 1), ScalarRequirement(ResourceType.MEM, 256)),
-      shellCommand = "sleep 3600")
+      shellCommand = "sleep 3600",
+      "*")
```
timcharper: Again, this is an invalid role for a run spec

examples/core-hello-world/src/main/scala/com/mesosphere/usi/examples/CoreHelloWorldFramework.scala
```diff
@@ -100,7 +100,8 @@ class CoreHelloWorldFramework(conf: Config) extends StrictLogging {
   val podId = PodId(s"hello-world.${UUID.randomUUID()}")
   val runSpec = RunSpec(
     resourceRequirements = List(ScalarRequirement.cpus(0.1), ScalarRequirement.memory(32)),
-    shellCommand = """echo "Hello, world" && sleep 3600"""
+    shellCommand = """echo "Hello, world" && sleep 3600""",
+    role = "*"
```
timcharper: Again

core/src/main/scala/com/mesosphere/usi/core/SchedulerLogicHandler.scala
```diff
@@ -118,15 +118,33 @@ private[core] class SchedulerLogicHandler(mesosCallFactory: MesosCalls) {
     frameResultBuilder.result
   }
 
+  def generateSuppressCalls(pendingLaunch: Set[PodId], newLaunched: Set[PodId]): List[Call] = {
+    val alreadyLaunchedRoles = newLaunched
```
timcharper: Let's reduce interim collections

core/src/main/scala/com/mesosphere/usi/core/SchedulerLogicHandler.scala
```diff
@@ -118,15 +118,33 @@ private[core] class SchedulerLogicHandler(mesosCallFactory: MesosCalls) {
...
+      .map(id => specs.podSpecs.get(id))
```
timcharper: We could remove the optional earlier, this is less allocations for no code complexity cost
```
  .collect { case id if specs.podSpecs.contains(id) => specs.podSpecs(id).role }
```

core/src/test/scala/com/mesosphere/usi/core/SchedulerLogicHandlerTest.scala
```diff
@@ -0,0 +1,16 @@
+package com.mesosphere.usi.core
+
+import com.mesosphere.mesos.client.MesosCalls
+import com.mesosphere.usi.core.helpers.MesosMock
+import com.mesosphere.usi.core.models.{Goal, PodId, PodSpec, PodSpecUpdated, RunSpec}
+import com.mesosphere.utils.UnitTest
+
+class SchedulerLogicHandlerTest extends UnitTest {
+  "output a mesos call with revive for new podspec role" in {
```
timcharper: I would like to see some more unit tests here.

* [x] Additional pod spec definitions for the same role results In additional revive calls
* [x] An incoming offer which matches outstanding pendingLaunched pods results in a suppressed call

timcharper (changes requested):

core/src/test/scala/com/mesosphere/usi/core/SchedulerTest.scala
alenkacz: so how do you launch podspec on unreserved resources then?

core/src/main/scala/com/mesosphere/usi/core/SchedulerLogicHandler.scala
alenkacz: nice, thanks 👍

core/src/test/scala/com/mesosphere/usi/core/SchedulerTest.scala
timcharper: That is the default; You just let the allocator allocate the resources for you. You never launch as "*", you always launch as a role. And this role field specifies the role as which you are launching. That will be the role from which quota is taken. If the are resources 

Refusing unreserved resources is more of a constraint.

timcharper: (see AllocationInfo)

timcharper: https://github.com/apache/mesos/blob/master/include/mesos/v1/mesos.proto#L1238

alenkacz: got it! thanks

timcharper (approved):

core/src/main/scala/com/mesosphere/usi/core/CachedPendingLaunch.scala
```diff
@@ -10,19 +15,28 @@ case class CachedPendingLaunch(pendingLaunch: Set[PodId] = Set.empty) {
     *
     * @param podIds
     */
-  def update(specs: SpecState, state: SchedulerState, podIds: Set[PodId]): CachedPendingLaunch = {
+  def update(specs: SpecState, state: SchedulerState, podIds: Set[PodId]): UpdateResult = {
```
zen-dog: Is this change relevant to suppress/revive?

core/src/main/scala/com/mesosphere/usi/core/CachedPendingLaunch.scala
```diff
@@ -10,19 +15,28 @@ case class CachedPendingLaunch(pendingLaunch: Set[PodId] = Set.empty) {
...
     var newPendingLaunch = pendingLaunch
+    var newToBeLaunched = Set.empty[PodId]
```
zen-dog, line 17: What's the difference between `newPendingLaunch` and `newToBeLaunched`?

core/src/main/scala/com/mesosphere/usi/core/SchedulerLogicHandler.scala
```diff
@@ -118,15 +118,38 @@ private[core] class SchedulerLogicHandler(mesosCallFactory: MesosCalls) {
...
-    this.cachedPendingLaunch = this.cachedPendingLaunch.update(specs, state, dirtyPodIds)
-    if (cachedPendingLaunch.pendingLaunch.nonEmpty)
-      SchedulerEvents(mesosCalls = List(mesosCallFactory.newRevive(None)))
-    else
-      SchedulerEvents.empty
+    val updateResult = this.cachedPendingLaunch.update(specs, state, dirtyPodIds)
+    this.cachedPendingLaunch = updateResult.newCachedPendingLaunch
+    val reviveCalls = updateResult.newToBeLaunched.iterator
+      .map(id => specs.podSpecs.get(id))
+      .collect { case Some(p) => p.runSpec.role }
```
zen-dog, line 148: Here too, filter for duplicate roles

core/src/main/scala/com/mesosphere/usi/core/SchedulerLogicHandler.scala
```diff
@@ -118,15 +118,38 @@ private[core] class SchedulerLogicHandler(mesosCallFactory: MesosCalls) {
...
   private def updateCachesAndRevive(
       specs: SpecState,
       state: SchedulerState,
       dirtyPodIds: Set[PodId]): SchedulerEvents = {
-    this.cachedPendingLaunch = this.cachedPendingLaunch.update(specs, state, dirtyPodIds)
-    if (cachedPendingLaunch.pendingLaunch.nonEmpty)
-      SchedulerEvents(mesosCalls = List(mesosCallFactory.newRevive(None)))
-    else
-      SchedulerEvents.empty
+    val updateResult = this.cachedPendingLaunch.update(specs, state, dirtyPodIds)
```
zen-dog, line 144: Why `this`?

core/src/main/scala/com/mesosphere/usi/core/CachedPendingLaunch.scala
```diff
@@ -10,19 +15,28 @@ case class CachedPendingLaunch(pendingLaunch: Set[PodId] = Set.empty) {
...
+    var newLaunched = Set.empty[PodId]
     podIds.foreach { podId =>
       val shouldBeLaunched = specs.podSpecs.get(podId).exists { podSpec =>
         pendingLaunch(podSpec.goal, state.podRecords.get(podId))
       }
-      if (shouldBeLaunched)
+      if (shouldBeLaunched) {
```
zen-dog, line 23: Could you please add a comment here explaining what is going into what set and when? So much `pending` and `launched` here :D

core/src/main/scala/com/mesosphere/usi/core/SchedulerLogicHandler.scala
```diff
@@ -118,15 +118,38 @@ private[core] class SchedulerLogicHandler(mesosCallFactory: MesosCalls) {
     frameResultBuilder.result
   }
 
+  def generateSuppressCalls(pendingLaunch: Set[PodId], newLaunched: Set[PodId]): List[Call] = {
+    val alreadyLaunchedRoles = newLaunched.iterator.collect {
+      case id if specs.podSpecs.contains(id) => specs.podSpecs(id).runSpec.role
+    }
```
zen-dog, line 124: You should "add" `.toSet` or `.distinct` call here to filter out duplicate roles

core/src/main/scala/com/mesosphere/usi/core/SchedulerLogicHandler.scala
```diff
@@ -118,15 +118,38 @@ private[core] class SchedulerLogicHandler(mesosCallFactory: MesosCalls) {
...
+
+    val rolesBeingLaunched = pendingLaunch
+      .map(id => specs.podSpecs.get(id))
+      .collect { case Some(p) => p.runSpec.role }
+
+    alreadyLaunchedRoles
+      .filterNot(rolesBeingLaunched)
```
zen-dog, line 131: Could you please add types to  `rolesBeingLaunched` and `alreadyLaunchedRoles`. Otherwise `.filterNot(rolesBeingLaunched)` looks a bit like code-golf ;)

core/src/test/scala/com/mesosphere/usi/core/SchedulerLogicHandlerTest.scala
```diff
@@ -0,0 +1,45 @@
+package com.mesosphere.usi.core
+
+import com.mesosphere.mesos.client.MesosCalls
+import com.mesosphere.usi.core.helpers.MesosMock
+import com.mesosphere.usi.core.models.{Goal, PodId, PodSpec, PodSpecUpdated, RunSpec}
+import com.mesosphere.usi.core.protos.ProtoBuilders
+import com.mesosphere.utils.UnitTest
+
+class SchedulerLogicHandlerTest extends UnitTest {
+  "produce a mesos call with revive for new podspec role" in {
```
zen-dog: nit: here and below `s/mesos/Mesos`

core/src/test/scala/com/mesosphere/usi/core/SchedulerLogicHandlerTest.scala
```diff
@@ -0,0 +1,45 @@
...
+      PodSpecUpdated(podId, Some(PodSpec(podId, Goal.Running, RunSpec(Seq.empty, "", "test-role")))))
+
+    When("that pod is launched")
+    val offer = MesosMock.createMockOffer()
+    val result = handler.handleMesosEvent(ProtoBuilders.newOfferEvent(offer))
+
+    Then("suppress call is generated for that role")
+    result.mesosCalls.exists(c => c.hasSuppress && c.getSuppress.getRoles(0) == "test-role") should be(true) withClue s"Expecting suppress call with role 'test-role' but got ${result.mesosCalls}"
+
+  }
```
zen-dog, line 43: Wdyt about a mixed test:
```
Given("two pods in different roles")

And("and one of them being launched")

Then("one suppress and one revive call are generated"
```
?

zen-dog (commented):
Nice PR, needs a bit of polish though.

core/src/main/scala/com/mesosphere/usi/core/CachedPendingLaunch.scala
alenkacz: hmm yeah, I struggled with the names here. Do you have any suggestions? Basically:
- pending launch are those who are still waiting to be launched (it's a superset of all)
- new to be launched are instances waiting to be launched that were added in this frame

core/src/main/scala/com/mesosphere/usi/core/SchedulerLogicHandler.scala
alenkacz: I am wondering if we need to do this since we're going to be debouncing at later stages... but I'll do it! We'll need to allocate some extra collections because of that though...

core/src/main/scala/com/mesosphere/usi/core/SchedulerLogicHandler.scala
alenkacz: 🤷‍♀️ that was already there. Looking at the class it looks like Tim is sometimes using this to indicate that he's working with the class variable not a local one. I don't really mind but I can remove it if you want.

core/src/test/scala/com/mesosphere/usi/core/SchedulerLogicHandlerTest.scala
alenkacz: will do

alenkacz: Can this even happen? We're talking about things being emitted from a single frame and in that frame we're always processing one event so we either process one pod update (new pod to be launched) or we launch a pod (we process an offer)

core/src/main/scala/com/mesosphere/usi/core/SchedulerLogicHandler.scala
zen-dog: `this` is usually used to distinguish between a local variable (e.g. method parameter) and a class variable with the same name. if you think it makes the code more readable 🤷‍♂️

core/src/main/scala/com/mesosphere/usi/core/CachedPendingLaunch.scala
```diff
@@ -1,6 +1,11 @@
 package com.mesosphere.usi.core
 import com.mesosphere.usi.core.models.{Goal, PodId, PodRecord}
 
+case class UpdateResult(
+    newCachedPendingLaunch: CachedPendingLaunch,
```
zen-dog: I still have issues with naming here. If everything that `UpdateResult` has is _new_ then maybe drop it? How about:
```scala
pendingLaunch: CachedPendingLaunch,
readyToLaunch: Set[PodId],
justLaunched: Set[PodId]
```

core/src/main/scala/com/mesosphere/usi/core/SchedulerLogicHandler.scala
```diff
@@ -118,15 +118,42 @@ private[core] class SchedulerLogicHandler(mesosCallFactory: MesosCalls) {
...
+      .toSeq
+      .distinct
+      .map { r =>
+        mesosCallFactory.newSuppress(Some(r))
+      }
+      .toList
+  }
+
   private def updateCachesAndRevive(
```
zen-dog: more like:
```suggestion
  private def updateCachesAndSuppressAndRevive(
```

zen-dog (commented):
I left a few nits about the naming, PTAL but otherwise lgtm

zen-dog (approved):

PR #70, opened 2019-03-13
Make `core-hello-world` example framework ready for deployment

zen-dog:
Summary:
- applied application plugin so we can build a tarball with all dependencies using `gradle :core-hello-world:distTar`
- replaced materialization to the `Source.queue` with `Source.ignore`. This is makes testing harder but queue is not suitable for production (back-pressure will kick in, effectively halting the stream)
- make mesos master url overridable from the environment variable

JIRA: DCOS-47868

### DIFF
--- a/examples/core-hello-world/src/main/scala/com/mesosphere/usi/examples/CoreHelloWorldFramework.scala
+++ b/examples/core-hello-world/src/main/scala/com/mesosphere/usi/examples/CoreHelloWorldFramework.scala
@@ -5,7 +5,7 @@ import java.util.UUID
 import akka.NotUsed
 import akka.actor.ActorSystem
 import akka.stream.scaladsl.{Flow, Keep, Sink, Source}
-import akka.stream.{ActorMaterializer, ActorMaterializerSettings}
+import akka.stream.ActorMaterializer
 import com.mesosphere.mesos.client.MesosClient
 import com.mesosphere.mesos.conf.MesosClientSettings
 import com.mesosphere.usi.core.Scheduler
@@ -44,7 +44,7 @@ import scala.util.{Failure, Success}
   */
 class CoreHelloWorldFramework(conf: Config) extends StrictLogging {
   implicit val system = ActorSystem()
-  implicit val mat = ActorMaterializer(ActorMaterializerSettings(system).withDebugLogging(true))
+  implicit val mat = ActorMaterializer()
   implicit val ec = system.dispatcher
 
   val settings = MesosClientSettings(conf.getString("master-url"))
--- a/examples/core-hello-world/src/main/scala/com/mesosphere/usi/examples/CoreHelloWorldFramework.scala
+++ b/examples/core-hello-world/src/main/scala/com/mesosphere/usi/examples/CoreHelloWorldFramework.scala
@@ -4,13 +4,24 @@ import java.util.UUID
 
 import akka.NotUsed
 import akka.actor.ActorSystem
-import akka.stream.{ActorMaterializer, ActorMaterializerSettings}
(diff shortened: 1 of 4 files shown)


examples/core-hello-world/src/main/scala/com/mesosphere/usi/examples/CoreHelloWorldFramework.scala
```diff
@@ -4,31 +4,21 @@ import java.util.UUID
...
-  PodSpec,
-  PodStatus,
-  PodStatusUpdated,
-  RunSpec,
-  SpecUpdated,
-  SpecsSnapshot,
-  StateEvent,
-  StateSnapshot
-}
+import com.mesosphere.usi.core.models.{Goal, PodId, PodSpec, PodStatus, PodStatusUpdated, RunSpec, SpecUpdated, SpecsSnapshot, StateEvent, StateSnapshot}
```
alenkacz: :trollface:

examples/core-hello-world/src/main/scala/com/mesosphere/usi/examples/CoreHelloWorldFramework.scala
```diff
@@ -43,7 +33,7 @@ import scala.sys.SystemProperties
   */
 class CoreHelloWorldFramework(conf: Config) extends StrictLogging {
   implicit val system = ActorSystem()
-  implicit val mat = ActorMaterializer()
+  implicit val mat = ActorMaterializer(ActorMaterializerSettings(system).withDebugLogging(true))
```
alenkacz: is debug logging the way to go for production use?

examples/core-hello-world/src/main/scala/com/mesosphere/usi/examples/CoreHelloWorldFramework.scala
```diff
@@ -145,13 +135,11 @@ class CoreHelloWorldFramework(conf: Config) extends StrictLogging {
...
         logger.info(s"Task status updates for podId: $id: ${taskStatuses}")
-        e
 
       case e =>
-        logger.error(s"Unhandled event: $e") // we ignore everything else for now
-        e
+        logger.warn(s"Unhandled event: $e") // we ignore everything else for now
     }
-    .toMat(Sink.queue())(Keep.right)
+    .toMat(Sink.ignore)(Keep.right)
```
alenkacz, line 153: since this is a simple example, can we maybe comment on why Sink.ignore is the right think to do here. Because I guess that in every other real framework that probably would not be the case... ?

alenkacz (approved):
LGTM :)

meln1k (approved):

examples/core-hello-world/src/main/scala/com/mesosphere/usi/examples/CoreHelloWorldFramework.scala
zen-dog: good catch!

examples/core-hello-world/src/main/scala/com/mesosphere/usi/examples/CoreHelloWorldFramework.scala
zen-dog: No, it's the other way around - the queue was good for testing (to pull state events out of it) but it also backpressures so not suitable for production.

examples/core-hello-world/src/main/scala/com/mesosphere/usi/examples/CoreHelloWorldFramework.scala
zen-dog: I hate scalafmt

PR #71, opened 2019-03-14
Ease Java interop.

jeschkies:
Summary:
This relates to mesosphere/mesos-plugin#23 and should some Java interop by
* providing JUnit extensions for Zookeeper server and Mesos cluster
* including `scala-logging` as a transitive dependency


build.gradle
```diff
@@ -55,6 +54,9 @@ allprojects {
         // Logging
         testRuntime group: 'ch.qos.logback', name: 'logback-classic', version: '1.1.7'
 
+	// Transitive dependencies
+        compile "com.typesafe.scala-logging:scala-logging_$scalaVersion:3.9.2"
```
alenkacz, line 57: 🤔 what changed here? why did we go from implementation to compile? I think we should use only `compile` and `api` so not sure what we're trying to do here?

jeschkies: The `java-library` Gradle plugin has `implementation` and `api`. The `scala` Gradle plugin has `compile` and `implementation`. As the comment says I declared `scala-logging` as a transitive dependencies like Akka so that the consumer won't have to guess the Scala version etc.

zen-dog (approved):

build.gradle
timcharper: scala-logging is only used at compile time? I think there are some things we use at runtime also... such as the structured logging stuff

test-utils/src/main/scala/com/mesosphere/utils/zookeeper/ZookeeperServerExtension.scala
```diff
@@ -0,0 +1,23 @@
+package com.mesosphere.utils.zookeeper
+import org.junit.jupiter.api.extension.{AfterAllCallback, BeforeAllCallback, ExtensionContext}
+
+/**
+  * Provides a JUnit 5 extension for booting a Zookeeper server for tests.
+  *
+  * @see [[com.mesosphere.utils.mesos.MesosClusterExtension]] for an example.
+  */
+class ZookeeperServerExtension extends BeforeAllCallback with AfterAllCallback {
```
timcharper, line 9: interesting; so idea here is that we'll use the same zookeeper instance throughout the tests? (isolate each test by prefix or something?)

persistence-zookeeper/src/test/scala/com/mesosphere/usi/storage/zookeeper/ZooKeeperPersistenceStoreTest.scala
```diff
@@ -13,15 +13,12 @@ import com.mesosphere.utils.zookeeper.ZookeeperServerTest
...
 import org.apache.zookeeper.KeeperException.{NoNodeException, NodeExistsException}
-import org.junit.runner.RunWith
-import org.scalatest.junit.JUnitRunner
 
 import scala.collection.concurrent.TrieMap
 import scala.collection.immutable.Seq
 import scala.concurrent.{Await, ExecutionContext}
 import scala.util.{Failure, Random, Success}
 
-@RunWith(classOf[JUnitRunner])
```
timcharper, line 24: I thought this was required; do we no longer need to specify RunWith?

timcharper (approved):

kvish (approved):

persistence-zookeeper/src/test/scala/com/mesosphere/usi/storage/zookeeper/ZooKeeperPersistenceStoreTest.scala
jeschkies: Nope, @alenkacz changed it in https://github.com/mesosphere/usi/pull/47.

test-utils/src/main/scala/com/mesosphere/utils/zookeeper/ZookeeperServerExtension.scala
jeschkies: This will be per testsuite as it's already the case for the [Scalatest fixture](https://github.com/mesosphere/usi/blob/master/test-utils/src/main/scala/com/mesosphere/utils/zookeeper/ZookeeperServer.scala).

build.gradle
jeschkies: > scala-logging is only used at compile time?

The name is misleading. It is a compile, runtime and transitive dependency. That means Projects depending on USI will pull in `scala-logging` as well.

PR #72, opened 2019-03-14
HelloWorld framework: Don't swallow exceptions in Main

timcharper:
Summary:
The HelloWorld framework was not catching exceptions when attempting to get a MesosClient connection. This led to the appears that USI was swallowing exceptions, when in fact the exceptions were being returned, but they were being swallowed. Since the main thread is suspended until all actor systems are terminated, it gave the appearance that the USI stream was stalled without any progress.


zen-dog (approved):
🚢

meln1k (approved):

PR #73, opened 2019-03-14
Introduce `OfferMatcher` interface

zen-dog:
Summary:
we extract the offer matching portion of the `MesosEventsLogic` behind an `OfferMatcher` interface separating it from the rest of the logic e.g. building `TaskInfo`s. The only method of the interface receives a Mesos offer and a set of PodSpecs and returns a map of all matched PodSpecs with the list of corresponding resources. This will allow us to easier test different offer matching strategies in the future.

### DIFF
--- a/core/src/main/scala/com/mesosphere/usi/core/logic/MesosEventsLogic.scala
+++ b/core/src/main/scala/com/mesosphere/usi/core/logic/MesosEventsLogic.scala
@@ -1,90 +1,33 @@
 package com.mesosphere.usi.core.logic
 
-import com.mesosphere.ImplicitStrictLogging
-import com.mesosphere.LoggingArgs
 import java.time.Instant
 
+import com.mesosphere.{ImplicitStrictLogging, LoggingArgs}
 import com.mesosphere.mesos.client.MesosCalls
 import com.mesosphere.usi.core._
-import com.mesosphere.usi.core.matching.ResourceMatcher
+import com.mesosphere.usi.core.matching.{FCFSOfferMatcher, OfferMatcher}
 import com.mesosphere.usi.core.models._
-import com.mesosphere.usi.core.models.resources.{ResourceRequirement, ResourceType}
 import org.apache.mesos.v1.scheduler.Protos.{Call => MesosCall, Event => MesosEvent}
 import org.apache.mesos.v1.{Protos => Mesos}
+import SchedulerLogicHelpers._
 
-import scala.annotation.tailrec
 import scala.collection.JavaConverters._
 
 /**
   * The current home for USI Mesos event related logic
   */
-private[core] class MesosEventsLogic(mesosCallFactory: MesosCalls) extends ImplicitStrictLogging {
-  import SchedulerLogicHelpers._
-  private case class ResourceMatch(podSpec: PodSpec, resources: Seq[Mesos.Resource])
-  @tailrec private def maybeMatchPodSpec(
-      remainingResources: Map[ResourceType, Seq[Mesos.Resource]],
-      matchedResources: List[Mesos.Resource],
-      resourceRequirements: List[ResourceRequirement])
-    : Option[(List[Mesos.Resource], Map[ResourceType, Seq[Mesos.Resource]])] = {
-    resourceRequirements match {
-      case Nil =>
-        Some((matchedResources, remainingResources))
-      case req :: rest =>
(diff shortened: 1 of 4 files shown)


zen-dog:
No functional changes were made in this PR

core/src/main/scala/com/mesosphere/usi/core/matching/OfferMatcher.scala
```diff
@@ -0,0 +1,21 @@
+package com.mesosphere.usi.core.matching
+
+import com.mesosphere.usi.core.models.PodSpec
+import org.apache.mesos.v1.{Protos => Mesos}
+
+/**
+  * Interface for the offer matcher implementations. Given a Mesos [[Mesos.Offer]] and a set of [[PodSpec]]s
+  * return a map of all matched PodSpecs with the list of matched Mesos resources.
+  */
+trait OfferMatcher {
```
timcharper, line 10: Is this premature generality? Can we just implement OfferMatcher and then extract it to be more generic in the future?

zen-dog: I'd rather have a trait as a clear separation line. May be it's premature **now** but this part of the code tends to get entangles with the rest (especially when constraints are added)

core/src/main/scala/com/mesosphere/usi/core/matching/FCFSOfferMatcher.scala
```diff
@@ -0,0 +1,63 @@
...
+      remainingResources: Map[ResourceType, Seq[Mesos.Resource]],
+      matchedResources: List[Mesos.Resource],
+      resourceRequirements: List[ResourceRequirement])
+    : Option[(List[Mesos.Resource], Map[ResourceType, Seq[Mesos.Resource]])] = {
+
+    resourceRequirements match {
+      case Nil =>
+        Some((matchedResources, remainingResources))
+      case req :: rest =>
+        ResourceMatcher.matchAndConsume(req, remainingResources.getOrElse(req.resourceType, Nil)) match {
```
meln1k: What do you think about `.map` instead of `match` here?

meln1k (approved):

core/src/main/scala/com/mesosphere/usi/core/logic/MesosEventsLogic.scala
```diff
@@ -1,90 +1,33 @@
...
-
-  @tailrec private def matchPodSpecsTaskRecords(
-      offer: Mesos.Offer,
-      remainingResources: Map[ResourceType, Seq[Mesos.Resource]],
-      result: Map[PodId, List[Mesos.TaskInfo]],
-      pendingLaunchPodSpecs: List[PodSpec]): Map[PodId, List[Mesos.TaskInfo]] = {
+private[core] class MesosEventsLogic(mesosCallFactory: MesosCalls, offerMatcher: OfferMatcher = new FCFSOfferMatcher())
+    extends ImplicitStrictLogging {
 
+  private[core] def matchOffer(offer: Mesos.Offer, specs: Iterable[PodSpec]): (Set[PodId], SchedulerEventsBuilder) = {
```
alenkacz, line 22: can we find a better name for this method? since it's not just matching and now there's offer matcher interface doing the matching

core/src/main/scala/com/mesosphere/usi/core/logic/MesosEventsLogic.scala
```diff
@@ -1,90 +1,33 @@
...
-                resources = matchedResources
-              )
-            }(collection.breakOut)
-
-            matchPodSpecsTaskRecords(offer, newRemainingResources, result.updated(podSpec.id, taskInfos), rest)
-          case None =>
-            matchPodSpecsTaskRecords(offer, remainingResources, result, rest)
-        }
+    val matchedSpecs: Map[PodSpec, scala.List[Mesos.Resource]] = offerMatcher.matchOffer(offer, specs)
+    val taskInfos = matchedSpecs.map {
```
alenkacz, line 27: these two lines... 😍 beautiful

alenkacz (approved):
nice change! I think the code got cleared a lot, I like it.

now it would be nice to have some FCFSOfferMatcher tests but otherwise 👍

zen-dog:
> now it would be nice to have some FCFSOfferMatcher tests

I put these in the capable hands of the core-developers 😛

PR #75, opened 2019-03-19
Eager cancel the graph fanout

timcharper:
Prior to this change, the closing of one of the outputs would not affect the
other.

With this change, if the Mesos client closes, then the stream for the scheduler
state events will also be closed.

If the scheduler state events stream is closed, then the Mesos client is
closed.


zen-dog (approved):
🚢 Nice!

alenkacz (approved):
👍

PR #76, opened 2019-03-19
Overload RunSpec constructor to allow java list

kvish:
Adds an easy constructor for use from Java. Currently to create a runSpec need to add a list to seq converter in java code:

```
 private <T> Seq<T> convertListToSeq(List<T> inputList) {
    return JavaConverters.asScalaIteratorConverter(inputList.iterator()).asScala().toSeq();
  }
```

moving this logic into USI

### DIFF
--- a/core-models/src/main/scala/com/mesosphere/usi/core/models/RunSpec.scala
+++ b/core-models/src/main/scala/com/mesosphere/usi/core/models/RunSpec.scala
@@ -1,6 +1,7 @@
 package com.mesosphere.usi.core.models
 
 import com.mesosphere.usi.core.models.resources.ResourceRequirement
+import scala.collection.JavaConverters._
 
 /**
   * Specification used to launch a [[PodSpec]]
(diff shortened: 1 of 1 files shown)


core-models/src/main/scala/com/mesosphere/usi/core/models/RunSpec.scala
```diff
@@ -16,4 +17,11 @@ import com.mesosphere.usi.core.models.resources.ResourceRequirement
 case class RunSpec(
     resourceRequirements: Seq[ResourceRequirement],
     shellCommand: String,
-    fetch: Seq[FetchUri] = Seq.empty)
+    fetch: Seq[FetchUri] = Seq.empty) {
+
+  def this(
```
alenkacz, line 22: I'll repost my comment on slack:

shouldn't we do all the java APIs in a separate namespace and ideally write them in java? Then we know they are java friendly :slightly_smiling_face: Akka has two namespaces as well and the ones for java are written in ... wait for it... Java :slightly_smiling_face:

EDIT: on a second thought, I am fine with scala if we're sure it's java friendly, I would still separate the namespaces though...

alenkacz (changes requested):
I am requesting changes just so we can have the discussion around what way do we want to go with java apis. Ultimately it's a bit up to you guys but I would prefer to have them separated so java consumers are not concerned with java APIs

jeschkies:
What do you think about using the builder pattern?

kvish:
ya @jeschkies that makes sense, we can just have a builder here that takes both args

timcharper (changes requested):
Let's go the builder pattern for the Java interface

PR #77, opened 2019-03-19
Introduce validation through state updates

alenkacz:
This is another possibility how to approach invalid PodSpecs that Tim prefers. Basically there are several ways how PodSpec can be invalid:
- it can request resources that would never be offered thus it will never be scheduled
- it can contain commands or networks that does not exist and it will always fail
- it can contain a configuration that we know upfront is invalid and we don't even need to push that podspec to offer matching

For all these cases we decided to use the same channel of communicating these errors back to framework and that's the StateEvents. 

PS: for now this is just a very rough PR without tests and some polishing until we agree this is the way we want to go.

### DIFF
--- a/core-models/src/main/scala/com/mesosphere/usi/core/models/PodSpec.scala
+++ b/core-models/src/main/scala/com/mesosphere/usi/core/models/PodSpec.scala
@@ -18,6 +18,7 @@ case class PodSpec(id: PodId, goal: Goal, runSpec: RunSpec)
 
 object PodSpec {
   type ValidationMessage = String
+  val Valid = Seq.empty
 
   /**
     * Verifies that every value in range is requested only once. Requesting same value multiple times would yield pod that will never be scheduled.
@@ -33,7 +34,7 @@ object PodSpec {
     if (staticPorts.distinct.length != staticPorts.length) {
       Seq(s"Every value inside RangeResource can be requested only once. Requirement: ${staticPorts.mkString(",")}")
     } else {
-      Seq.empty
+      Valid
     }
   }
 
@@ -46,7 +47,7 @@ object PodSpec {
         .map(s => s"${s.resourceType}:${s.amount}")
         .mkString(",")}")
     } else {
-      Seq.empty
+      Valid
     }
   }
 
--- a/core-models/src/main/scala/com/mesosphere/usi/core/models/PodSpec.scala
+++ b/core-models/src/main/scala/com/mesosphere/usi/core/models/PodSpec.scala
@@ -1,5 +1,7 @@
 package com.mesosphere.usi.core.models
 
+import com.mesosphere.usi.core.models.resources.{ExactValue, RangeRequirement, ResourceRequirement, ScalarRequirement}
+
 /**
   * Framework implementation owned specification of some Pod that should be launched.
   *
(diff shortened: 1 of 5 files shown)


core-models/src/main/scala/com/mesosphere/usi/core/models/StateEvent.scala
```diff
@@ -27,6 +27,7 @@ object StateSnapshot {
   */
 sealed trait StateUpdated extends StateEvent
 
+case class PodInvalid(id: PodId, errors: Seq[String]) extends StateUpdated with PodStateEvent
```
timcharper: After thinking about this more... it really is a bug if the framework provides invalid pod specs. I don't think we should have a special status message for it. Can we just use podStatus? Expose the message the same way we would a regular failure.

alenkacz: I don't want to encode that in `case class PodStatus(id: PodId, taskStatuses: Map[TaskId, Mesos.TaskStatus])`. Maybe we can derive another type from PodStatus with extra fields and empty task statuses? WDYT, @timcharper

core-models/src/main/scala/com/mesosphere/usi/core/models/StateEvent.scala
```diff
@@ -27,6 +27,7 @@ object StateSnapshot {
   */
 sealed trait StateUpdated extends StateEvent
```
zen-dog, line 28: I don't quite understand why we need two sealed traits `StateUpdated` and `StateEvent`. What's the difference? /cc @timcharper

core-models/src/main/scala/com/mesosphere/usi/core/models/StateEvent.scala
zen-dog: I agree with Alena, having an empty `tasksStatuses` and an `errors: Seq[String]` there seems weird. I'd rather make the API expressive and readable. 
But maybe we want to a general type for errors? Not every error will be Pod related e.g. we might need to bubble up Mesos `message Error` event.

core/src/main/scala/com/mesosphere/usi/core/SchedulerLogicHandler.scala
```diff
@@ -85,13 +86,36 @@ private[core] class SchedulerLogicHandler(mesosCallFactory: MesosCalls) {
     */
   private var cachedPendingLaunch = CachedPendingLaunch(Set.empty)
 
+  def validateEvent(msg: SpecEvent): Seq[PodInvalid] = msg match {
+    case SpecsSnapshot(podSpecs, _) =>
+      podSpecs
+        .map(p => {
```
zen-dog: nit: you don't need the extra `{`, just make it `.map { p =>...`

core/src/main/scala/com/mesosphere/usi/core/SchedulerLogicHandler.scala
```diff
@@ -85,13 +86,36 @@ private[core] class SchedulerLogicHandler(mesosCallFactory: MesosCalls) {
...
+          PodSpec.isValid(p) match {
+            case Nil => None
+            case err => Some(PodInvalid(p.id, err))
+          }
+        })
+        .collect { case Some(p) => p }
```
zen-dog: just use:
```suggestion
        .flatten
```

core/src/main/scala/com/mesosphere/usi/core/SchedulerLogicHandler.scala
```diff
@@ -85,13 +86,36 @@ private[core] class SchedulerLogicHandler(mesosCallFactory: MesosCalls) {
...
+    case PodSpecUpdated(_, Some(podSpec)) =>
+      PodSpec.isValid(podSpec) match {
```
zen-dog: another neat trick is to use `toList` method:
```
@ Some(1).toList
res2: List[Int] = List(1)

@ None.toList
res3: List[Nothing] = List()
```
so simply:
```suggestion
      PodSpec.isValid(podSpec).toList
```

zen-dog (approved):
I left a few comments so PTAL. But, otherwise, I like the idea of handling invalid pods through state events.

core-models/src/main/scala/com/mesosphere/usi/core/models/PodSpec.scala
```diff
@@ -13,3 +15,45 @@ package com.mesosphere.usi.core.models
...
+    */
+  private def validateStaticRangeRequirementsUnique(runSpec: RunSpec): Seq[ValidationMessage] = {
+    val staticPorts = runSpec.resourceRequirements.collect {
+      case RangeRequirement(requestedValues, _, _) => requestedValues
+    }.flatten.collect { case ExactValue(value) => value }
+
+    if (staticPorts.distinct.length != staticPorts.length) {
+      Seq(s"Every value inside RangeResource can be requested only once. Requirement: ${staticPorts.mkString(",")}")
+    } else {
+      Seq.empty
```
jeschkies: What do you think about `val Valid = Seq.empty`? So that we have 
```suggestion
      Valid
```

core-models/src/main/scala/com/mesosphere/usi/core/models/StateEvent.scala
jeschkies: Oh man. This would be super hard to deal with from the framework perspective. Please help me understand. Is there any validation that we cannot do during the `PodSpecUpdated` construction?

jeschkies (changes requested):
Unless there is a good reason we cannot validate a `PodSpecUpdated` during its construction I'm opposed to this API.

core-models/src/main/scala/com/mesosphere/usi/core/models/PodSpec.scala
jeschkies: push

core-models/src/main/scala/com/mesosphere/usi/core/models/StateEvent.scala
jeschkies: Why is it a `StateUpdated`? Should it not extend a `ErrorEvent` or something similar?

core-models/src/main/scala/com/mesosphere/usi/core/models/PodSpec.scala
alenkacz: 👍

core-models/src/main/scala/com/mesosphere/usi/core/models/StateEvent.scala
alenkacz: `StateUpdated` seems redundant to me at this point. @timcharper ?

core/src/main/scala/com/mesosphere/usi/core/SchedulerLogicHandler.scala
alenkacz: nice! And in the end, all this can be a flatMap :)

core-models/src/main/scala/com/mesosphere/usi/core/models/StateEvent.scala
jeschkies: Ping @alenkacz

core-models/src/main/scala/com/mesosphere/usi/core/models/StateEvent.scala
```diff
@@ -27,6 +27,12 @@ object StateSnapshot {
...
 
+/**
+  * Captures a pod event that was caused by user of USI core submitting an invalid PodSpec.
```
jeschkies:
```suggestion
  * Captures a pod event that was caused by a [[SpecEvent]] submitted by the user.
```

jeschkies (approved):

PR #78, opened 2019-03-26
Accept spec snapshot as input.

jeschkies:
Blocks mesosphere/mesos-plugin#26. Please see it for an example use case.

### DIFF
--- a/core/src/main/scala/com/mesosphere/usi/core/javaapi/Scheduler.scala
+++ b/core/src/main/scala/com/mesosphere/usi/core/javaapi/Scheduler.scala
@@ -1,0 +1,33 @@
+package com.mesosphere.usi.core.javaapi
+
+import akka.NotUsed
+import akka.stream.javadsl
+import com.mesosphere.mesos.client.{MesosCalls, MesosClient}
+import com.mesosphere.usi.core.{Scheduler => ScalaScheduler}
+import com.mesosphere.usi.core.models.{SpecUpdated, SpecsSnapshot, StateEvent, StateSnapshot}
+import org.apache.mesos.v1.scheduler.Protos.{Call => MesosCall, Event => MesosEvent}
+
+object Scheduler {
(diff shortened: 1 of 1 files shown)


core/src/main/scala/com/mesosphere/usi/core/Scheduler.scala
```diff
@@ -47,26 +47,26 @@ import scala.collection.JavaConverters._
  *                                     +----------------------+
  */
 object Scheduler {
-  type SpecInput = (SpecsSnapshot, Source[SpecUpdated, Any])
 
   type StateOutput = (StateSnapshot, Source[StateEvent, Any])
 
-  def fromClient(client: MesosClient): Flow[SpecInput, StateOutput, NotUsed] = {
+  def fromClient(client: MesosClient, specsSnapshot: SpecsSnapshot): Flow[SpecUpdated, StateOutput, NotUsed] = {
```
jeschkies: This is way simpler to use and should have the same effect of keeping the ordering.

jeschkies (commented):
I would also like the factory methods to return a `Future[Tuple2[StateSnapeshot, Flow[SpecUpdated, StateEvent]]]`. Take a look at https://github.com/mesosphere/mesos-plugin/pull/26/files#diff-b917078c299c2eeaf0758f34498e6fc7R86. How would one reconcile the tasks after a fail over and maybe queue new spec updates? This would be easier with

```java
var pair = Scheduler.fromClient(client, SpecsSnapshot.empty());
var schedulerFlow = pair._2;
var initialStates = pair._1;
this.updates = Source.<SpecUpdated>queue(256, OverflowStrategy.fail())
            .via(schedulerFlow)
            .toMat(Sink.foreach(this::updateState), Keep.left())
            .run(materializer);

reconcile(initialStates);
```

`reconcile` can then safely call `updates.offer`.

PR #80, opened 2019-03-29
Use InMemoryPersistenceStorage for persisting pod records

takirala:
Add a persistence storage interface and create a flow from it to use in the Scheduler graph. This PR adds the following changes

- Create a generic RecordRepository trait. Currently, only `PodRecordRepository` is implemented.
- Handle the initial snapshot loading from the `PodRecordRepository`.
- Create a scheduler graph component that uses the `PodRecordRepository` to persist the `SchedulerEvents`. This component itself performs materialization for every `SchedulerEvents` element.

#### Testing
- Add a `RepositoryBehavior` that can be used to verify all the implementors of `RecordRepository`. Currently used only for InMemory storage but can/will be extended to other implementations (such as zk).
- Add the akka-stream-testkit library dependency to `test-utils` module so that it can be used in other modules.
- Add a unit test for the persistence flow component.
- Add an integration test to verify the atmost once launch guarantee.

@jeschkies made the initial PR #51 that laid some groundwork for this.

JIRA : DCOS-49274

### DIFF
--- a/core/src/main/scala/com/mesosphere/usi/core/CallerThreadExecutionContext.scala
+++ b/core/src/main/scala/com/mesosphere/usi/core/CallerThreadExecutionContext.scala
@@ -4,6 +4,7 @@ import java.util.concurrent.Executor
 import scala.concurrent.ExecutionContext
 import scala.concurrent.ExecutionContextExecutor
 
+// TODO : This is a duplicated in com.mesosphere.usi.async. Ideally, this needs to be in a shared module.
 private[usi] object CallerThreadExecutionContext {
   val executor: Executor = (command: Runnable) => command.run()
 
--- a/core/src/main/scala/com/mesosphere/usi/core/CallerThreadExecutionContext.scala
+++ b/core/src/main/scala/com/mesosphere/usi/core/CallerThreadExecutionContext.scala
@@ -1,0 +1,11 @@
+package com.mesosphere.usi.core
+
+import java.util.concurrent.Executor
+import scala.concurrent.ExecutionContext
+import scala.concurrent.ExecutionContextExecutor
+
+private[usi] object CallerThreadExecutionContext {
+  val executor: Executor = (command: Runnable) => command.run()
+
+  implicit val context: ExecutionContextExecutor = ExecutionContext.fromExecutor(executor)
+}
--- a/core/src/main/scala/com/mesosphere/usi/core/Scheduler.scala
+++ b/core/src/main/scala/com/mesosphere/usi/core/Scheduler.scala
@@ -66,8 +66,7 @@ object Scheduler {
 
   val PipeliningLimit = ConfigFactory.load().getInt("persistence.pipeline-limit")
 
-  def fromSnapshot(specsSnapshot: SpecsSnapshot, client: MesosClient, podRecordRepository: PodRecordRepository)(
-      implicit materializer: Materializer): Flow[SpecUpdated, StateOutput, NotUsed] =
+  def fromSnapshot(specsSnapshot: SpecsSnapshot, client: MesosClient, podRecordRepository: PodRecordRepository): Flow[SpecUpdated, StateOutput, NotUsed] =
     Flow[SpecUpdated]
       .prefixAndTail(0)
       .map { case (_, rest) => specsSnapshot -> rest }
@@ -204,8 +203,7 @@ object Scheduler {
 
   }
 
-  def fromClient(client: MesosClient, podRecordRepository: PodRecordRepository)(
-      implicit materializer: Materializer): Flow[SpecInput, StateOutput, NotUsed] = {
+  def fromClient(client: MesosClient, podRecordRepository: PodRecordRepository): Flow[SpecInput, StateOutput, NotUsed] = {
     if (!isMultiRoleFramework(client.frameworkInfo)) {
       throw new IllegalArgumentException(
         "USI scheduler provides support for MULTI_ROLE frameworks only. " +
@@ -217,8 +215,7 @@ object Scheduler {
   def fromFlow(
       mesosCallFactory: MesosCalls,
       podRecordRepository: PodRecordRepository,
-      mesosFlow: Flow[MesosCall, MesosEvent, Any])(
-      implicit materializer: Materializer): Flow[SpecInput, StateOutput, NotUsed] = {
+      mesosFlow: Flow[MesosCall, MesosEvent, Any]): Flow[SpecInput, StateOutput, NotUsed] = {
     Flow.fromGraph {
       GraphDSL.create(unconnectedGraph(mesosCallFactory, podRecordRepository), mesosFlow)((_, _) => NotUsed) {
         implicit builder =>
@@ -260,8 +257,7 @@ object Scheduler {
       (stateSnapshot, stateUpdates)
   }
 
-  private[core] def unconnectedGraph(mesosCallFactory: MesosCalls, podRecordRepository: PodRecordRepository)(
-      implicit materializer: Materializer): BidiFlow[SpecInput, StateOutput, MesosEvent, MesosCall, NotUsed] = {
+  private[core] def unconnectedGraph(mesosCallFactory: MesosCalls, podRecordRepository: PodRecordRepository): BidiFlow[SpecInput, StateOutput, MesosEvent, MesosCall, NotUsed] = {
     val schedulerLogicGraph = new SchedulerLogicGraph(mesosCallFactory, podRecordRepository.readAll())
     BidiFlow.fromGraph {
       GraphDSL.create(schedulerLogicGraph) { implicit builder => (schedulerLogic) =>
@@ -291,24 +287,25 @@ object Scheduler {
     }
   }
 
-  private[core] def persistenceFlow(podRecordRepository: PodRecordRepository)(
-      implicit materializer: Materializer): Flow[SchedulerEvents, SchedulerEvents, NotUsed] = {
-    Flow[SchedulerEvents].mapAsync(1) { events =>
-      persistEvents(events, podRecordRepository).map(_ => events)(CallerThreadExecutionContext.context)
-    }
-  }
+  private[core] def persistenceFlow(podRecordRepository: PodRecordRepository): Flow[SchedulerEvents, SchedulerEvents, NotUsed] = {
 
-  private def persistEvents(events: SchedulerEvents, podRecordRepository: PodRecordRepository)(
-      implicit materializer: Materializer): Future[Done] = {
-    Source(events.stateEvents).collect {
-      case PodRecordUpdated(_, Some(podRecord)) =>
-        () =>
-          podRecordRepository.store(podRecord)
-      case PodRecordUpdated(podId, None) =>
-        () =>
-          podRecordRepository.delete(podId)
-    }.mapAsync(PipeliningLimit)(call => call())
-      .runWith(Sink.ignore)
+    Flow[SchedulerEvents].mapConcat { events =>
+      val ops: Vector[Future[Option[SchedulerEvents]]] = events.stateEvents.collect {
+        case PodRecordUpdated(_, Some(podRecord)) =>
+          podRecordRepository
+            .store(podRecord)
+            .map { _ =>
+              None
+            }(CallerThreadExecutionContext.context)
+        case PodRecordUpdated(podId, None) =>
+          podRecordRepository
+            .delete(podId)
+            .map { _ =>
+              None
+            }(CallerThreadExecutionContext.context)
+      }(collection.breakOut)
+      ops :+ Future.successful(Some(events))
+    }.mapAsync(PipeliningLimit)(identity).collect { case Some(events) => events }
   }
 
   private def isMultiRoleFramework(frameworkInfo: FrameworkInfo): Boolean =
--- a/core/src/main/scala/com/mesosphere/usi/core/Scheduler.scala
+++ b/core/src/main/scala/com/mesosphere/usi/core/Scheduler.scala
@@ -1,6 +1,5 @@
 package com.mesosphere.usi.core
 
-import akka.Done
 import akka.NotUsed
 import akka.stream.{BidiShape, FlowShape, Materializer}
 import akka.stream.scaladsl.{BidiFlow, Broadcast, Flow, GraphDSL, Sink, Source}
@@ -17,7 +16,6 @@ import com.mesosphere.usi.repository.PodRecordRepository
 import org.apache.mesos.v1.Protos.FrameworkInfo
 import org.apache.mesos.v1.scheduler.Protos.{Call => MesosCall, Event => MesosEvent}
 import scala.collection.JavaConverters._
-import scala.concurrent.Future
 
 /**
   * Provides the scheduler graph component. The component has two inputs, and two outputs:
@@ -132,10 +130,9 @@ object Scheduler {
       mesosCallFactory: MesosCalls,
       podRecordRepository: PodRecordRepository
   )(implicit materializer: Materializer): BidiFlow[SpecInput, StateOutput, MesosEvent, MesosCall, NotUsed] = {
+    val schedulerLogicGraph = new SchedulerLogicGraph(mesosCallFactory, podRecordRepository.readAll())
     BidiFlow.fromGraph {
-      GraphDSL.create(
-        new SchedulerLogicGraph(mesosCallFactory, podRecordRepository.readAll().runWith(Sink.head)),
-      ) { implicit builder => (schedulerLogic) =>
+      GraphDSL.create(schedulerLogicGraph) { implicit builder => (schedulerLogic) =>
         {
           import GraphDSL.Implicits._
 
@@ -162,32 +159,18 @@ object Scheduler {
     }
   }
 
+  val PipeliningLimit = 128
+
   private[core] def persistenceFlow(
       podRecordRepository: PodRecordRepository
   )(implicit materializer: Materializer): Flow[SchedulerEvents, SchedulerEvents, NotUsed] = {
     Flow[SchedulerEvents].mapAsync(1) { events =>
-      val (storeRecords, deleteRecords) = events.stateEvents.collect { case x: PodRecordUpdated => x }
-        .partition(_.newRecord.isDefined)
-      if (storeRecords.isEmpty && deleteRecords.isEmpty) {
-        Future.successful(events)
-      } else {
-        val storeResult = if (storeRecords.nonEmpty) {
-          Source(storeRecords).collect { case PodRecordUpdated(_, Some(record)) => record }
-            .via(podRecordRepository.storeFlow)
-            .grouped(storeRecords.size)
-            .map(_ => Done)
-        } else Source.single(Done)
-        val deleteResult = if (deleteRecords.nonEmpty) {
-          Source(deleteRecords)
-            .map(_.id)
-            .via(podRecordRepository.deleteFlow)
-            .grouped(deleteRecords.size)
-            .map(_ => Done)
-        } else Source.single(Done)
-        Source
-          .zipWithN[Done, SchedulerEvents](_ => events)(List(storeResult, deleteResult))
-          .runWith(Sink.head) // This materialization is cheap compared to the IO operation.
-      }
+      Source(events.stateEvents).collect {
+        case PodRecordUpdated(_, Some(podRecord)) => podRecordRepository.store(podRecord)
+        case PodRecordUpdated(podId, None) => podRecordRepository.delete(podId)
+      }.mapAsync(PipeliningLimit)(identity)
+        .runWith(Sink.ignore)
+        .map(_ => events)(CallerThreadExecutionContext.context)
     }
   }
 
--- a/core/src/main/scala/com/mesosphere/usi/core/Scheduler.scala
+++ b/core/src/main/scala/com/mesosphere/usi/core/Scheduler.scala
@@ -17,7 +17,6 @@ import com.typesafe.config.ConfigFactory
 import org.apache.mesos.v1.Protos.FrameworkInfo
 import org.apache.mesos.v1.scheduler.Protos.{Call => MesosCall, Event => MesosEvent}
 import scala.collection.JavaConverters._
-import scala.collection.immutable
 import scala.concurrent.{Future, Promise}
 import scala.util.{Failure, Success}
 
@@ -297,17 +296,24 @@ object Scheduler {
 
   private[core] def persistenceFlow(
       podRecordRepository: PodRecordRepository): Flow[SchedulerEvents, SchedulerEvents, NotUsed] = {
-    Flow[SchedulerEvents].mapConcat { events =>
-      val ops: List[() => Future[Option[SchedulerEvents]]] = events.stateEvents.collect {
-        case PodRecordUpdated(_, Some(podRecord)) =>
-          () =>
-            podRecordRepository.store(podRecord).map(_ => None)(CallerThreadExecutionContext.context)
-        case PodRecordUpdated(podId, None) =>
-          () =>
-            podRecordRepository.delete(podId).map(_ => None)(CallerThreadExecutionContext.context)
-      }
-      (() => Future.successful(Some(events))) +: ops
-    }.mapAsync(PipeliningLimit)(call => call()).collect { case Some(events) => events }
+    Flow[SchedulerEvents]
+      .mapConcat(persistEvents(_, podRecordRepository))
+      .mapAsync(PipeliningLimit)(call => call())
+      .collect { case Some(events) => events }
+  }
+
+  private def persistEvents(
+      events: SchedulerEvents,
+      podRecordRepository: PodRecordRepository): List[() => Future[Option[SchedulerEvents]]] = {
+    val ops: List[() => Future[Option[SchedulerEvents]]] = events.stateEvents.collect {
+      case PodRecordUpdated(_, Some(podRecord)) =>
+        () =>
+          podRecordRepository.store(podRecord).map(_ => None)(CallerThreadExecutionContext.context)
+      case PodRecordUpdated(podId, None) =>
+        () =>
+          podRecordRepository.delete(podId).map(_ => None)(CallerThreadExecutionContext.context)
+    }
+    ops :+ (() => Future.successful(Some(events)))
   }
 
   private def isMultiRoleFramework(frameworkInfo: FrameworkInfo): Boolean =
--- a/core/src/main/scala/com/mesosphere/usi/core/Scheduler.scala
+++ b/core/src/main/scala/com/mesosphere/usi/core/Scheduler.scala
@@ -4,6 +4,7 @@ import akka.{Done, NotUsed}
 import akka.stream.scaladsl.{BidiFlow, Broadcast, Flow, GraphDSL, Sink, SinkQueueWithCancel, Source}
 import akka.stream.{BidiShape, FlowShape, KillSwitches, Materializer, OverflowStrategy, QueueOfferResult}
 import com.mesosphere.mesos.client.{MesosCalls, MesosClient}
+import com.mesosphere.usi.core.conf.SchedulerSettings
 import com.mesosphere.usi.core.models.{
   PodRecordUpdated,
   SpecEvent,
@@ -64,16 +65,7 @@ object Scheduler {
 
   type StateOutput = (StateSnapshot, Source[StateEvent, Any])
 
-  val PipeliningLimit = ConfigFactory.load().getInt("persistence.pipeline-limit")
-
-  def fromSnapshot(
-      specsSnapshot: SpecsSnapshot,
-      client: MesosClient,
-      podRecordRepository: PodRecordRepository): Flow[SpecUpdated, StateOutput, NotUsed] =
-    Flow[SpecUpdated]
-      .prefixAndTail(0)
-      .map { case (_, rest) => specsSnapshot -> rest }
-      .via(fromClient(client, podRecordRepository))
+  private val schedulerSettings = SchedulerSettings.fromConfig(ConfigFactory.load().getConfig("scheduler"))
 
   def asFlow(specsSnapshot: SpecsSnapshot, client: MesosClient, podRecordRepository: PodRecordRepository)(
       implicit materializer: Materializer): Future[(StateSnapshot, Flow[SpecUpdated, StateEvent, NotUsed])] = {
@@ -85,7 +77,6 @@ object Scheduler {
     snap.map { snapshot =>
       (snapshot, Flow.fromSinkAndSourceCoupled(sink, source))
     }
-
   }
 
   /**
@@ -265,7 +256,7 @@ object Scheduler {
   private[core] def unconnectedGraph(
       mesosCallFactory: MesosCalls,
       podRecordRepository: PodRecordRepository): BidiFlow[SpecInput, StateOutput, MesosEvent, MesosCall, NotUsed] = {
-    val schedulerLogicGraph = new SchedulerLogicGraph(mesosCallFactory, podRecordRepository.readAll())
+    val schedulerLogicGraph = new SchedulerLogicGraph(mesosCallFactory, () => podRecordRepository.readAll())
     BidiFlow.fromGraph {
       GraphDSL.create(schedulerLogicGraph) { implicit builder => (schedulerLogic) =>
         {
@@ -298,7 +289,7 @@ object Scheduler {
       podRecordRepository: PodRecordRepository): Flow[SchedulerEvents, SchedulerEvents, NotUsed] = {
     Flow[SchedulerEvents]
       .mapConcat(persistEvents(_, podRecordRepository))
-      .mapAsync(PipeliningLimit)(call => call())
+      .mapAsync(schedulerSettings.persistencePipelineLimit)(call => call())
       .collect { case Some(events) => events }
   }
 
--- a/core/src/main/scala/com/mesosphere/usi/core/Scheduler.scala
+++ b/core/src/main/scala/com/mesosphere/usi/core/Scheduler.scala
@@ -1,9 +1,8 @@
 package com.mesosphere.usi.core
 
 import akka.NotUsed
-import akka.stream.Attributes
-import akka.stream.{BidiShape, FlowShape, Inlet, Outlet}
-import akka.stream.stage.{GraphStage, GraphStageLogic, InHandler, OutHandler}
+import akka.stream.{Attributes, BidiShape, FlowShape, Inlet, Outlet}
+import akka.stream.stage.{AsyncCallback, GraphStage, GraphStageLogic, InHandler, OutHandler}
 import akka.stream.scaladsl.{BidiFlow, Broadcast, Flow, GraphDSL, Sink, Source}
 import com.mesosphere.mesos.client.{MesosCalls, MesosClient}
 import com.mesosphere.usi.core.models.{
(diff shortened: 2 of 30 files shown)


core/src/main/scala/com/mesosphere/usi/core/Scheduler.scala
```diff
@@ -130,4 +147,25 @@ object Scheduler {
       }
     }
   }
+
+  private def persistenceFlow(
+      podRecordRepository: PodRecordRepository): Flow[SchedulerEvents, SchedulerEvents, NotUsed] =
+    Flow[SchedulerEvents].mapAsync(1) { events =>
+      import CallerThreadExecutionContext.context
```
jeschkies: I don't quite understand the context. Why was that introduced?

core/src/main/scala/com/mesosphere/usi/core/Scheduler.scala
```diff
@@ -130,4 +147,25 @@ object Scheduler {
       }
     }
   }
+
+  private def persistenceFlow(
```
jeschkies: I would have assumed that the repository would maybe provide this interface.

core/src/main/scala/com/mesosphere/usi/core/Scheduler.scala
```diff
@@ -130,4 +147,25 @@ object Scheduler {
...
+      podRecordRepository: PodRecordRepository): Flow[SchedulerEvents, SchedulerEvents, NotUsed] =
+    Flow[SchedulerEvents].mapAsync(1) { events =>
+      import CallerThreadExecutionContext.context
+      Future
+        .sequence(events.stateEvents.map {
```
jeschkies: Why not use a stream here as well?

core/src/main/scala/com/mesosphere/usi/core/SchedulerLogicGraph.scala
```diff
@@ -82,12 +86,17 @@ class SchedulerLogicGraph(mesosCallFactory: MesosCalls)
...
+      val startGraph = this.getAsyncCallback[Try[Map[PodId, PodRecord]]] {
+        case Success(initialSnapshot) =>
+          handler.loadInitialPodRecords(initialSnapshot)
+          maybePull()
+        case Failure(ex) =>
+          this.failStage(ex)
       }
 
+      override def preStart(): Unit =
+        initialPodRecords.onComplete(startGraph.invoke)(CallerThreadExecutionContext.context)
```
jeschkies: I is kind of odd. For user we ask them to deal with `Source[(Snapshot, Source[StateEvent])]` but internally we are not using that. Why are we not passing the state snapshot via a stream? Are we not loosing backpressure here? Especially recovering state was slowing down Marathon in the past so I would assume that we leverage backpressure just here so that the USI scheduler flow is not overloaded.

persistence/src/main/scala/com/mesosphere/usi/repository/RecordRepository.scala
```diff
@@ -14,33 +14,23 @@ trait RecordRepository {
...
-    * Updates the record if it exists. If the record is missing,
-    * the future will be failed with an [[RecordNotFoundException]].
-    *
-    * @param record
-    * @return id of the updated record
+    * Retrieves all the existing records (if any).
+    * @return Map containing all the current pod records. Can be empty if there are no pod records.
     */
-  def update(record: Record): Future[RecordId]
+  def readAll(): Future[Map[RecordId, Record]]
```
jeschkies: Should we provide a stream instead to leverage backpressure?

persistence/src/main/scala/com/mesosphere/usi/repository/RecordRepository.scala
```diff
@@ -14,33 +14,23 @@ trait RecordRepository {
...
 
   /**
     * Deletes the record if it exists. If the record is missing,
     * * the future will be failed with an [[RecordNotFoundException]].
```
jeschkies: I think we want to be idempotent and *not* fail if the record does not exist.

jeschkies (dismissed):

core/src/main/scala/com/mesosphere/usi/core/Scheduler.scala
takirala: I don't understand how the repository can provide this interface. `core` module should control how the persistence storage is read/written to right ?

core/src/main/scala/com/mesosphere/usi/core/Scheduler.scala
takirala: This was introduced to perform cheap operations such as `.map(_ => events)` without switching the context. If we don't do future transformations in scope, we don't need this.

core/src/main/scala/com/mesosphere/usi/core/Scheduler.scala
takirala: In the current flow of things:

```
[SchedulerEvents] ------>        [Persistence]    ----------------> [StateOutput, MesosCalls]
emitted by 				 process ALL the StateEvents
Scheduler logic			 in single SchedulerEvents
						 object and THEN emits the 
						 object downstream.
						 backpressures the upstream 
						 until ALL the StateEvents
						 in SchedulerEvents are processed.
						 We can use a stream here but i 
					     am not sure what advantage it would
					     add. We are using `.mapAsync` with 				
						 parallelism 1 to process each
						 SchedulerEvents object - one at a time
						 with backpressure.
```

Currently the persistence layer emits the **exact** `SchedulerEvents` object that it receives, are you perhaps inferring that we don't need to do this and emit a "split-down" SchedulerEvents based on the stream processing speed ?

core/src/main/scala/com/mesosphere/usi/core/SchedulerLogicGraph.scala
takirala: At least in the scope of this PR, I am not trying to handle adhoc-snapshots. Assumption in this PR is that we receive the snapshot only once during bootstrap (or crash recovery) and we intentionally block all inlets until we can load the snapshot and be ready for the processing. I initially had an approach using FanIn3Shape but decided this is simpler and cleaner to get the initial snapshot loading.

persistence/src/main/scala/com/mesosphere/usi/repository/RecordRepository.scala
takirala: my bad, this description is outdated and the actual code does not reflect this, i will clean this up and get rid of unused `RecordNotFoundException`

persistence/src/main/scala/com/mesosphere/usi/repository/RecordRepository.scala
takirala: Current structure assumes we are intentionally blocking inlets until the state can be loaded. Based on the outcome of that structure, we can chose to change this to be a stream instead.

core/src/main/scala/com/mesosphere/usi/core/Scheduler.scala
jeschkies: In some place in Marathon we used Akka streams instead of `Future.sequence` to avoid too many futures. Also, there is not backpressure for loading.

core/src/main/scala/com/mesosphere/usi/core/SchedulerLogicGraph.scala
jeschkies: I think @timcharper has to explain the interface then once more to me. We do not expect adhoc snapshots but it is odd that our internal storage uses plain futures when we prescribe the nested sources to users. Why would we not use streams here as well?

persistence/src/main/scala/com/mesosphere/usi/repository/RecordRepository.scala
jeschkies: Where are we blocking? I must be missing something?

core/src/main/scala/com/mesosphere/usi/core/Scheduler.scala
jeschkies: So, I think @zen-dog can help. I thought we had something like a `Flow[StoreOp, Try]` So that one could map the updates.

```
val persitenceFlow: Flow[SchedulerEvents, SchedulerEvents] = {
  Flow[SchedulerEvents]
    .flatMapConcat(_.stateEvents)
    .map {
       case ... => UpdateOp | DeleteOp
    }
    .via(repository.flow)
    ...
}
```

This might be a little to complicated though it also omits then mapping back to `SchedulerEvents`.

core/src/main/scala/com/mesosphere/usi/core/Scheduler.scala
timcharper: It had to be called something, can't make the object the execution context.

core/src/main/scala/com/mesosphere/usi/core/Scheduler.scala
```diff
@@ -130,4 +151,81 @@ object Scheduler {
...
+                    case x: PodRecordUpdated => x
+                  }.partition(_.newRecord.isDefined)
+                  eventCounter.addAndGet(storeRecords.size + deleteRecords.size)
+                  nextElement = Some(schedulerEvents)
+                  if (eventCounter.get() == 0) {
+                    maybePushToPull()
+                  } else {
+                    // TODO : batch transactions
+                    Source(storeRecords)
+                      .map(_.newRecord.get)
```
jeschkies: Is `newRecord` an `Option`? If so I'd use `collect`:
```suggestion
                      .collect { case Some(newRecord) => newRecord }
```

core/src/main/scala/com/mesosphere/usi/core/CallerThreadExecutionContext.scala
```diff
@@ -0,0 +1,11 @@
+package com.mesosphere.usi.core
+
+import java.util.concurrent.Executor
+import scala.concurrent.ExecutionContext
+import scala.concurrent.ExecutionContextExecutor
+
+private[usi] object CallerThreadExecutionContext {
```
zen-dog, line 8: We already have one `com.mesosphere.usi.async.CallerThreadExecutionContext`

core/src/main/scala/com/mesosphere/usi/core/Scheduler.scala
```diff
@@ -130,4 +151,81 @@ object Scheduler {
...
+                      .via(podRecordRepository.storeFlow)
+                      .runWith(Sink.foreach(_ => {
+                        if (eventCounter.decrementAndGet() == 0) maybePushToPull()
```
meln1k: This is not thread-safe. See https://doc.akka.io/docs/akka/current/stream/stream-customize.html#thread-safety-of-custom-operators.

meln1k (dismissed):

core/src/main/scala/com/mesosphere/usi/core/Scheduler.scala
zen-dog: `maybePushToPull` happens on another thread so you should not call it like this

core/src/main/scala/com/mesosphere/usi/core/CallerThreadExecutionContext.scala
takirala: Good catch, although what do you suggest i do here ? The class `com.mesosphere.usi.async.CallerThreadExecutionContext` is in a different (`metrics-dropwizard`) module. I can't fathom a reasonable common module both the `core` and `metrics-dropwizard` have to share this same thread exec context.

core/src/main/scala/com/mesosphere/usi/core/Scheduler.scala
```diff
@@ -130,4 +149,25 @@ object Scheduler {
       }
     }
   }
+
+  private def persistenceFlow(
+      podRecordRepository: PodRecordRepository): Flow[SchedulerEvents, SchedulerEvents, NotUsed] =
+    Flow[SchedulerEvents].mapAsync(1) { events =>
+      import CallerThreadExecutionContext.context
+      Future
+        .sequence(events.stateEvents.map {
```
zen-dog: I would go for something like:
```scala
mapAsync(1) { events => 
val (updated, deleted) = events.stateEvents.partition(....)
val updatedF: Future[Done] = Source(updated)
 .via(podRepository.storeFlow)
 .runWith(Sink.ignore)

val deletedF: Future[Done] = Source(deleted)
 .via(podRepository.deleteFlow)
 .runWith(Sink.ignore)

updated.andThen(deleted) 
}
```
It has the downside of 2x materialization but that might be ok for now.

core/src/test/resources/application.conf
```diff
@@ -0,0 +1,16 @@
...
+    materializer {
+      debug {
+        # Enables the fuzzing mode which increases the chance of race conditions
+        # by aggressively reordering events and making certain operations more
+        # concurrent than usual.
+        # This setting is for testing purposes, NEVER enable this in a production
+        # environment!
+        # To get the best results, try combining this setting with a throughput
+        # of 1 on the corresponding dispatchers.
+        fuzzing-mode = on
```
takirala, line 12: I found this useful when I was playing around with a custom GraphStage approach I had at some point and I think retaining this would be useful and makes testing more aggressive, i can revert this if there is sth i missed.

core/src/main/scala/com/mesosphere/usi/core/CallerThreadExecutionContext.scala
timcharper: I'd say just leave two copies for now, with a note that there's a duplicate in the other module

zen-dog: > I can't fathom a reasonable common module both the `core` and `metrics-dropwizard` have to share this same thread exec context.

yeah, we'll need something like `utils` package in the future. this duplication is ok for now

core/src/main/scala/com/mesosphere/usi/core/SchedulerLogicGraph.scala
timcharper: In the future, we will not be sending a podSpec snapshot. This is going away.

core/src/test/scala/com/mesosphere/usi/core/SchedulerTest.scala
```diff
@@ -101,4 +105,55 @@ class SchedulerTest extends AkkaUnitTest with Inside {
     output.cancel()
     mesosCompleted.futureValue shouldBe Done
   }
+
+  "Persistence flow pipelines writes" in {
+    Given("a slow persistence storage and a flow using it")
+    val delayPerElement = 50.millis // Delay precision is 10ms
```
timcharper: Let's please avoid using fixed delays in tests.

examples/core-hello-world/src/test/scala/com/mesosphere/usi/examples/CoreHelloWorldFrameworkTest.scala
```diff
@@ -31,6 +47,68 @@ class CoreHelloWorldFrameworkTest extends AkkaUnitTest with MesosClusterTest wit
...
+    val (_, sub2) = runGraphAndFeedSnapshot(newScheduler, specsSnapshot)
+    val e4 = sub2.requestNext()
+    e4 shouldBe a[StateSnapshot]
+    e4.asInstanceOf[StateSnapshot].podRecords should contain theSameElementsAs podRecords
+
+    And("no new pod records have been created")
+    val newPodRecords = customPersistenceStore.readAll().futureValue.values
+    newPodRecords should contain theSameElementsAs podRecords
+
+    And(s"no further elements should be emitted")
```
timcharper: I think the linter is probably going to complain about the extra erroneous "s"

examples/core-hello-world/src/test/scala/com/mesosphere/usi/examples/CoreHelloWorldFrameworkTest.scala
```diff
@@ -31,6 +47,68 @@ class CoreHelloWorldFrameworkTest extends AkkaUnitTest with MesosClusterTest wit
     }
   }
 
+  "CoreHelloWorldFramework should adhere to at most once launch guarantee during a crash-recovery" in {
```
timcharper, line 50: This is a great test!

core/src/test/scala/com/mesosphere/usi/core/SchedulerTest.scala
```diff
@@ -101,4 +105,55 @@ class SchedulerTest extends AkkaUnitTest with Inside {
     output.cancel()
     mesosCompleted.futureValue shouldBe Done
   }
+
+  "Persistence flow pipelines writes" in {
```
timcharper, line 112: After reading this test several times, I'm inclined to submit that we should just delete this test. I don't think that it's testing anything meaningful; it appears to be testing Akka itself, which already has tremendous test coverage.

Perhaps a test that would be more meaningful would be a unit test for the persistence flow itself which asserts that delete and create operations are applied in order, and that the order of elements being processed is preserved (perhaps some fuzzing can be applied by causing the storage operation Future to complete in a non-linear order)

timcharper (approved):
Please see my comment about the unit test; with rewrite it to just specifically test the ordering guarantees, and avoid using delays in our tests as possible.

core/src/test/scala/com/mesosphere/usi/core/SchedulerTest.scala
takirala: I agree that this test does not create much value. It was useful when i had a custom [graph stage logic](https://github.com/mesosphere/usi/blob/92c1c088336351b4fa3375b1bc009e97cc082968/core/src/main/scala/com/mesosphere/usi/core/Scheduler.scala#L166-L217) but since we got rid of that, we don't need this exact test anymore. I have rewritten the test using `Random.nextInt(100)` to apply "fuzzing", lmk if that is not ideal.

takirala:
Ping @zen-dog @meln1k @jeschkies please review :)

core/src/main/scala/com/mesosphere/usi/core/Scheduler.scala
```diff
@@ -80,14 +97,19 @@ object Scheduler {
     * @param specsSnapshot Snapshot of the current specs
     * @return Snapshot of the current state, as well as Source which produces StateEvents and Sink which accepts SpecEvents
     */
-  def asSourceAndSink(specsSnapshot: SpecsSnapshot, client: MesosClient)(
-      implicit mat: Materializer): (Future[StateSnapshot], Source[StateEvent, NotUsed], Sink[SpecUpdated, NotUsed]) = {
-    val flow = fromClient(client)
+  def asSourceAndSink(
```
zen-dog: Why all the re-formats here? Don't we have `scalafmt` to take care of that?

core/src/main/scala/com/mesosphere/usi/core/Scheduler.scala
```diff
@@ -269,4 +302,22 @@ object Scheduler {
       }
     }
   }
+
+  val PipeliningLimit = 128
```
zen-dog: Let's put this magic constant at least into some companion object.

core/src/main/scala/com/mesosphere/usi/core/Scheduler.scala
```diff
@@ -269,4 +302,22 @@ object Scheduler {
...
+
+  private[core] def persistenceFlow(
+      podRecordRepository: PodRecordRepository
+  )(implicit materializer: Materializer): Flow[SchedulerEvents, SchedulerEvents, NotUsed] = {
+    Flow[SchedulerEvents].mapAsync(1) { events =>
+      Source(events.stateEvents).collect {
+        case PodRecordUpdated(_, Some(podRecord)) => podRecordRepository.store(podRecord)
+        case PodRecordUpdated(podId, None) => podRecordRepository.delete(podId)
+      }.mapAsync(PipeliningLimit)(identity)
```
zen-dog: At this point in the stream, you've already created all the `store` and `delete` futures. But since `SchedulerEvent` are generated based on **one** new podspec or **one** new task status update, I believe this shouldn't be a problem. So let's move on with this solution.

core/src/test/scala/com/mesosphere/usi/core/SchedulerTest.scala
```diff
@@ -101,4 +105,41 @@ class SchedulerTest extends AkkaUnitTest with Inside {
...
+    Given("a persistence storage and a flow using it")
+    val fuzzyPodRecordRepo = new InMemoryPodRecordRepository {
+      val rand = scala.util.Random
+      override def store(record: PodRecord): Future[Done] = {
+        super.store(record).flatMap { _ =>
+          akka.pattern.after(rand.nextInt(100).millis, system.scheduler)(Future.successful(Done))
```
zen-dog, line 118: Nice usage of akkas `after` 👍

core/src/test/scala/com/mesosphere/usi/core/SchedulerTest.scala
```diff
@@ -101,4 +105,41 @@ class SchedulerTest extends AkkaUnitTest with Inside {
...
+        }
+      }
+    }
+    val (pub, sub) = TestSource
+      .probe[SchedulerEvents]
+      .via(Scheduler.persistenceFlow(fuzzyPodRecordRepo))
+      .toMat(TestSink.probe[SchedulerEvents])(Keep.both)
+      .run()
+
+    Given("a list of fake pod record events")
```
zen-dog: nit:
```suggestion
    And("a list of fake pod record events")
```

core/src/main/scala/com/mesosphere/usi/core/SchedulerLogicHandler.scala
```diff
@@ -110,6 +112,33 @@ private[core] class SchedulerLogicHandler(mesosCallFactory: MesosCalls) {
...
+      builder.process { (specs, state, _) =>
+        mesosEventsLogic.processEvent(specs, state, cachedPendingLaunch.pendingLaunch)(event)
+      }
+    }
+  }
+
+  def handlePodRecordSnapshot(podRecords: Map[PodId, PodRecord]): SchedulerEvents = {
+    handleFrame { builder =>
+      builder.process { (specs, state, _) =>
+        if (state.podRecords.nonEmpty || specs.podSpecs.nonEmpty) {
```
zen-dog: I don't understand this check: why shouldn't there be `podRecords` OR `podSpecs`?

zen-dog (commented):
Looks better and better. Left a few nits and a question.

core/src/main/scala/com/mesosphere/usi/core/SchedulerLogicHandler.scala
takirala: Since the snapshot is the first element that should/would be fed in to the scheduler graph, this is just a sanity check that everything so far is empty. Would you like to see it re-written in a different manner ?

core/src/main/scala/com/mesosphere/usi/core/Scheduler.scala
takirala: hm, this is in the scope of `object Scheduler {`, Do you mean to say to put this in a new object ?

core/src/main/scala/com/mesosphere/usi/core/Scheduler.scala
takirala: sorry, my intellij over-smarted me, reverted the changes.

core/src/main/scala/com/mesosphere/usi/core/SchedulerLogicHandler.scala
takirala: Also, see Tim's comment https://github.com/mesosphere/usi/pull/80#discussion_r275599366 I _think_ this code would be deleted eventually.

core/src/main/scala/com/mesosphere/usi/core/Scheduler.scala
timcharper: Yeah, this is my impression also

core/src/main/scala/com/mesosphere/usi/core/CallerThreadExecutionContext.scala
jeschkies: Can we add a comment with a link to a JIRA or just a comment? I feel this will never be addressed and will confuse future maintainers.

core/src/main/scala/com/mesosphere/usi/core/Scheduler.scala
jeschkies: @zen-dog already provided a way to configure USI with `reference.conf` and `application.conf`. See https://github.com/mesosphere/usi/blob/master/mesos-client/src/main/scala/com/mesosphere/mesos/conf/MesosClientSettings.scala.

core/src/main/scala/com/mesosphere/usi/core/Scheduler.scala
jeschkies: Please excuse my ignorance I probably miss some context.

First off, is `Source(...).collect...mapAsync.runWith` returning a future? I don't get how `Flow[SchedulerEvents].mapAsync(1)` would work otherwise. Could we create a simple helper function:

```scala
def presistChanges(SchedulerEvents): Future[???] = {
  Source(events.stateEvents).collect {
    case PodRecordUpdated(_, Some(podRecord)) => podRecordRepository.store(podRecord)
    case PodRecordUpdated(podId, None) => podRecordRepository.delete(podId)
  }
  .mapAsync(PipeliningLimit)(identity)
  .runWith(Sink.ignore)
  .map(_ => events)(CallerThreadExecutionContext.context)
}

Flow[SchedulerEvents].mapAsync(1)(persistChanges)
```

Second, on first sight it seems we are calling `store` and `delete` for all `stateEvents` and then limit them. I expect Akka to only make the call when `mapAsync` pulls the next items. Is that correct @timcharper? What do you think about making this more explicit by providing the call as anonymous functions?

```scala
def presistChanges(SchedulerEvents): Future[???] = {
  Source(events.stateEvents).collect {
    case PodRecordUpdated(_, Some(podRecord)) => () => podRecordRepository.store(podRecord)
    case PodRecordUpdated(podId, None) => () => podRecordRepository.delete(podId)
  }
  .mapAsync(PipeliningLimit) { call => call() }
  .runWith(Sink.ignore)
  .map(_ => events)(CallerThreadExecutionContext.context)
}
```

core/src/main/scala/com/mesosphere/usi/core/SchedulerLogicGraph.scala
jeschkies: Cool. Could we add a longer comment above?

core/src/main/scala/com/mesosphere/usi/core/SchedulerLogicHandler.scala
jeschkies: What do you think about using `assert` instead?

timcharper: podRecords snapshot is staying

core/src/main/scala/com/mesosphere/usi/core/SchedulerLogicGraph.scala
timcharper: To be clear, removal of podSpec snapshots will not preclude the need for startGraph callback and this preStart code; this will stay.

core/src/main/scala/com/mesosphere/usi/core/Scheduler.scala
takirala: After discussing this with @timcharper , we refactored the code to apply Pipelining limit across multiple `SchedulerEvents` objects, i have also applied @jeschkies suggestion on using `() => Future[..]` instead of `Future[..]`. The new approach uses Future's but restricts them with an upper limit. Added a test case to verify the same.

core/src/main/scala/com/mesosphere/usi/core/Scheduler.scala
```diff
@@ -269,4 +294,22 @@ object Scheduler {
       }
     }
   }
+
+  private[core] def persistenceFlow(
+      podRecordRepository: PodRecordRepository): Flow[SchedulerEvents, SchedulerEvents, NotUsed] = {
+    Flow[SchedulerEvents].mapConcat { events =>
+      val ops: List[() => Future[Option[SchedulerEvents]]] = events.stateEvents.collect {
```
jeschkies: What do you think of factoring this out into a helper function?

core/src/main/scala/com/mesosphere/usi/core/SchedulerLogicGraph.scala
jeschkies: Ah, because this loads, the pod records, right? I still don't quite understand why it is a callback. What do you think of passing `initialPodRecods` to `createLogic`? This would simplify things a bit and the loading of the pod records should not be part of the logic anyways IMHO.

takirala: Not sure if/how I can pass an extra param to `override def createLogic`. One thought I had was to set an attribute on graph component but it feels like an overkill. Do you have any other thoughts ?

otoh, i don't understand why loading the pod records should not be part of logic - we are creating a graph component which should not start pulling elements from upstream until a certain precondition is satisfied. In our case, the precondition is that the given future completes with an expected (Success) result. Since this future completion is non deterministic (it can even fail), i think it is fair to register the call back in `prestart` method. Ultimately, this callback decides whether (if at all) or not to start pulling elements from upstream.

jeschkies: I missed that `createLogic` is predefined. However, `initialPodRecords` could simply be the loaded record set. Why should it be a future?

core/src/main/scala/com/mesosphere/usi/core/Scheduler.scala
```diff
@@ -56,15 +64,23 @@ object Scheduler {
 
   type StateOutput = (StateSnapshot, Source[StateEvent, Any])
 
-  def fromSnapshot(specsSnapshot: SpecsSnapshot, client: MesosClient): Flow[SpecUpdated, StateOutput, NotUsed] =
-    Flow[SpecUpdated].prefixAndTail(0).map { case (_, rest) => specsSnapshot -> rest }.via(fromClient(client))
+  val PipeliningLimit = ConfigFactory.load().getInt("persistence.pipeline-limit")
+
+  def fromSnapshot(
```
meln1k: Do we actually need this method?

core/src/main/scala/com/mesosphere/usi/core/SchedulerLogicGraph.scala
```diff
@@ -41,22 +42,26 @@ object SchedulerLogicGraph {
   * It's existence is only warranted by forecasted future needs. It's kept as a graph with an internal buffer as we will
   * likely need timers, other callbacks, and additional output ports (such as an offer event stream?).
   */
-class SchedulerLogicGraph(mesosCallFactory: MesosCalls)
+private[core] class SchedulerLogicGraph(
+    mesosCallFactory: MesosCalls,
+    initialPodRecords: => Future[Map[PodId, PodRecord]])
```
meln1k: Is it possible to use a plain Map[PodId, PodRecord] instead of Future?

core/src/main/scala/com/mesosphere/usi/core/Scheduler.scala
zen-dog: Ah, I see. You don't have to use `reference.conf` right away, but we should put all related settings into one place. A case class `SchedulerSettings.scala` would be ok (see other similar files). Also, the variable should be renamed to something more representative e.g. `schedulerPersistenceConcurrencyLimit` (probably not the best name I came up with)

core/src/main/scala/com/mesosphere/usi/core/SchedulerLogicGraph.scala
takirala: By making pod records a future - we have the option to chose to **either** succeed loading the snapshot _and then_ start the graph **or** mark graph stage as failed to start. If it's not a future, then we have to handle the failed future before creating this graph component (and it brings the question - if we fail to make the persistence call, should we fail to create the blueprint of the graph ?) cc @meln1k

takirala: Also, note that the `SchedulerLogicHanlder` has a `Map` instead of a `Future[Map]` - https://github.com/mesosphere/usi/pull/80/files#diff-704496199082f7d74e9d62d315f0a7abR66 and it does make sense to me because if we have a "handler", then it can be expected to have a valid snapshot. But in case of `SchedulerLogicGraph` where  we are "building" the graph, we can chose to verify a precondition (in this case - successful completion of the given Future) before we chose to start the graph.

I have made the `initialPodRecords` to be a anon function to delay the pod record loading. If you still think that having a `Future` in graph pre condition is not ideal, I can reformat the code to complete the future in `Scheduler` and fail fast before a graph blue print is generated. wdyt ?

jeschkies: I'm just wondering if loading the snapshot in the scheduler factory would be simpler. We can always handle errors there. Ideally the factory method would fail if the snapshot loading would fail. The caller would not even have a failed graph stage. How should the user handle a failed graph stage anyways?

core/src/main/scala/com/mesosphere/usi/core/Scheduler.scala
```diff
@@ -230,30 +243,36 @@ object Scheduler {
...
-      (StateSnapshot.empty, stateUpdates)
+      (stateSnapshot, stateUpdates)
   }
 
-  def unconnectedGraph(
-      mesosCallFactory: MesosCalls): BidiFlow[SpecInput, StateOutput, MesosEvent, MesosCall, NotUsed] = {
+  private[core] def unconnectedGraph(
+      mesosCallFactory: MesosCalls,
+      podRecordRepository: PodRecordRepository): BidiFlow[SpecInput, StateOutput, MesosEvent, MesosCall, NotUsed] = {
+    val schedulerLogicGraph = new SchedulerLogicGraph(mesosCallFactory, loadPodRecords(podRecordRepository))
```
jeschkies, line 263: This is getting closer to it but I don't understand why I cannot be a async. The two main factory methods `asFlow` and `asSinkAndSource` return futures. So you could simply call

```scala
  def asFlow(specsSnapshot: SpecsSnapshot, client: MesosClient, , podRecordRepository: PodRecordRepository)(
      implicit materializer: Materializer): Future[(StateSnapshot, Flow[SpecUpdated, StateEvent, NotUsed])] = async {

    implicit val ec = scala.concurrent.ExecutionContext.Implicits.global //only for ultra-fast non-blocking onComplete

    val initialRecords = await(loadRecords(repository))
    val (snap, source, sink) = asSourceAndSink(specsSnapshot, client, initialRecords)
    (await(snap), Flow.fromSinkAndSourceCoupled(sink, source))
}
```

takirala: Discusses this offline, In an ideal world we should be using async-await construct to do this in a non-blocking fashion. But factory methods `fromFlow` and `fromClient` requires this to be done in a blocking fashion. I guess we can find a more cleaner solution once we finalize on a limited number of factory methods.

Based on the isolated fact that _we can't make any progress until we load the records_ I guess this is okay for now.

takirala:
@zen-dog @meln1k can you please re-review this :)

examples/core-hello-world/src/main/scala/com/mesosphere/usi/examples/CoreHelloWorldFramework.scala
```diff
@@ -127,10 +119,26 @@ object CoreHelloWorldFramework extends StrictLogging {
...
     )
 
-    val specsSnapshot = SpecsSnapshot(
+    SpecsSnapshot(
       podSpecs = Seq(podSpec),
       reservationSpecs = Seq.empty
     )
+  }
+
+  def buildGraph(
```
zen-dog: `s/buildGraph/init` may be? Or `fromFrameworkInfo`? This method doesn't create any graphs directly.

examples/core-hello-world/src/main/scala/com/mesosphere/usi/examples/CoreHelloWorldFramework.scala
```diff
@@ -140,7 +148,7 @@ object CoreHelloWorldFramework extends StrictLogging {
     // Note: this is hardly realistic since an orchestrator will need to react to StateEvents by sending SpecUpdates
     // to the scheduler. We're making our lives easier by ignoring this part for now - all we care about is to start
     // a "hello-world" task once.
-    Source.maybe
+    val result: Future[Done] = Source.maybe
```
zen-dog:
```suggestion
    val completed: Future[Done] = Source.maybe
```

examples/core-hello-world/src/test/scala/com/mesosphere/usi/examples/CoreHelloWorldFrameworkTest.scala
```diff
@@ -31,6 +47,68 @@ class CoreHelloWorldFrameworkTest extends AkkaUnitTest with MesosClusterTest wit
...
+
+    And("an initial pod spec is launched")
+    val specsSnapshot = CoreHelloWorldFramework.generateSpecSnapshot
+    val (_, sub1) = runGraphAndFeedSnapshot(scheduler, specsSnapshot)
+
+    Then("receive an empty snapshot followed by other state events")
+    val e1 = sub1.requestNext()
+    e1 shouldBe a[StateSnapshot]
+    e1.asInstanceOf[StateSnapshot].podRecords shouldBe empty
+    val e2 = sub1.requestNext()
```
zen-dog: I'm not a big fan of `eN` kind of names. Wdyt about making them more descriptive e.g. `snapshot`, `record`, `status` etc. or maybe reusing `var event`?

persistence-zookeeper/src/test/scala/com/mesosphere/usi/repository/InMemoryRepositoryTest.scala
```diff
@@ -0,0 +1,13 @@
+package com.mesosphere.usi.repository
+
+import com.mesosphere.utils.UnitTest
+import com.typesafe.scalalogging.StrictLogging
+
+class InMemoryRepositoryTest extends UnitTest with RepositoryBehavior with StrictLogging {
```
zen-dog, line 7: I'm confused by the location of this file (and other _in-memory-repository_ related files). Why are those in the `persistence-zookeeper` sub-module? Why is `InMemoryPodRecordRepository.scala` in the `persistence` sub-module? Ideally:
```
usi
  |
  ...
   - persistence # traits and interfaces, no implementation
   - persistence-zookeeper # zookeeper implementation
   - test-utils # In-memory implementation here (see DummyMetrics for example)? 
      It is not supposed to be used in production so this seems appropriate
```

zen-dog (changes requested):
I left a few nits about naming and similar. My biggest question mark is the location of the `InMemoryStore*` related files. See my comment. Once clarified, this PR is good with me!

examples/core-hello-world/src/test/scala/com/mesosphere/usi/examples/CoreHelloWorldFrameworkTest.scala
timcharper: Yeah, actually, lets do inside here:

```
inside(sub1.requestNext) {
  case snapshot: Snapshot =>
    snapshot.podRecords shouldBe empty
}
```

persistence-zookeeper/src/test/scala/com/mesosphere/usi/repository/InMemoryRepositoryTest.scala
takirala: I guess it was done so that we don't need to have 
```java
implementation project(':test-utils')
```
and can just do 
```java
implementation project(':persistence')
testCompile project(':test-utils')
```
in order to instantiate the persistence layer. I have moved `InMemoryPodRecordRepository` to `test-utils` and added a note to `build.gradle` file to replace the `implementation project(':test-utils')` with `testCompile project(':test-utils')` when we have a `persistence-zookeeper` implementation. 

The only thing I am unsure about is the location of `InMemoryRepositoryTest` which is currently in `persistence-zookeeper`. I can not move this to `test-utils` as it would cause a cyclic dependency. I guess having this here is _fine_ as its intention is to verify the `RepositoryBehavior` which would be used to verify a zk persistence layer in future. wdyt ?

core/src/main/scala/com/mesosphere/usi/core/Scheduler.scala
jeschkies: `fromClient` and `fromFlow` could just be non-blocking. The signature of these methods is not set in stone.

jeschkies (approved):
Aside from the blocking call in the `Scheduler` factory methods I'm fine with this patch 🎉

zen-dog (approved):
🎉

takirala:
Thanks for the reviews @zen-dog @jeschkies @meln1k and @timcharper :)

PR #83, opened 2019-04-11
Remove the size limit on mesos http stream

meln1k:
By default, Akka HTTP restricts the size of any HTTP entity to 8Mb. This usually helps to prevent OOM, but it's not very applicable when we want to consume an endless stream from mesos. This fix removes the size limit on that mesos HTTP entity.


alenkacz (approved):

zen-dog (approved):
🚢 nice find!

PR #87, opened 2019-05-09
 Support multiple Mesos master URLs. 

jeschkies:
Summary:
This change makes the Mesos master URLs a list of addresses to support HA mode.
All URLs are tried one after another. It also makes the `MesosClientSettings` a simple
class to ensure backwards compatibility.

JIRA issues: DCOS-53671


examples/core-hello-world/src/main/resources/application.conf
```diff
@@ -1,8 +1,8 @@
 # This is still the default configuration but it's good to see what can be changed
 mesos-client {
 
-  master-url: "127.0.0.1:5050"
-  master-url: ${?MESOS_MASTER_URL}
+  master-url: [ "127.0.0.1:5050" ]
```
timcharper: Let's put a protocol here (and force a protocol). Maybe another PR?

timcharper: Perhaps also we should change the key name from `master-url` to `master-urls` ?

mesos-client/src/main/scala/com/mesosphere/mesos/conf/MesosClientSettings.scala
```diff
@@ -1,25 +1,57 @@
...
+    val sourceBufferSize: Int) {
+
+  private def copy(
+      masters: Seq[URL] = this.masters,
+      redirectRetries: Int = this.redirectRetries,
+      idleTimeout: FiniteDuration = this.idleTimeout,
+      sourceBufferSize: Int = this.sourceBufferSize) =
+    new MesosClientSettings(masters, redirectRetries, idleTimeout, sourceBufferSize)
+
+  def withMasters(urls: URL*): MesosClientSettings = copy(masters = urls)
```
timcharper: Let's please just have one method. `withMaster` is unclear whether it appends or replaces the current masters.

examples/core-hello-world/src/main/resources/application.conf
jeschkies: That's a bug from my side. Thanks.

mesos-client/src/main/scala/com/mesosphere/mesos/conf/MesosClientSettings.scala
jeschkies: Java does not support var args. `withMasters(urls: URL*)` is a `withMasters(scala.collection.Seq<URL> urls)` in Java which is annoying to use.

examples/core-hello-world/src/main/resources/application.conf
```diff
@@ -1,10 +1,10 @@
 # This is still the default configuration but it's good to see what can be changed
 mesos-client {
 
-  master-url: "127.0.0.1:5050"
-  master-url: ${?MESOS_MASTER_URL}
+  master-url: [ "http://127.0.0.1:5050" ]
+  master-url: [ ${?MESOS_MASTER_URL} ]
```
zen-dog: Does env var override work for arrays?

zen-dog: To answer my own question - it does, but you can only pass **one** value.

zen-dog (approved):
If it's green - go for it! 🚢

examples/core-hello-world/src/main/resources/application.conf
timcharper: Oof, yeah. A quick search reveals that we probably need to resort to splitting for env-var overrides :'(

https://github.com/lightbend/config/issues/163

timcharper (changes requested):
Lets get clarity on how we will solve the environment variable overrides; I'm wondering if we will want to do the comma delimited authority approach. I've a thread open in the Mesos slack: https://mesos.slack.com/archives/C1K2TAJ1L/p1557739907081100

mesos-client/src/main/resources/reference.conf
```diff
@@ -1,10 +1,10 @@
 mesos-client {
 
-  # Mesos master host:port
-  master-url: "127.0.0.1:5050"
+  # Comma separated list of Mesos master protocol://host:port,...
```
timcharper: docs doesn't seem to match the implementation

timcharper (changes requested):

mesos-client/src/main/resources/reference.conf
timcharper:
```suggestion
  # Comma separated list of Mesos master protocol://host1:port1,protocol://host2:port2,...
  # protocol must either be http or https
```

timcharper (approved):

PR #89, opened 2019-05-15
Inject scheduler settings.

jeschkies:
Summary:
The settings should be injected and the loader should allow for passing a class loader.

This is a follow up to #87.


takirala (approved):
looks like #90 will change some of the things in this PR, but LGTM otherwise.

timcharper (approved):

PR #92, opened 2019-05-17
Emit unrecognized pod events.

jeschkies:
Summary:
In cases when USI has no pod record nor spec it will emit an
`UnrecognizedPod` state event and lets the framework author decide how
to handle this case.

JIRA issues: DCOS-54000


jeschkies:
Tests are still missing.

timcharper:
Why do we need this? The intention is that the implementation framework replicates the state by applying events; unrecognized pods are already identifiable as statuses without records. A separate mechanism is not needed.

jeschkies:
Won't do.

PR #94, opened 2019-05-21
Handle JWT session for IAM.

jeschkies:
Summary:
Mesos can require authentication to allow access to the v1 API. This
patch introduces a `Session` object to handle credentials and session
tokens for requests. Tokens are provided by a `CredentialsProvider`.

We provide a simple provider for basic authentication and one that
authenticates with DC/OS to retrieve a JSON Web token to get access
to the Mesos API hosted by a DC/OS cluster.

The session object uses the provider to make authenticated calls to Mesos.
The object also refreshes the token transparently.

JIRA issues: DCOS-53496


jeschkies:
This is just an approach. I researched a few clients and like [Requests' Session Object](https://2.python-requests.org/en/master/user/advanced/#session-objects) the most. The IAM client should be straight forward. JWT renewal should be triggered in the connection flow that a session provides, IMHO. That would be stateful though.

mesos-client/src/main/scala/com/mesosphere/mesos/client/CredentialsProvider.scala
```diff
@@ -0,0 +1,78 @@
...
+  def tokens(): Source[HttpCredentials, NotUsed]
+}
+
+case class JwtProvider(uid: String, privateKey: String, root: URL)(implicit system: ActorSystem, materializer: ActorMaterializer, context: ExecutionContext) extends CredentialsProvider {
+  import org.json4s._
+  import org.json4s.JsonDSL._
+  import JsonMethods.{parse, render, compact}
+
+  private def expireIn(duration: Duration): Long = Instant.now.getEpochSecond + duration.toSeconds
+  private val claim = JObject(("uid", uid), ("exp", expireIn(60.minutes))) // TODO: configure token expiration.
```
timcharper: I really like the stabby-arrow syntax for tuples in this context:

```suggestion
  private val claim = JObject("uid" -> uid, "exp" -> expireIn(60.minutes)) // TODO: configure token expiration.
```

mesos-client/src/main/scala/com/mesosphere/mesos/client/CredentialsProvider.scala
```diff
@@ -0,0 +1,78 @@
...
+
+  val acsTokenRequest: HttpRequest = {
+    val token = JwtJson4s.encode(claim, privateKey, RS256)
+    val data: String = compact(render(JObject(("uid", uid), ("token", token))))
```
timcharper:
```suggestion
    val data: String = compact(render(JObject("uid" -> uid, "token" -> token)))
```

mesos-client/src/main/scala/com/mesosphere/mesos/client/CredentialsProvider.scala
```diff
@@ -0,0 +1,78 @@
...
+      entity = HttpEntity(ContentTypes.`application/json`, data)
+    )
+  }
+
+  /**
+    * Provides a [[Source]] of DC/OS acs tokens wrapped in [[HttpCredentials]]. The token is refreshed on each pull.
+    *
+    * @return A new source of [[HttpCredentials]]
+    */
+  def tokens(): Source[HttpCredentials, NotUsed] = {
```
timcharper: neat!

mesos-client/src/main/scala/com/mesosphere/mesos/client/CredentialsProvider.scala
```diff
@@ -0,0 +1,78 @@
...
+    val port = if(root.getPort == -1) root.getDefaultPort else root.getPort
+
+    Source.repeat(acsTokenRequest)
+      .via(Http().outgoingConnectionHttps(root.getHost, port))
+      .mapAsync(1) { response =>
```
timcharper: After thinking about this, I don't think Akka streams should be used for this.

1. It's not as lazy as you think. MapAsync adds an implicit buffer which will cause multiple http requests to be issued ahead of time to fill the buffer before any elements are requested; therefore, by the time they are requested, they are stale.
2. This concern is just as easily handled by a method which can be called, on demand, and return a future. Making it an Akka stream source provides virtually no benefit over this.

timcharper (changes requested):

mesos-client/src/main/scala/com/mesosphere/mesos/client/CredentialsProvider.scala
```diff
@@ -0,0 +1,78 @@
...
+import org.json4s.native.JsonMethods
+import pdi.jwt.JwtJson4s
+import pdi.jwt.JwtAlgorithm.RS256
+
+import scala.concurrent.ExecutionContext
+import scala.concurrent.duration._
+
+trait CredentialsProvider {
+
+  def tokens(): Source[HttpCredentials, NotUsed]
```
timcharper: I suppose tests for this method are pending?

mesos-client/src/main/scala/com/mesosphere/mesos/client/CredentialsProvider.scala
jeschkies: Yeah, I thought there would be an easy way to first pull from the tokens stream and then just consume Mesos calls. I'm working on a simple graph stage today and hope that it''l work out.

mesos-client/src/main/scala/com/mesosphere/mesos/client/MesosClient.scala
```diff
@@ -122,7 +127,74 @@ trait MesosClient {
...
+      )
+
+    case class SessionFlow() extends GraphStage[FlowShape[Array[Byte], HttpRequest]] {
+
+      private val callsInlet = Inlet[Array[Byte]]("specs")
+      private val requestsOutlet = Outlet[HttpRequest]("effects")
+      override val shape = FlowShape.of(callsInlet, requestsOutlet)
+
+      override def createLogic(inheritedAttributes: Attributes): GraphStageLogic = {
+        new GraphStageLogic(shape) {
```
jeschkies: There is something wrong here since we only process one call. @timcharper and @meln1k, any ideas?

mesos-client/src/test/scala/com/mesosphere/mesos/client/SessionFlowTest.scala
```diff
@@ -0,0 +1,50 @@
...
+            logger.info(s"Processing $index")
+            if (index % 2 == 1)  Success(HttpResponse(StatusCodes.Unauthorized))
+            else Success(HttpResponse())
+        }
+      Source(calls)
+        .via(f.session.post(connection))
+        .runForeach(println)
+        .futureValue
+    }
+  }
```
jeschkies: You will see requests for "second" and "fourth" being dropped. This is not what we want. `recoverWithRetries` requires a new source of elements which we don't have.

jeschkies:
I tried to find something on retrying single requests and found this article https://blog.colinbreck.com/backoff-and-retry-error-handling-for-akka-streams/. Which is kind of funny since the simplify connection pool.

examples/mesos-client-example/src/main/resources/application.conf
```diff
@@ -27,4 +27,13 @@ akka {
     # Set to `infinite` to completely disable idle timeouts.
     idle-timeout = infinite
   }
+
+  ssl-config {
+    trustManager = {
+      stores = [
+        { type: "PEM", path: ${?DCOS_CERT}  }     # Added trust store
```
jeschkies, line 34: This breaks the tests as there is no `DCOS_CERT` env var defined.

jeschkies:
Following https://blog.colinbreck.com/integrating-akka-streams-and-akka-actors-part-ii/ I think we should manage the HTTP connection and the session in an actor. We want sequential requests anyways so the backpressure on TCP level is not that important.

mesos-client/src/main/scala/com/mesosphere/mesos/client/Session.scala
```diff
@@ -0,0 +1,220 @@
...
+
+object SessionActor {
+  def props(
+      credentialsProvider: CredentialsProvider,
+      requestFactory: (Array[Byte], Option[HttpCredentials]) => HttpRequest): Props = {
+    Props(new SessionActor(credentialsProvider, requestFactory))
+  }
+}
+
+class SessionActor(
```
jeschkies, line 125: This is an new proposal following the Colin Breck's [posts](https://blog.colinbreck.com/integrating-akka-streams-and-akka-actors-part-ii/). I have to admit that he is right: Handling state and retries is much simpler in an actor. Please compare `Session#post` with `Session#postSimple`. `Flow.ask` will also yield backpressure and since we are doing single sequential HTTP requests anyways we would not benefit from a connection pool anyways. Strict cluster send a `Connection: close` header so we have to reestablish the connection for each Mesos call anyways.

mesos-client/src/main/scala/com/mesosphere/mesos/client/MesosClient.scala
```diff
@@ -363,42 +385,31 @@ class MesosClientImpl(
...
-
   override val mesosSink: Sink[Call, Future[Done]] =
     Flow[Call]
       .via(sharedKillSwitch.flow[Call])
       .via(debug("Sending "))
       .via(callSerializer)
-      .via(requestBuilder)
-      .via(httpConnection)
+      .via(session.postSimple)
+//      .via(session.post(session.connectionPool))
```
meln1k: I think it should be removed.

mesos-client/src/main/scala/com/mesosphere/mesos/client/Session.scala
```diff
@@ -0,0 +1,220 @@
...
+    }
+
+  /** @return The connection pool for this session. */
+  def connectionPool(implicit system: ActorSystem): Flow[HttpRequest, Try[HttpResponse], NotUsed] = {
+    // Disable pipelining.
+    val poolSettings = ConnectionPoolSettings("").withMaxConnections(1).withPipeliningLimit(1)
+    if (isSecured) {
+      Flow[HttpRequest]
+        .map(_ -> NotUsed)
+        .via(Http().cachedHostConnectionPoolHttps(host = url.getHost, port = port, settings = poolSettings))
```
meln1k: I feel like `cachedHostConnectionPool` is not needed here since one shouldn't cache connections to mesos.

jeschkies: We don't need it at all if we opt for the actor solution.

mesos-client/src/main/scala/com/mesosphere/mesos/client/Session.scala
```diff
@@ -0,0 +1,220 @@
...
+
+        // Queue current call again.
+        self ! originalCall
+      } else {
+        originalSender ! response
+      }
+  }
+}
+
+case class SessionFlow(
```
timcharper: I'm not entirely convinced that a custom graph stage is needed here; can we perhaps go over this tomorrow together?

jeschkies: That's what the walkthrough is for. I'd use the actor below. Remember we have to keep the session token state somewhere and refresh it if a request returns an `HTTP 401 Unauthorized`. After some research I'm uncertain that a stream could handle this easier than a simple actor. However, I'm open to suggestions.

jeschkies:
A rough interaction diagram [Mesos Client Auth Workflow.pdf](https://github.com/mesosphere/usi/files/3261838/Mesos.Client.Auth.Workflow.pdf)

mesos-client/src/main/scala/com/mesosphere/mesos/client/CredentialsProvider.scala
```diff
@@ -0,0 +1,95 @@
...
+  * @param uid The user id.
+  * @param privateKey The JWT secret.
+  * @param root The DC/OS root URL.
+  * @param system The [[ActorSystem]] used for HTTP requests.
+  * @param materializer The [[ActorMaterializer]] use to unmarshal HTTP responses.
+  * @param context The execution context for transforming the future HTTP response.
+  * @see See [[https://docs.mesosphere.com/1.11/security/ent/iam-api/#obtaining-an-authentication-token]] for details on
+  *      the protocol.
+  */
+case class JwtProvider(uid: String, privateKey: String, root: URL)(
```
zen-dog: Class name should be more like `DCOSServiceAccountProvider` or similar

mesos-client/src/main/scala/com/mesosphere/mesos/client/CredentialsProvider.scala
```diff
@@ -0,0 +1,95 @@
...
+      method = HttpMethods.POST,
+      uri = Uri(s"$root/acs/api/v1/auth/login"),
+      entity = HttpEntity(ContentTypes.`application/json`, data)
+    )
+  }
+
+  override def nextToken(): Future[HttpCredentials] = {
+    logger.debug(s"Fetching next token from $root")
+    Http().singleRequest(acsTokenRequest).flatMap { response =>
+      // TODO: Use json unmarshaller directly.
```
zen-dog: TODO!

mesos-client/src/main/scala/com/mesosphere/mesos/client/MesosClient.scala
```diff
@@ -158,40 +159,58 @@ object MesosClient extends StrictLogging with StrictLoggingFlow {
...
-    val httpConnection = Http().outgoingConnection(url.getHost, url.getPort)
+    def createPostRequest(bytes: Array[Byte], maybeCredentials: Option[HttpCredentials]): HttpRequest =
+      HttpRequest(
+        HttpMethods.POST,
+        uri = Uri(s"${url.getPath}/api/v1/scheduler"),
+        entity = HttpEntity(ProtobufMediaType, bytes),
+        headers = headers.Accept(ProtobufMediaType) :: maybeCredentials.map(Authorization(_)).toList
+      )
+
+    // TODO: Probably should already create session.
```
zen-dog: TODO!

mesos-client/src/main/scala/com/mesosphere/mesos/client/MesosClient.scala
```diff
@@ -365,40 +386,25 @@ class MesosClientImpl(
...
-        case status if status.isFailure() =>
-          logger.info(s"A request to mesos failed with response: ${response}")
-          response.discardEntityBytes()
-          throw new IllegalStateException(s"Failed to send a call to mesos")
-        case _ =>
-          logger.debug(s"Mesos call response: $response")
-          response.discardEntityBytes()
+      if (response.status.isFailure()) {
+        // TODO: Do not block.
+        val body = Await.result(Unmarshal(response.entity).to[String], 10.seconds)
```
zen-dog: I get that you want to log the body, but the blocking is meh

mesos-client/src/main/scala/com/mesosphere/mesos/client/MesosClient.scala
```diff
@@ -365,40 +386,25 @@ class MesosClientImpl(
...
-      response.status match {
-        case status if status.isFailure() =>
-          logger.info(s"A request to mesos failed with response: ${response}")
-          response.discardEntityBytes()
-          throw new IllegalStateException(s"Failed to send a call to mesos")
-        case _ =>
-          logger.debug(s"Mesos call response: $response")
-          response.discardEntityBytes()
+      if (response.status.isFailure()) {
+        // TODO: Do not block.
```
zen-dog: TODO!

mesos-client/src/main/scala/com/mesosphere/mesos/client/Session.scala
```diff
@@ -0,0 +1,129 @@
...
+
+  override def preStart(): Unit = {
+    super.preStart()
+    credentialsProvider.nextToken().pipeTo(self)
+  }
+
+  override def receive: Receive = initializing
+
+  def initializing: Receive = {
+    // TODO: Handle nextToken errors.
```
zen-dog: TODO!

mesos-client/src/main/scala/com/mesosphere/mesos/client/Session.scala
```diff
@@ -0,0 +1,171 @@
...
+      throw ex
+    case _ => stash()
+  }
+
+  def initialized(credentials: HttpCredentials): Receive = {
+    case call: Array[Byte] =>
+      val request = requestFactory(call, Some(credentials))
+      val originalSender = sender()
+      logger.info("Processing next call.")
+      // The TLS handshake for each connection might be an overhead. We could potentially reuse a connection.
```
jeschkies, line 159: @timcharper, @zen-dog and @meichstedt, I will not implement connection reuse in this PR. I cut scope to get the Keep Alive Framework and the new Jenkins on Soak and MWT. We can always redesign the `SessionActor` to reuse TLS sessions.

mesos-client/src/main/scala/com/mesosphere/mesos/client/CredentialsProvider.scala
zen-dog: How is this resolved exactly? 🤔

examples/mesos-client-example/src/main/scala/com/mesosphere/mesos/examples/MesosClientExampleFramework.scala
```diff
@@ -24,13 +26,15 @@ import scala.collection.JavaConverters._
...
-  implicit val materializer = ActorMaterializer()
+class MesosClientExampleFramework(settings: MesosClientSettings, authorization: Option[CredentialsProvider] = None)(
+    implicit system: ActorSystem,
+    materializer: Materializer)
+    extends StrictLoggingFlow {
   implicit val executionContext = system.dispatcher
 
   val frameworkInfo = FrameworkInfo
     .newBuilder()
+    .setPrincipal("strict-usi")
```
zen-dog: nit: Historically the principal is the same as framework name so `mesos-client-example` in this case? I'm open to better names but keep them the same for simplicity. (don't forget the docs above)

examples/mesos-client-example/src/main/scala/com/mesosphere/mesos/examples/MesosClientExampleFramework.scala
```diff
@@ -73,9 +77,38 @@ class MesosClientExampleFramework(settings: MesosClientSettings) extends StrictL
...
+    * 1. Launch strict cluster and setup the DC/OS CLI.
+    * 2. Create key pair:
+    *   {{{dcos security org service-accounts keypair usi.private.pem usi.pub.pem}}}
+    * 3. Create user strict-usi:
+    *   {{{dcos security org service-accounts create -p usi.pub.pem -d "For testing USI on strict" strict-usi}}}
+    * 4. Grant strict-usi access:
+    *   {{{curl -L -X PUT -k -H "Authorization: token=$(dcos config show core.dcos_acs_token)" "$(dcos config show core.dcos_url)/acs/api/v1/acls/dcos:superuser/users/strict-usi/full"}}}
+    * 5. Download SSL certs:
+    *   {{{wget --no-check-certificate -O dcos-ca.crt "$(dcos config show core.dcos_url)/ca/dcos-ca.crt"}}}
+    * 6. Run with {{{DCOS_CERT=path/to/dcos-ca.crt ./gradlew :mesos-client-example:run --stacktrace --args "https://<DC/OS IP> path/to/usi.private.pem"}}}
```
zen-dog, line 92: 👏 for docs!

examples/mesos-client-example/src/main/scala/com/mesosphere/mesos/examples/MesosClientExampleFramework.scala
```diff
@@ -73,9 +77,38 @@ class MesosClientExampleFramework(settings: MesosClientSettings) extends StrictL
...
+    */
   def main(args: Array[String]): Unit = {
-    MesosClientExampleFramework(MesosClientSettings.load())
+
+    require(args.length == 2, "Too many arguments")
```
zen-dog:
```suggestion
    require(args.length == 2, "Please provide two arguments: <mesos-url> <private-key-file>")
```

examples/mesos-client-example/src/main/scala/com/mesosphere/mesos/examples/MesosClientExampleFramework.scala
```diff
@@ -73,9 +77,38 @@ class MesosClientExampleFramework(settings: MesosClientSettings) extends StrictL
...
+
+    implicit val system = ActorSystem()
+    implicit val materializer = ActorMaterializer()
+    implicit val context = system.dispatcher
+
+    val dcosRoot = new URL(args(0))
+    val privateKey = scala.io.Source.fromFile(args(1)).mkString
+    val provider = DcosServiceAccountProvider("strict-usi", privateKey, dcosRoot)
```
zen-dog: Could you please make the `provider` optional, depending on whether the `private-key` is passed? This example should run locally and against permissive clusters.

mesos-client/src/main/scala/com/mesosphere/mesos/client/CredentialsProvider.scala
```diff
@@ -0,0 +1,93 @@
...
+  private def expireIn(duration: Duration): Long = Instant.now.getEpochSecond + duration.toSeconds
+
+  /**
+    * Construct a claim to authenticate and retrieve a session token. The timeout is for this token ''not'' the session
+    * token that we are requesting. That means if ou login request takes more than twenty seconds the authentication will
+    * fail.
+    *
+    * @return a claim for the user that expires in twenty seconds.
+    */
+  private def claim: JsObject = Json.obj("uid" -> uid, "exp" -> expireIn(20.seconds))
```
zen-dog: Lets put `20.seconds` in the companion object at least

mesos-client/src/main/scala/com/mesosphere/mesos/client/CredentialsProvider.scala
```diff
@@ -0,0 +1,93 @@
...
+
+    HttpRequest(
+      method = HttpMethods.POST,
+      uri = Uri(s"$root/acs/api/v1/auth/login"),
+      entity = HttpEntity(ContentTypes.`application/json`, data)
+    )
+  }
+
+  override def nextToken(): Future[HttpCredentials] = async {
+    logger.debug(s"Fetching next token from $root")
```
zen-dog: Since this happens every ~20min I'd increase logging level:
```suggestion
    logger.info(s"Fetching next token from $root")
```

mesos-client/src/main/scala/com/mesosphere/mesos/client/MesosClient.scala
```diff
@@ -158,40 +158,57 @@ object MesosClient extends StrictLogging with StrictLoggingFlow {
...
+    def createPostRequest(bytes: Array[Byte], maybeCredentials: Option[HttpCredentials]): HttpRequest =
+      HttpRequest(
+        HttpMethods.POST,
+        uri = Uri(s"${url.getPath}/api/v1/scheduler"),
+        entity = HttpEntity(ProtobufMediaType, bytes),
+        headers = headers.Accept(ProtobufMediaType) :: maybeCredentials.map(Authorization(_)).toList
+      )
+
+    val port = if (url.getPort == -1) url.getDefaultPort else url.getPort
+    val isSecured: Boolean = url.getProtocol == "https"
```
zen-dog: wdyt:
```suggestion
    val httpConnection = url.getProtocol match {
      case "https" => Http().outgoingConnectionHttps(url.getHost, port)
      case _ => Http().outgoingConnection(url.getHost, port)
    }
```
```

File: mesos-client/src/test/scala/com/mesosphere/mesos/client/SessionActorTest.scala [hunk 27]
```diff
@@ -0,0 +1,38 @@
...
+import akka.http.scaladsl.model.headers.HttpCredentials
+import akka.testkit.TestProbe
+import com.mesosphere.utils.AkkaUnitTest
+
+import scala.concurrent.Future
+
+class SessionActorTest extends AkkaUnitTest {
+
+  "The Session Actor" should {
+    "replied with a response to the original sender" in {
```
zen-dog, line 18: Only one test? I understand that mocking IAM responses might be difficult but we need more tests. And maybe we don't need a complicated mock (like `app_mock.py`), a simple akka-http server [like this](https://github.com/mesosphere/marathon/blob/master/src/test/scala/mesosphere/marathon/core/health/impl/HealthCheckWorkerTest.scala#L59) might already suffice.

mesos-client/src/main/scala/com/mesosphere/mesos/client/Session.scala
```diff
@@ -0,0 +1,171 @@
...
+      entity = HttpEntity(MesosClient.ProtobufMediaType, bytes),
+      headers = MesosClient.MesosStreamIdHeader(streamId) :: maybeCredentials.map(Authorization(_)).toList
+    )
+  }
+
+  /** @return A flow that makes Mesos calls and outputs HTTP responses. */
+  def post(implicit system: ActorSystem, mat: Materializer): Flow[Array[Byte], HttpResponse, NotUsed] =
+    authorization match {
+      case Some(credentialsProvider) =>
+        implicit val askTimeout = util.Timeout(20.seconds)
```
zen-dog: Again, lets put the `20.seconds` in the companion object ;)

mesos-client/src/main/scala/com/mesosphere/mesos/client/Session.scala
```diff
@@ -0,0 +1,171 @@
...
+        val sessionActor = system.actorOf(SessionActor.props(credentialsProvider, createPostRequest))
```
zen-dog, line 50: So we're recreating the session actor for each request?? I don't think this is what we want.

mesos-client/src/main/scala/com/mesosphere/mesos/client/Session.scala
```diff
@@ -0,0 +1,171 @@
...
+      logger.error("Fetching the next token failed", ex)
+      throw ex
+    case _ => stash()
+  }
+
+  def initialized(credentials: HttpCredentials): Receive = {
+    case call: Array[Byte] =>
+      val request = requestFactory(call, Some(credentials))
+      val originalSender = sender()
+      logger.info("Processing next call.")
```
zen-dog:
```suggestion
      logger.debug("Processing next call.")
```

mesos-client/src/main/scala/com/mesosphere/mesos/client/Session.scala
```diff
@@ -0,0 +1,171 @@
...
+      // The TLS handshake for each connection might be an overhead. We could potentially reuse a connection.
+      Http()(context.system)
+        .singleRequest(request)
+        .onComplete {
+          case Success(response) =>
+            logger.info(s"HTTP response: ${response.status}")
```
zen-dog:
```suggestion
            logger.debug(s"HTTP response: ${response.status}")
```

mesos-client/src/main/scala/com/mesosphere/mesos/client/Session.scala
```diff
@@ -0,0 +1,171 @@
...
+            self ! SessionActor.Response(call, originalSender, response)
+          case Failure(ex) =>
+            logger.error("HTTP request failed", ex)
+            // Fail stream.
+            originalSender ! Status.Failure(ex)
+        }
+    case SessionActor.Response(originalCall, originalSender, response) =>
+      logger.info(s"Call replied with ${response.status}")
```
zen-dog:
```suggestion
      logger.debug(s"Call replied with ${response.status}")
```

zen-dog (approved):
Nice work! I've left a few nits, my biggest concern is around lack of tests for the `SessionActor`. Once this is addressed, we should be able to merge it!

mesos-client/src/main/scala/com/mesosphere/mesos/client/Session.scala
```diff
@@ -0,0 +1,171 @@
...
+  /**
+    * A connection flow factory that will create a flow that only processes one request at a time.
+    *
+    * It will reconnect if the server closes the connection after a successful request.
+    *
+    * @return A connection flow for single requests.
+    */
+  private def connection(implicit system: ActorSystem, mat: Materializer): Flow[HttpRequest, HttpResponse, NotUsed] = {
+    val poolSettings = ConnectionPoolSettings("").withMaxConnections(1).withPipeliningLimit(1)
+    if (isSecured) {
```
zen-dog: Wdyt about making this check inside the `via(Http()...` stage? It'll be less repetition of lines overall.

mesos-client/src/main/scala/com/mesosphere/mesos/client/MesosClient.scala
timcharper: We actually have another implementation where isSecured is used elsewhere.

I would prefer a shared method for detecting this, one that throws an exception or returns an empty optional if that string is neither "https" nor "http"

mesos-client/src/main/scala/com/mesosphere/mesos/client/MesosClient.scala
```diff
@@ -201,18 +218,20 @@ object MesosClient extends StrictLogging with StrictLoggingFlow {
               throw MesosRedirectException(leader)
             case _ =>
               response.discardEntityBytes()
-              throw new IllegalArgumentException(s"Mesos server error: $response")
+              import scala.concurrent.duration._
+              val error = Await.result(Unmarshal(response.entity).to[String], 1.minutes)
```
timcharper: I'm really concerned about this await.result in the middle of an active stream stage. Blocking insider the Akka dispatcher is explicitly discouraged.

Can we instead change the map to mapAsync(1), and have the subsequent values be either Future.successful (response, …) etc. perhaps we could use the async/await pattern on the `case _`

mesos-client/src/main/scala/com/mesosphere/mesos/client/Session.scala
```diff
@@ -0,0 +1,171 @@
...
+  *
+  * It behaves similar to Python Request's [[https://2.python-requests.org/en/master/user/advanced/#session-objects Session object]].
+  * Thus it provides methods to construct and connect to Mesos.
+  *
+  * @param url The Mesos master URL.
+  * @param streamId The Mesos stream ID. See the [[http://mesos.apache.org/documentation/latest/scheduler-http-api/#calls docs]] for details.
+  * @param authorization A [[CredentialsProvider]] if the connection is secured.
+  */
+case class Session(url: URL, streamId: String, authorization: Option[CredentialsProvider] = None) {
+  lazy val isSecured: Boolean = url.getProtocol == "https"
```
timcharper: We've done this a few times here, perhaps we should have a helper method?

mesos-client/src/main/scala/com/mesosphere/mesos/client/Session.scala
timcharper: Yikes, yeah, agreed. It doesn't seem like we want to re-instantiate this after every request.

Just in case if somebody is using a USI framework and has a really good reason to change this value, but doesn't have time to build and update the framework.

mesos-client/src/main/scala/com/mesosphere/mesos/client/Session.scala
timcharper: Also, let's have it configurable via an undocumented environment variable?

mesos-client/src/main/scala/com/mesosphere/mesos/client/Session.scala
```diff
@@ -0,0 +1,171 @@
...
+
+  /**
+    * A connection flow factory that will create a flow that only processes one request at a time.
+    *
+    * It will reconnect if the server closes the connection after a successful request.
+    *
+    * @return A connection flow for single requests.
+    */
+  private def connection(implicit system: ActorSystem, mat: Materializer): Flow[HttpRequest, HttpResponse, NotUsed] = {
+    val poolSettings = ConnectionPoolSettings("").withMaxConnections(1).withPipeliningLimit(1)
```
timcharper, line 68: Can we add a comment here on why connection pool settings receives an empty string?

mesos-client/src/test/scala/com/mesosphere/mesos/client/SessionActorTest.scala
timcharper: Yeah, I would really like to see an integration test which uses a mock IAM service, testing the following scenarios:

* Happy path (authentication works)
* semi-happy path (authentication fails initially because the token is invalid), a new token is fetched and the call in question is resubmitted.

timcharper (changes requested):

mesos-client/src/main/scala/com/mesosphere/mesos/client/MesosClient.scala
jeschkies: I've forgot to remove it. I will not add it in this PR.

mesos-client/src/main/scala/com/mesosphere/mesos/client/Session.scala
jeschkies: > So we're recreating the session actor for each request?? I don't think this is what we want.

We don't. Akka streams is a little confusing is this regard. Checkout the usage of `Session.post`. It is used in the `via` block to create the blue print of the flow. It is *not* called each time a request is processed, ie during materialization.

Stuff like this let's me sometimes question whether Akka streams was really the right choice.

mesos-client/src/main/scala/com/mesosphere/mesos/client/Session.scala
jeschkies: `reference.conf` it is.

mesos-client/src/main/scala/com/mesosphere/mesos/client/Session.scala
jeschkies: Ah nice. Thanks.

mesos-client/src/main/scala/com/mesosphere/mesos/client/Session.scala
zen-dog: It's a closure over the `sessionActor`! Got it

mesos-client/src/main/scala/com/mesosphere/mesos/client/MesosClient.scala
```diff
@@ -365,40 +395,23 @@ class MesosClientImpl(
...
-      response.status match {
-        case status if status.isFailure() =>
-          logger.info(s"A request to mesos failed with response: ${response}")
-          response.discardEntityBytes()
-          throw new IllegalStateException(s"Failed to send a call to mesos")
-        case _ =>
-          logger.debug(s"Mesos call response: $response")
-          response.discardEntityBytes()
+      if (response.status.isFailure()) {
+        logger.info(s"A request to Mesos failed with response: ${response.status}")
```
zen-dog, line 396: !!!
```suggestion
        logger.info(s"A request to Mesos failed with response: ${response.status}")
        response.discardEntityBytes()
```

mesos-client/src/main/scala/com/mesosphere/mesos/client/Session.scala
```diff
@@ -0,0 +1,184 @@
...
+    * A connection flow factory that will create a flow that only processes one request at a time.
+    *
+    * It will reconnect if the server closes the connection after a successful request.
+    *
+    * @return A connection flow for single requests.
+    */
+  private def connection(implicit system: ActorSystem, mat: Materializer): Flow[HttpRequest, HttpResponse, NotUsed] = {
+    // Constructs the connection pool settings with defaults and overrides the max connections and pipelining limit so
+    // that only one request at a time is processed. See https://doc.akka.io/docs/akka-http/current/configuration.html
+    // for details.
```
zen-dog, line 66:
```suggestion
    // for details.
    // *IMPORTANT*: DO NOT CHANGE maxConnections OR pipeliningLimit! Otherwise, USI won't guarantee request order to Mesos!
```

mesos-client/src/main/scala/com/mesosphere/mesos/client/Session.scala
```diff
@@ -0,0 +1,184 @@
...
+  override def preStart(): Unit = {
+    super.preStart()
+    credentialsProvider.nextToken().pipeTo(self)
+  }
+
+  override def receive: Receive = initializing
+
+  def initializing: Receive = {
+    case credentials: HttpCredentials =>
+      logger.info("Retrieved token")
```
zen-dog:
```suggestion
      logger.info("Retrieved IAM session token")
```

mesos-client/src/main/scala/com/mesosphere/mesos/client/Session.scala
```diff
@@ -0,0 +1,184 @@
...
+      context.become(initialized(credentials))
+      unstashAll()
+    case Status.Failure(ex) =>
+      logger.error("Fetching the next token failed", ex)
```
zen-dog:
```suggestion
      logger.error("Fetching the next IAM session token failed", ex)
```

mesos-client/src/main/scala/com/mesosphere/mesos/client/Session.scala
```diff
@@ -0,0 +1,184 @@
...
+      throw ex
+    case _ => stash()
+  }
+
+  def initialized(credentials: HttpCredentials): Receive = {
+    case call: Array[Byte] =>
+      val request = requestFactory(call, Some(credentials))
+      val originalSender = sender()
+      logger.debug("Processing next call.")
```
zen-dog:
```suggestion
      logger.debug("Processing next Mesos call.")
```

mesos-client/src/main/scala/com/mesosphere/mesos/client/Session.scala
```diff
@@ -0,0 +1,184 @@
...
+      // The TLS handshake for each connection might be an overhead. We could potentially reuse a connection.
+      Http()(context.system)
+        .singleRequest(request)
+        .onComplete {
+          case Success(response) =>
+            logger.debug(s"HTTP response: ${response.status}")
+            self ! SessionActor.Response(call, originalSender, response)
+          case Failure(ex) =>
+            logger.error("HTTP request failed", ex)
```
zen-dog: Which HTTP request?

mesos-client/src/main/scala/com/mesosphere/mesos/client/Session.scala
```diff
@@ -0,0 +1,184 @@
...
+    case call: Array[Byte] =>
+      val request = requestFactory(call, Some(credentials))
+      val originalSender = sender()
+      logger.debug("Processing next call.")
+      // The TLS handshake for each connection might be an overhead. We could potentially reuse a connection.
+      Http()(context.system)
+        .singleRequest(request)
+        .onComplete {
+          case Success(response) =>
+            logger.debug(s"HTTP response: ${response.status}")
```
zen-dog: What HTTP response?

mesos-client/src/main/scala/com/mesosphere/mesos/client/Session.scala
```diff
@@ -0,0 +1,184 @@
...
+            self ! SessionActor.Response(call, originalSender, response)
+          case Failure(ex) =>
+            logger.error("HTTP request failed", ex)
+            // Fail stream.
+            originalSender ! Status.Failure(ex)
+        }
+    case SessionActor.Response(originalCall, originalSender, response) =>
+      logger.debug(s"Call replied with ${response.status}")
+      if (response.status == StatusCodes.Unauthorized) {
+        logger.info("Refreshing token")
```
zen-dog: What token?

mesos-client/src/test/scala/com/mesosphere/mesos/client/SessionActorTest.scala
```diff
@@ -0,0 +1,124 @@
...
+    *
+    * All request respond with the number of calls the endpoint received. Each response takes the head of [[responseCodes]].
+    * If the list is exhausted the stub will respond with an HTTP 501 error.
+    *
+    * @param responseCodes The HTTP codes the stub will use for [[responseCodes.length]] requests.
+    */
+  case class MesosStub(var responseCodes: List[StatusCode]) {
+    import akka.http.scaladsl.server.Directives._
+
+    var calls: Int = 0
```
zen-dog: This should be either `AtomicInteger` or `@volatile`

mesos-client/src/test/scala/com/mesosphere/mesos/client/SessionActorTest.scala
```diff
@@ -0,0 +1,124 @@
...
+
+    val route =
+      path("mesos") {
+        post {
+          calls += 1
+          responseCodes match {
+            case head :: tail =>
+              responseCodes = tail
```
zen-dog, line 132: Here too, you're manipulating state from a different context (the routes). You could make `responseCodes` a local var marked as `@volatile`

mesos-client/src/test/scala/com/mesosphere/mesos/client/SessionActorTest.scala
```diff
@@ -0,0 +1,124 @@
...
+      val sessionActor =
+        system.actorOf(SessionActor.props(BasicAuthenticationProvider("user", "password"), mesos.createRequest))
+
+      When("we make a call through the session actor")
+      val response = (sessionActor ? Array.empty[Byte]).futureValue.asInstanceOf[HttpResponse]
+
+      Then("we receive an HTTP 200")
+      response.status should be(StatusCodes.OK)
+
+      And("Mesos was called only once")
```
zen-dog:
```suggestion
      And("Mesos was called twice: one declined call and one with the fresh token")
```

mesos-client/src/test/scala/com/mesosphere/mesos/client/SessionActorTest.scala
```diff
@@ -0,0 +1,124 @@
...
+      Unmarshal(response.entity).to[String].futureValue.toInt should be(2)
```
zen-dog, line 72: It's somewhat paranoid but can we make sure that the credentials provider has indeed issued a second token?

mesos-client/src/test/scala/com/mesosphere/mesos/client/SessionActorTest.scala
jeschkies: I though Akka HTTP is processing a request at a time but will do.

zen-dog: It doesn't matter how many requests at a time - the important part is that processing happens in a different execution context. The second request might not see the changes done by the first.

mesos-client/src/test/scala/com/mesosphere/mesos/client/SessionActorTest.scala
```diff
@@ -0,0 +1,146 @@
...
+  class Fixture() {
+
+    def requestFactory(body: Array[Byte], credentials: Option[HttpCredentials]): HttpRequest = ???
+  }
+
+  /**
+    * A simple [[CredentialsProvider]] stub that counts how many times [[CredentialsProvider.nextToken()]] was called.
+    */
+  class CountingCredentialsProvider() extends CredentialsProvider {
+    var calls: Int = 0
```
zen-dog:
```suggestion
    @volatile var calls: Int = 0
```

zen-dog (approved):
🚢 🎉 🆒

mesos-client/src/test/scala/com/mesosphere/mesos/client/SessionActorTest.scala
```diff
@@ -0,0 +1,146 @@
...
+  class Fixture() {
+
+    def requestFactory(body: Array[Byte], credentials: Option[HttpCredentials]): HttpRequest = ???
+  }
+
+  /**
+    * A simple [[CredentialsProvider]] stub that counts how many times [[CredentialsProvider.nextToken()]] was called.
+    */
+  class CountingCredentialsProvider() extends CredentialsProvider {
+    @volatile var calls: Int = 0
```
jeschkies, line 88: I don't think this was required but fine.

timcharper (approved):

mesos-client/src/test/scala/com/mesosphere/mesos/client/SessionActorTest.scala
zen-dog: It was!

mesos-client/src/main/scala/com/mesosphere/mesos/client/CredentialsProvider.scala
```diff
@@ -0,0 +1,97 @@
...
+
+  case class SessionToken(token: String)
+  implicit val sessionRead: Reads[SessionToken] = (JsPath \ "token").read[String].map(SessionToken)
+
+  /** @return a new request for a session token. */
+  def acsTokenRequest: HttpRequest = {
+
+    // Tokenize the claim.
+    val token = JwtJson.encode(claim, privateKey, RS256)
+    val data: String = Json.stringify(Json.obj("uid" -> uid, "token" -> token))
```
jgehrcke, line 71: Hey was this work and code based on https://wiki.mesosphere.com/pages/viewpage.action?pageId=123209735?

I would appreciate the usage of the following terms:
- "DC/OS authentication token"
- "service login token"

for consistency, and for easier code readability. 

What is a session token here?

jgehrcke: The usage of "authenticate" in the code comments here probably means "log in".

The "login" is the process for exchanging primary credentials (e.g. service account credentials) into secondary credentials (DC/OS authentication token).

jeschkies: Thanks for the hint. Please see my follow up https://github.com/mesosphere/usi/pull/99.

PR #95, opened 2019-05-29
Simplify Mesos client example.

jeschkies:
Summary:
This will materialize the Mesos call flow only once.

### DIFF
--- a/examples/mesos-client-example/src/main/scala/com/mesosphere/mesos/examples/MesosClientExampleFramework.scala
+++ b/examples/mesos-client-example/src/main/scala/com/mesosphere/mesos/examples/MesosClientExampleFramework.scala
@@ -44,25 +44,25 @@ class MesosClientExampleFramework(settings: MesosClientSettings) extends StrictL
   val client = Await.result(MesosClient(settings, frameworkInfo).runWith(Sink.head), 10.seconds)
 
   client.mesosSource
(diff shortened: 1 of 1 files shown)


examples/mesos-client-example/src/main/scala/com/mesosphere/mesos/examples/MesosClientExampleFramework.scala
```diff
@@ -43,26 +43,23 @@ class MesosClientExampleFramework(settings: MesosClientSettings) extends StrictL
...
-                filters = Some(Filters.newBuilder().setRefuseSeconds(5.0).build())
-            ))
-          .runWith(client.mesosSink)
-      }
-
-    })
+  client.mesosSource.mapConcat { event =>
+    if (event.getType == Event.Type.SUBSCRIBED) {
+      logger.info("Successfully subscribed to mesos")
+      List.empty
```
zen-dog, line 49: You could do it like this. However, IMHO the code on the left side, though less optimized, captures the idea of "event loop reacting to events" better than the one on the right side.

jeschkies: Kind of. Anyhow, this example did not expose an issue in `outgoingConnection` which is closed when the server sends the `Connection: close` header. The fix will be part of #94.

PR #96, opened 2019-05-29
Notify downstream termination for scheduler flow

timcharper:
We were using fromSinkAndSource, which does not notify the flow downstream
(source) if the upstream (sink) receives a failure. Switching to
fromSinkAndSourceCoupled resolves this.


zen-dog (approved):
👍

meln1k (approved):

PR #99, opened 2019-06-13
Rename tokens.

jeschkies:
Summary:
Follow token names described in [wiki](https://wiki.mesosphere.com/pages/viewpage.action?pageId=123209735).

### DIFF
--- a/mesos-client/src/main/scala/com/mesosphere/mesos/client/CredentialsProvider.scala
+++ b/mesos-client/src/main/scala/com/mesosphere/mesos/client/CredentialsProvider.scala
@@ -61,7 +61,8 @@ case class DcosServiceAccountProvider(uid: String, privateKey: String, root: URL
   private def claim: JsObject = Json.obj("uid" -> uid, "exp" -> expireIn(DcosServiceAccountProvider.CLAIM_LIFETIME))
 
   case class AuthenticationToken(token: String)
-  implicit val authenticationTokenRead: Reads[AuthenticationToken] = (JsPath \ "token").read[String].map(AuthenticationToken)
+  implicit val authenticationTokenRead: Reads[AuthenticationToken] =
+    (JsPath \ "token").read[String].map(AuthenticationToken)
 
   /** @return a new request for an authentication token. */
   def loginRequest: HttpRequest = {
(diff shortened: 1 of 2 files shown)


zen-dog (approved):

jeschkies:
ping @jgehrcke

jgehrcke:
pong! looking :)

mesos-client/src/main/scala/com/mesosphere/mesos/client/CredentialsProvider.scala
```diff
@@ -60,15 +60,16 @@ case class DcosServiceAccountProvider(uid: String, privateKey: String, root: URL
...
 
-  /** @return a new request for a session token. */
-  def acsTokenRequest: HttpRequest = {
+  /** @return a new request for an authentication token. */
+  def loginRequest: HttpRequest = {
 
     // Tokenize the claim.
-    val token = JwtJson.encode(claim, privateKey, RS256)
-    val data: String = Json.stringify(Json.obj("uid" -> uid, "token" -> token))
+    val serviceLoginToken = JwtJson.encode(claim, privateKey, RS256)
```
jgehrcke, line 72: lovely :)

mesos-client/src/main/scala/com/mesosphere/mesos/client/CredentialsProvider.scala
```diff
@@ -60,15 +60,16 @@ case class DcosServiceAccountProvider(uid: String, privateKey: String, root: URL
     */
   private def claim: JsObject = Json.obj("uid" -> uid, "exp" -> expireIn(DcosServiceAccountProvider.CLAIM_LIFETIME))
```
jgehrcke: `DcosServiceAccountProvider.CLAIM_LIFETIME` is probably defined far away from this line. The name is a bit cryptic. This defines the lifetime of the service login token. Can we rename that variable to something more expressive (`SERVICE_LOGIN_TOKEN_LIFETIME`) or maybe just write an explicit `20` right here? Then it's also clear that this needs to be kept in sync with the code comment `@return a claim for the user that expires in twenty seconds.`.

Just a nit of course .. :-)

jgehrcke (approved):
Some more remarks, but no expectations :-)

Thanks for working on this at all!

jgehrcke:
Thank you @jeschkies -- very cool! :rocket:

PR #101, opened 2019-09-05
Start USI cli.

jeschkies:
Summary:
The CLI consumer scheduler commands via stdin and outputs pod events via
stdout. nput and output should be encoded in Server Sent Events.


jeschkies:
Won't do.

PR #102, opened 2019-09-06
Bootstrap gRPC based USI service.

jeschkies:
Summary:
The service expose the raw USI input and output streams.


jeschkies:
Won't do.

PR #103, opened 2019-09-13
Introduce and use version script

timcharper:
This version script aids with auto-versionining USI by counting the number of commits to master since some point in the history.

Commits not-yet-merged to master are versioned by counting the number of
commits since their merge-base (the point of deviation); further, commits
not-yet-merged are treated as snapshot. If origin/master is at version 0.1.15,
and tharper/some-new-feature deviated two commits before origin/master, it's
version will be 0.1.13-deadbee-SNAPSHOT.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -15,10 +15,20 @@ ext {
     akkaVersion = '2.5.19'
 }
 
+def readCurrentVersion = {
+  def out = new ByteArrayOutputStream()
+
+  exec{
+      commandLine "./version"
+      standardOutput = out;
+  }
+  out.toString().replaceAll("\\s", "");
+}
+
 allprojects {
 
     group = 'com.mesosphere.usi'
-    version = '0.1'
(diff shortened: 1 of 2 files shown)


build.gradle
```diff
@@ -15,10 +15,20 @@ ext {
     akkaVersion = '2.5.19'
 }
 
+def readCurrentVersion = {
+  def out = new ByteArrayOutputStream()
+
+  exec{
+      commandLine "./version"
```
jeschkies, line 22: I was just wondering why we are not using Scala but I guess the `version` script is from Marathon.

jeschkies (approved):
Thanks. Should we also improve the Readme?

ANeumann82 (approved):

build.gradle
timcharper: It's very useful to have the script, IMO. Experience has informed me that moving all of our build automation to Scala is undesirable

PR #106, opened 2019-10-02
Add RunTemplate and preliminary support for TaskGroup

timcharper:


core-models/src/main/scala/com/mesosphere/usi/core/models/Ids.scala
```diff
@@ -6,6 +6,16 @@ package com.mesosphere.usi.core.models
   */
 case class PodId(value: String)
 
+/**
+  * Identifier for a pod's task; does not include podId
```
jeschkies: Could you give an example?

core-models/src/main/scala/com/mesosphere/usi/core/models/RunTemplateLike.scala
```diff
@@ -0,0 +1,134 @@
...
+
+sealed trait RunTemplateLike {
+  final lazy val allResources: Seq[ResourceRequirement] = {
+    val b = Seq.newBuilder[ResourceRequirement]
+    taskResourceRequirements.values.foreach { r =>
+      r.foreach(b += _)
+    }
+    executorResources.foreach { b += _ }
+    b.result()
+  }
```
jeschkies: `allResources = taskResourceRequirements.valueIterator.toVector` should do exactly that.

core-models/src/main/scala/com/mesosphere/usi/core/models/RunTemplateLike.scala
```diff
@@ -0,0 +1,134 @@
...
+  val taskResourceRequirements: Map[PartialTaskId, Seq[ResourceRequirement]]
+  val executorResources: Seq[ResourceRequirement]
+  val role: String
+  private[usi] def buildOperation(
```
jeschkies: What do you think about
```suggestion
  private[usi] def buildLaunchOperation(
```

core-models/src/main/scala/com/mesosphere/usi/core/models/RunTemplateLike.scala
```diff
@@ -0,0 +1,134 @@
...
+class LaunchGroupBuilder(val role: String, executorBuilder: ExecutorBuilderLike, tasks: Map[PartialTaskId, TaskBuilderLike])
+    extends RunTemplateLike
+    with StrictLogging {
+  val taskResourceRequirements = tasks.map { case (taskId, taskBuilder) => taskId -> taskBuilder.resourceRequirements }
+
+  override val executorResources: Seq[ResourceRequirement] = Nil
+
+  override private[usi] def buildOperation(
+      matchedOffer: Mesos.Offer,
+      taskIdStrategy: CurriedPodTaskIdStrategy,
```
jeschkies: If I understand correctly the user can define how tasks are named. Should we allow this flexibility?

core-models/src/main/scala/com/mesosphere/usi/core/models/Ids.scala
```diff
@@ -6,6 +6,16 @@ package com.mesosphere.usi.core.models
...
+  * @param value
+  */
+case class PartialTaskId(value: String)
```
jeschkies: What is the relation between a `TaskId` and a `PartialTaskId`?

core-models/src/main/scala/com/mesosphere/usi/core/models/RunTemplateLike.scala
```diff
@@ -0,0 +1,134 @@
...
+  val role: String
+  private[usi] def buildOperation(
+      matchedOffer: Mesos.Offer,
+      taskIdStrategy: CurriedPodTaskIdStrategy,
+      executorResources: Seq[Mesos.Resource],
+      taskResources: Map[PartialTaskId, Seq[Mesos.Resource]])
+    : Either[Mesos.Offer.Operation.Launch, Mesos.Offer.Operation.LaunchGroup]
+}
+
+class LegacyLaunchBuilder(val role: String, taskBuilder: TaskBuilderLike) extends RunTemplateLike {
```
jeschkies: This is a little confusing to me. Maybe it's the naming but why would a builder extend a template?

jeschkies (commented):
Thanks. I have some open questions.

core-models/src/main/scala/com/mesosphere/usi/core/models/RunTemplateLike.scala
jeschkies: How about making the build a field of the template?

core-models/src/main/scala/com/mesosphere/usi/core/models/Ids.scala
timcharper: Added docs

timcharper: And tests

core-models/src/main/scala/com/mesosphere/usi/core/models/RunTemplateLike.scala
timcharper: Sure

core-models/src/main/scala/com/mesosphere/usi/core/models/Ids.scala
timcharper: Clarified this by renaming to TaskName.

Also, docs.

core-models/src/main/scala/com/mesosphere/usi/core/models/RunTemplateLike.scala
timcharper: Yeah, WIP PR did not include attention to naming :) I've renamed it. I think this is an ok use of inheritance but am happy to hear your ideas on how composition would be better here.

core-models/src/main/scala/com/mesosphere/usi/core/linting/Linting.scala
```diff
@@ -0,0 +1,47 @@
+package com.mesosphere.usi.core.linting
+
+import com.mesosphere.usi.core.models.resources.{ExactValue, RangeRequirement, ResourceRequirement, ScalarRequirement}
+import com.mesosphere.usi.core.models.template.RunTemplate
+import com.mesosphere.usi.core.models.template.RunTemplate.KeyedResourceRequirement
+
+class Linting {
```
jeschkies, line 7: This is not a blocker IMHO but what to you think about using Scalactic's [Or and Every](http://www.scalactic.org/user_guide/OrAndEvery)? We probably should decide for a validation pattern earlier than later.

core-models/src/main/scala/com/mesosphere/usi/core/models/Ids.scala
```diff
@@ -6,6 +6,17 @@ package com.mesosphere.usi.core.models
   */
 case class PodId(value: String)
 
+/**
+  * Name for a pod's task. The TaskName and PodId are combined via PodTaskIdStrategy to generate a TaskId.
```
jeschkies, line 10: Nice. So in Marathon world we might have `myGroup_myApp.marathon-b6ff5fa5-7714-11e7-a55c-5ecf1c4671f6._app.1`.
*  `myGroup_myApp.marathon-b6ff5fa5-7714-11e7-a55c-5ecf1c4671f6` is the instance id aka pod id in USI
* `_app.1` is the task name
* both combined are the task id

Did I get this right?

core-models/src/main/scala/com/mesosphere/usi/core/models/PodTaskIdStrategy.scala
```diff
@@ -0,0 +1,39 @@
...
+      else {
+        if (podId.value.contains(".")) {
+          logger.error(
+            s"podId ${podId.value} contains a '.', which is the delimiter character; this results in an unparseable taskId.")
+        }
+        if (taskName.value.contains(".")) {
+          logger.error(
+            s"taskName ${taskName.value} for podId ${podId.value} contains a '.', which is the delimiter character; this results in an unparseable taskId.")
+        }
+        TaskId(podId.value + "." + taskName.value)
```
jeschkies, line 30: This is probably too much for now but I think `TaskId` should be a trait so that we can define `case class DefaultTaskId(podId: PodId, taskName: TaskName)`. It's always convenient to extract them again without parsing.

core-models/src/main/scala/com/mesosphere/usi/core/models/RunningPodSpec.scala
```diff
@@ -31,45 +31,4 @@ case class RunningPodSpec(id: PodId, runSpec: RunTemplate) extends PodSpec {
...
-    }
-  }
-
-  def isValid(runSpec: RunTemplate): Seq[ValidationMessage] = {
-    val uniqueRangeRequirements = validateStaticRangeRequirementsUnique(runSpec)
-    val scalarRequirements = validateScalarRequirements(runSpec.resourceRequirements)
-    uniqueRangeRequirements ++ scalarRequirements
-  }
-}
+object RunningPodSpec {}
```
jeschkies: Where is this used?

core-models/src/main/scala/com/mesosphere/usi/core/models/TaskBuilder.scala
```diff
@@ -0,0 +1,12 @@
+package com.mesosphere.usi.core.models
+
+import com.mesosphere.usi.core.models.resources.ResourceRequirement
+import org.apache.mesos.v1.{Protos => Mesos}
+
+trait TaskBuilder {
```
jeschkies, line 10: Could you add docs since this is user facing?

core-models/src/main/scala/com/mesosphere/usi/core/models/template/SimpleRunTemplateFactory.scala
```diff
@@ -0,0 +1,87 @@
...
+
+  case class SimpleTaskInfoBuilder(
+      resourceRequirements: Seq[ResourceRequirement],
+      shellCommand: String,
+      role: String,
+      fetch: Seq[FetchUri] = Seq.empty,
+      dockerImageName: Option[String] = None)
+      extends TaskBuilder {
+
+    override def buildTask(
```
jeschkies, line 28: I wonder if we could/should make this private to `[usi]`. What do you think?

core-models/src/main/scala/com/mesosphere/usi/core/linting/Linting.scala
```diff
@@ -0,0 +1,47 @@
...
+    }.flatten.collect { case ExactValue(value) => value }
+
+    if (staticPorts.distinct.length != staticPorts.length) {
+      Seq(s"Every value inside RangeResource can be requested only once. Requirement: ${staticPorts.mkString(",")}")
+    } else {
+      Nil
+    }
+  }
+
+  private def validateScalarRequirements(resourceRequirements: Seq[ResourceRequirement]): Seq[LintingWarning] = {
```
ANeumann82, line 29: Nit: Maybe use the same signature as above, and pass in the full runTemplate instead of only the resourceRequirements?

Also: Should we validate that each scalar is only required once here, or is that done somewhere else? I.e. two requirements for "cpu" should be invalid

core-models/src/main/scala/com/mesosphere/usi/core/models/template/RunTemplate.scala
```diff
@@ -0,0 +1,129 @@
...
+  private[usi] def buildLaunchOperation(
+      matchedOffer: Mesos.Offer,
+      taskIdStrategy: CurriedPodTaskIdStrategy,
+      executorResources: Seq[Mesos.Resource],
+      taskResources: Map[TaskName, Seq[Mesos.Resource]])
+    : Either[Mesos.Offer.Operation.Launch, Mesos.Offer.Operation.LaunchGroup]
+}
+
+object RunTemplate extends StrictLogging {
+  case class KeyedResourceRequirement(entityKey: Option[TaskName], requirement: ResourceRequirement)
```
ANeumann82, line 40:
```suggestion
  /**
   * Used to identify to which task a requirement belongs. An empty entityKey indicates that the requirement belongs to the executor 
   **/
  case class KeyedResourceRequirement(entityKey: Option[TaskName], requirement: ResourceRequirement)
```

core-models/src/main/scala/com/mesosphere/usi/core/models/template/RunTemplate.scala
```diff
@@ -0,0 +1,129 @@
...
+
+    Right(launchGroup)
+  }
+}
+
+class LegacyLaunchRunTemplate(val role: String, taskBuilder: TaskBuilder) extends RunTemplate {
+  final override val taskResourceRequirements: Map[TaskName, Seq[ResourceRequirement]] =
+    Map(TaskName.empty -> taskBuilder.resourceRequirements)
+
+  final override val executorResources: Seq[ResourceRequirement] = Nil
```
ANeumann82, line 116: Shouldn't this be an empty seq? In the code above we access with with .foreach without checking for Nil...

```suggestion
  final override val executorResources: Seq[ResourceRequirement] = Seq.empty
```

core/src/main/scala/com/mesosphere/usi/core/logic/MesosEventsLogic.scala
```diff
@@ -19,25 +18,38 @@ import scala.collection.JavaConverters._
 private[core] class MesosEventsLogic(mesosCallFactory: MesosCalls, offerMatcher: OfferMatcher = new FCFSOfferMatcher())
     extends ImplicitStrictLogging {
 
+  val podTaskIdStrategy: PodTaskIdStrategy = PodTaskIdStrategy.DefaultStrategy
```
ANeumann82, line 21: This will be settable later on, so we can use different strategies for e.g. legacy metronome IDs, right?

core/src/main/scala/com/mesosphere/usi/core/logic/MesosEventsLogic.scala
```diff
@@ -121,28 +114,34 @@ private[core] class MesosEventsLogic(mesosCallFactory: MesosCalls, offerMatcher:
...
+            }
+
+            SchedulerEvents(
+              stateEvents = List(PodStatusUpdatedEvent(podId, Some(newStatus))),
+              mesosCalls = if (taskStatus.hasUuid) {
+                // frameworks should accept only status updates that have UUID set
+                // http://mesos.apache.org/documentation/latest/scheduler-http-api/#acknowledge
+                List(mesosCallFactory.newAcknowledge(taskStatus.getAgentId, taskStatus.getTaskId, taskStatus.getUuid))
+              } else {
+                Nil
```
ANeumann82, line 137: Do we need a warning or error log here? Or can we safely ignore it?

ANeumann82 (approved):
Nothing serious, nice work

core-models/src/main/scala/com/mesosphere/usi/core/models/RunTemplateLike.scala
timcharper: This has been replaced with a Map, and, please see this included the executorResources also

core-models/src/main/scala/com/mesosphere/usi/core/models/Ids.scala
timcharper: Yeah, that's right.

Also, the PodTaskIdStrategy will be pluggable so that frameworks can customize how this maps

core-models/src/main/scala/com/mesosphere/usi/core/models/PodTaskIdStrategy.scala
timcharper: Ah, interesting point!

core-models/src/main/scala/com/mesosphere/usi/core/models/RunningPodSpec.scala
timcharper: Not sure if IntelliJ added this for me or if I was going to add a method and then quit :) I'll clean up

core/src/main/scala/com/mesosphere/usi/core/logic/MesosEventsLogic.scala
timcharper: Yeah, that's the idea. Not all frameworks will have the same opinion about how task names in a taskGroup related to a taskId.

core/src/main/scala/com/mesosphere/usi/core/logic/MesosEventsLogic.scala
timcharper: No, this is the expected way for frameworks to handle task status updates; acknowledge only needed if status has a uuid

core-models/src/main/scala/com/mesosphere/usi/core/models/template/RunTemplate.scala
timcharper: Nil is List.empty which implements Seq

core-models/src/main/scala/com/mesosphere/usi/core/models/template/SimpleRunTemplateFactory.scala
timcharper: We cannot make it private; it's a publicly implementable trait method

core-models/src/main/scala/com/mesosphere/usi/core/linting/Linting.scala
timcharper: This is just moved code (out of Mesos Event Logic and in to something that can be optionally used if wanted). I'd rather not deal with it in detail right now.

core-models/src/main/scala/com/mesosphere/usi/core/linting/Linting.scala
timcharper: As mentioned, this code was moved our of MesosEventLogic. I think I'd prefer to not do a lot with it right now, other than to move it, since much of it doesn't apply in the RunTemplate world.

core-models/src/main/scala/com/mesosphere/usi/core/models/template/RunTemplate.scala
```diff
@@ -0,0 +1,133 @@
...
+   * Used to identify to which task a requirement belongs. An empty entityKey indicates that the requirement belongs to the executor 
+   **/
+  case class KeyedResourceRequirement(entityKey: Option[TaskName], requirement: ResourceRequirement)
+
+  private[models] def setTaskInfo(
+      b: Mesos.TaskInfo.Builder,
+      matchedOffer: Mesos.Offer,
+      taskName: TaskName,
+      taskIdStrategy: CurriedPodTaskIdStrategy,
+      resources: Seq[Mesos.Resource]): Unit = {
```
jeschkies, line 47: Are these the matched resources?

core/src/main/scala/com/mesosphere/usi/core/logic/MesosEventsLogic.scala
```diff
@@ -48,56 +60,37 @@ private[core] class MesosEventsLogic(mesosCallFactory: MesosCalls, offerMatcher:
...
-      )
-      logger.info(
-        s"Launching taskId${if (op.getLaunch.getTaskInfosCount > 1) "s"} : [{}] for offerId {}",
-        op.getLaunch.getTaskInfosList.asScala.map(_.getTaskId.getValue).mkString(", "),
-        offer.getId.getValue
-      )(
-        LoggingArgs("offerId" -> offer.getId.getValue).and("mesosOperation" -> "LAUNCH")
-      )
+      val operations = launchCommands.values.map {
+        case Right(launchGroup) =>
```
jeschkies, line 64: Nice!

jeschkies (approved):

PR #107, opened 2019-10-02
Port Mesos cluster fixes from Marathon.

jeschkies:
Summary:
This is a port mesosphere/marathon#7057 to USI. We also set fixed ports for the Zookeeper
server. The random ports were causing a flakiness in the Jenkins plugin.

### DIFF
--- a/test-utils/src/main/scala/com/mesosphere/utils/mesos/MesosTest.scala
+++ b/test-utils/src/main/scala/com/mesosphere/utils/mesos/MesosTest.scala
@@ -338,7 +338,9 @@ case class MesosCluster(
         s"--port=$port",
         s"--resources=${resources.resourceString()}",
         s"--master=$masterUrl",
-        s"--work_dir=${workDir.getAbsolutePath}") ++ extraArgs,
+        s"--work_dir=${workDir.getAbsolutePath}",
+        s"--cgroups_root=mesos$port", // See MESOS-9960 for more info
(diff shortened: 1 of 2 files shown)


test-utils/src/main/scala/com/mesosphere/utils/mesos/MesosTest.scala
```diff
@@ -338,7 +338,9 @@ case class MesosCluster(
         s"--port=$port",
         s"--resources=${resources.resourceString()}",
         s"--master=$masterUrl",
-        s"--work_dir=${workDir.getAbsolutePath}") ++ extraArgs,
+        s"--work_dir=${workDir.getAbsolutePath}",
+        s"--cgroups_root=mesos$port", // See MESOS-9960 for more info
+        s"""--executor_environment_variables={"GLOG_v": "2"}""") ++ extraArgs,
```
zen-dog: This particular line was needed to test executor failures in Marathon. Not sure if you want to have it active or just commented out for the rainy day.

zen-dog (approved):

test-utils/src/main/scala/com/mesosphere/utils/mesos/MesosTest.scala
timcharper: Is it super noisy to enable? Let's comment it out if so

timcharper (approved):

PR #108, opened 2019-10-07
Do not close namespaced Zookeeper client.

jeschkies:
Summary:
This fixes the USI test utils migration in mesosphere/marathon#7084. The
original fixture in Marathon would not close the namespaced client but
the underlying Zookeeper client.


zen-dog (approved):
🚢

ANeumann82 (approved):

PR #109, opened 2019-10-08
Set restricted roles for Mesos.

jeschkies:
Summary:
This blocks mesosphere/marathon#7084.


test-utils/src/main/scala/com/mesosphere/utils/mesos/MesosTest.scala
```diff
@@ -328,6 +328,7 @@ case class MesosCluster(
       "MESOS_SYSTEMD_ENABLE_SUPPORT" -> "false",
       "MESOS_SWITCH_USER" -> "false"
     ) ++
+      config.restrictedToRoles.map("MESOS_ROLES" -> _).to[Seq] ++
```
zen-dog, line 327: So, you probably want to remove `"MESOS_ROLES" -> "public,test"` line 325 above?

jeschkies: Man, I was wondering how our integration tests worked 🙈 Thanks.

ANeumann82 (approved):

zen-dog (approved):

PR #113, opened 2019-10-10
Configure Mesos role for keep-alive-framework.

jeschkies:
Summary:
The example framework cannot run on strict because we hard coded the
Mesos role. With this patch it becomes optional. We also introduce a
command line argument parser.

### DIFF
--- a/examples/keep-alive-framework/src/main/scala/com/mesosphere/usi/helloworld/KeepAliveFramework.scala
+++ b/examples/keep-alive-framework/src/main/scala/com/mesosphere/usi/helloworld/KeepAliveFramework.scala
@@ -95,17 +95,16 @@ class KeepAliveFramework(settings: KeepAliveFrameWorkSettings, authorization: Op
       +-----------------------+
 
     */
-  source
+  val end = source
     .via(keepAliveWatcher)
     .prepend(Source(specsSnapshot.map { spec =>
       LaunchPod(spec.id, spec.runSpec)
     }))
-    .to(sink)
-    .run()
+    .runWith(sink)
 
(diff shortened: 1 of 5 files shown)


examples/keep-alive-framework/src/main/scala/com/mesosphere/usi/helloworld/KeepAliveFramework.scala
```diff
@@ -95,64 +95,111 @@ class KeepAliveFramework(settings: KeepAliveFrameWorkSettings, authorization: Op
...
       LaunchPod(spec.id, spec.runSpec)
     }))
-    .to(sink)
-    .run()
+    .runWith(sink)
 
   // We let the framework run "forever"
-  io.StdIn.readLine("Keep alive framework is started")
-
+  val result = Await.result(end, Duration.Inf)
```
jeschkies, line 106: I was hoping that the stream would finish with an error when the client cannot connect but it keeps running. `asSourceAndSink` seems to be broken or I am missing something obvious. @timcharper, how can we let the framework crash on a connection error etc?

ANeumann82 (approved):

examples/keep-alive-framework/src/main/scala/com/mesosphere/usi/helloworld/KeepAliveFramework.scala
timcharper: Hmm, we have unit tests specifically for this (`asSourceAndSink` error propagation). Maybe something else is going on?

timcharper (approved):

PR #117, opened 2019-10-25
Region awareness support

timcharper:
We add support for region awareness in USI. USI will detect the region
of the Mesos master and default to launching pods in that region. A
generic, implementable DomainSelector is implemented with default
implementations to match the home region and to match a specific region.

JIRA Issues: DCOS_OSS-5628


core/src/main/scala/com/mesosphere/usi/core/matching/OfferMatcher.scala
```diff
@@ -1,35 +1,80 @@
...
 
-/**
-  * Interface for the offer matcher implementations. Given a Mesos [[Mesos.Offer]] and a set of [[RunningPodSpec]]s
-  * return a map of all matched PodSpecs with the list of matched Mesos resources.
-  */
-trait OfferMatcher {
+import scala.annotation.tailrec
+import scala.collection.JavaConverters._
+
+class OfferMatcher(masterDomainInfo: Mesos.DomainInfo) {
```
jeschkies, line 10: What was the motivation to move the implementation here?

core-models/src/main/scala/com/mesosphere/usi/core/models/faultdomain/DomainSelector.scala
```diff
@@ -0,0 +1,7 @@
+package com.mesosphere.usi.core.models.faultdomain
+
+import org.apache.mesos.v1.Protos.DomainInfo
+
+trait DomainSelector {
```
jeschkies: I wonder if this is rather a matcher. And if so maybe we could unify the interfaces for matching. What do you think?

core/src/main/scala/com/mesosphere/usi/core/matching/OfferMatcher.scala
timcharper: It was premature generalization.

core-models/src/main/scala/com/mesosphere/usi/core/models/faultdomain/DomainSelector.scala
timcharper: It's a special kind of matcher which can take in to consideration the Mesos domain

kaiwalyajoshi (approved):

PR #118, opened 2019-10-28
Consume same Mesos master addresses as libprocess.

jeschkies:
Summary:
The Mesos scheduler consumes Mesos master addresses in the following
formats:
   * host:port
   * zk://host1:port1,host2:port2,.../path
   * zk://username:password@host1:port1,host2:port2,.../path

In case it's a `zk://` address we are fetching the Mesos master
information from the provided path. It follows the parsing and master
identification as described in the [High Availability](http://mesos.apache.org/documentation/latest/high-availability/#implementation-details)
documentation and implemented in [detector/zookeeper](https://github.com/apache/mesos/blob/master/src/master/detector/zookeeper.cpp).

JIRA issues: DCOS_OSS-5588, DCOS-53671


mesos-client/src/main/scala/com/mesosphere/mesos/client/MasterDetector.scala
```diff
@@ -0,0 +1,114 @@
...
+    }
+  }
+}
+
+case class Zookeeper(master: String) extends MasterDetector with StrictLogging {
+  require(master.startsWith("zk://"), s"$master does not start with zk://")
+
+  case class ZkUrl(auth: Option[String], servers: String, path: String)
+
+  implicit val mesosInfoRead: Reads[Protos.MasterInfo] = (
```
jeschkies: Could we parse using protobuf Json?

mesos-client/src/main/scala/com/mesosphere/mesos/client/MasterDetector.scala
```diff
@@ -0,0 +1,114 @@
...
+        client.getZookeeperClient.getConnectionTimeoutMs,
+        java.util.concurrent.TimeUnit.MILLISECONDS)) {
+      throw new IllegalStateException("Failed to connect to Zookeeper. Will exit now.")
+    }
+
+    val asyncClient = AsyncCuratorFramework.wrap(client)
+
+    async {
+      val children = await(asyncClient.getChildren.forPath(path).toScala).asScala
+      val leader = children.filter(_.startsWith("json.info")).min
```
jeschkies: Is there a Curator/Zookeeper recipe that does this for us?

mesos-client/src/main/scala/com/mesosphere/mesos/client/MasterDetector.scala
```diff
@@ -0,0 +1,126 @@
...
+    Protos.MasterInfo.newBuilder().setHostname(hostname).setPort(port).setId(id).setIp(ip).build()
+  }
+
+  override def isValid(): Boolean = Try(parse()).map(_ => true).getOrElse(false)
+
+  override def getMaster()(implicit ex: ExecutionContext): CompletionStage[URL] = {
+    val ZkUrl(auth, servers, path) = parse()
+
+    val client = {
+      val clientBuilder = CuratorFrameworkFactory.builder().connectString(servers).retryPolicy(new RetryOneTime(100))
```
zen-dog: magic constant ☹️

mesos-client/src/main/scala/com/mesosphere/mesos/client/MasterDetector.scala
```diff
@@ -0,0 +1,126 @@
...
+      (JsPath \ "port").read[Int] ~
+      (JsPath \ "id").read[String] ~
+      (JsPath \ "ip").read[Int]
+  ) { (hostname, port, id, ip) =>
+    Protos.MasterInfo.newBuilder().setHostname(hostname).setPort(port).setId(id).setIp(ip).build()
+  }
+
+  override def isValid(): Boolean = Try(parse()).map(_ => true).getOrElse(false)
+
+  override def getMaster()(implicit ex: ExecutionContext): CompletionStage[URL] = {
```
zen-dog: Why did you reimplement all this ZK connection/reading logic here? All of that already exists in `persistence-zookeeper` module and can look like:
```
  val client: CuratorFramework = zkClient()
  val factory: AsyncCuratorBuilderFactory = AsyncCuratorBuilderFactory(client)
  val store: ZooKeeperPersistenceStore = new ZooKeeperPersistenceStore(DummyMetrics, factory, parallelism = 1)

  val bytes = await(store.read(leaderPath))
```

zen-dog (changes requested):
Left a few comments

mesos-client/src/main/scala/com/mesosphere/mesos/client/MasterDetector.scala
jeschkies: I'm not sure I want to pull in `usi-persistence` and `usi-persistence-zookeeper`. I was actually hoping that there is a Curator recipe but I only found  [LeaderElection](https://curator.apache.org/curator-recipes/leader-election.html) which seems to participate in the election and is not a read only. Maybe @timcharper knows something.

zen-dog: > 'm not sure I want to pull in `usi-persistence` and `usi-persistence-zookeeper`.

I agree. That's why I would argue that this `MasterDetector` belongs someplace else. And no, there is no recipe for that - you're basically just reading a node.

zen-dog: Additionally: though thread-safe, it is considered _a bad practice to have > 1 zkClients_ in the project. It should be passed to the `MasterDetector`.

jeschkies: > you're basically just reading a node.

The children sorted and then one node. I wonder if this should in in an transaction  so that there is no new leader.

zen-dog: You can use `ZookeeperPersistenceStore` to read multiple nodes (and/or children) in a transaction. But you **can't** have this `filter.min` logic in one:
```
val children = await(asyncClient.getChildren.forPath(path).toScala)
val leader = children.filter(_.startsWith("json.info")).min
```

jeschkies: I moved it out. This is still ugly. We'd have to inject the client which I don't really like. Maybe I should just drop this and move the code into the Mesos plugin.

mesos-master-detector/build.gradle
```diff
@@ -0,0 +1,17 @@
...
+
+    // Transitive dependencies
+    compile project(':persistence-zookeeper')
+    compile "de.heikoseeberger:akka-http-play-json_$scalaVersion:1.26.0"
+    compile "org.apache.mesos:mesos:$mesosVersion"
+}
+
+publishing.publications {
+    mavenJava(MavenPublication) {
+        artifactId 'mesos-master-detector'
```
zen-dog, line 15: I was thinking about smth. like `mesos-util` or just `util`? 🤔 A module for such a small task seems somewhat excessive. Keep in mind, we will need more of these helpers (like leader-election etc.) so this might be worth it.

mesos-master-detector/src/main/scala/com/mesosphere/mesos/MasterDetector.scala
```diff
@@ -0,0 +1,136 @@
...
+  ) { (hostname, port, id, ip) =>
+    Protos.MasterInfo.newBuilder().setHostname(hostname).setPort(port).setId(id).setIp(ip).build()
+  }
+
+  override def isValid(): Boolean = Try(parse()).map(_ => true).getOrElse(false)
+
+  override def getMaster()(implicit ex: ExecutionContext): CompletionStage[URL] = {
+    val ZkUrl(auth, servers, path) = parse()
+
+    val client = {
```
zen-dog, line 76: `client` should be injected here. The overall project will initialize it and pass into different modules like persistency, mesos-client, (master-detection, leader-election etc.)

mesos-master-detector/build.gradle
timcharper: Let's just put this in the mesos-client

timcharper (approved):
My vote is that we just merge this with `mesos-client`

mesos-master-detector/src/main/scala/com/mesosphere/mesos/MasterDetector.scala
jeschkies: We cannot inject the client because the credentials come form the connection string. It's a chicken an egg problem.

mesos-master-detector/build.gradle
jeschkies: It was there but it would require a dependency on `persistence-zookeeper`. I think `mesos-master-detector` is fine until we have more uses cases. I don't like `util` packages because the often become dustbins for everything. Also, this would be rather a `zookeeper-util`.

mesos-master-detector/src/main/scala/com/mesosphere/mesos/MasterDetector.scala
jeschkies: I could inject a factory method: `clientFactory: (authDigest) -> CuratorFramework`.

zen-dog:
> My vote is that we just merge this with `mesos-client`

Imho that doesn't make things simpler because it introduces an unnecessary dependency on the `zookeeper-persistence` module but if you insist. 

- `mesos-client` should then depend on `implementation project(':zookeeper-persistence')`
- `MasterDetector::apply` method should get the `store: ZooKeeperPersistenceStore` injected

mesos-master-detector/src/main/scala/com/mesosphere/mesos/MasterDetector.scala
```diff
@@ -0,0 +1,137 @@
...
+      throw new IllegalStateException("Failed to connect to Zookeeper. Will exit now.")
+    }
+
+    val clientSettings = AsyncCuratorBuilderSettings(createOptions = Set.empty, compressedData = false)
+    val factory: AsyncCuratorBuilderFactory = AsyncCuratorBuilderFactory(client, clientSettings)
+    val store: ZooKeeperPersistenceStore = new ZooKeeperPersistenceStore(metrics, factory, parallelism = 1)
+
+    async {
+      val children = await(store.children(path, false)).get
+      logger.info(s"Found children $children")
```
zen-dog:
```suggestion
      logger.info(s"Found mesos leader node children $children")
```

mesos-master-detector/src/main/scala/com/mesosphere/mesos/MasterDetector.scala
```diff
@@ -0,0 +1,137 @@
...
+      val leader = children.filter(_.startsWith("json.info")).min
+
+      val leaderPath = s"$path/$leader"
+      logger.info(s"Connecting to Zookeeper at $servers and fetching Mesos master from $leaderPath.")
+
+      val Node(_, bytes) = await(store.read(leaderPath)).get
+      logger.info(s"Mesos leader data: ${bytes.decodeString(StandardCharsets.UTF_8)}")
+      client.close()
```
zen-dog: We should not keep a dangling zk connection in case something goes wrong. Wdyt about adding:
```
async {
...
}.recover{
      case t: Throwable =>
          logger.error("Failed to get mesos master leader node from ZK: ", t)
          client.close()
    }
```

mesos-master-detector/src/main/scala/com/mesosphere/mesos/MasterDetector.scala
```diff
@@ -0,0 +1,146 @@
...
+      val url = new URL(s"http://${masterInfo.getHostname}:${masterInfo.getPort}")
+      url
+    }
+
+    // Ensure Zookeeper client ist closed.
+    future.onComplete {
+      case Failure(t) =>
+        logger.error("Failed to get Mesos master leader node from ZK: ", t)
+        client.close()
+      case _ => ()
```
zen-dog: we should `client.close()` here too

zen-dog (approved):
This [last comment](https://github.com/mesosphere/usi/pull/118/files#r341637257) and 🚢 it

mesos-master-detector/src/main/scala/com/mesosphere/mesos/MasterDetector.scala
jeschkies: 🙈 of course.

mesos-master-detector/src/main/scala/com/mesosphere/mesos/MasterDetector.scala
```diff
@@ -0,0 +1,147 @@
...
+      val Node(_, bytes) = await(store.read(leaderPath)).get
+      logger.info(s"Mesos leader data: ${bytes.decodeString(StandardCharsets.UTF_8)}")
+
+      val masterInfo = Json.parse(bytes.decodeString(StandardCharsets.UTF_8)).as[Protos.MasterInfo]
+      // TODO: how do we know it's http or https.
+      val url = new URL(s"http://${masterInfo.getHostname}:${masterInfo.getPort}")
+      url
+    }
+
+    // Ensure Zookeeper client ist closed.
```
zen-dog:
```suggestion
    // Ensure Zookeeper client is closed.
```

PR #119, opened 2019-10-30
Exit JVM if an error is thrown or the main loop exits

ANeumann82:
Fixed some timeout issues on startup
Fixed issues with PodIds containing .
Exit JVM if an error is thrown or the main loop exits

JIRA issues: DCOS_OSS-5633

### DIFF
--- a/mesos-client/src/main/resources/reference.conf
+++ b/mesos-client/src/main/resources/reference.conf
@@ -33,17 +33,4 @@ akka {
     # outgoing connection might be closed if no call is issued to Mesos for a while
     idle-timeout = infinite
   }
-
-  stream {
-
-    # Default materializer settings
-    materializer {
-
-      # Cleanup leaked publishers and subscribers when they are not used within a given
-      # deadline
-      subscription-timeout {
-        timeout = 25s
-      }
-    }
-  }
 }
--- a/mesos-client/src/main/resources/reference.conf
+++ b/mesos-client/src/main/resources/reference.conf
@@ -33,4 +33,17 @@ akka {
     # outgoing connection might be closed if no call is issued to Mesos for a while
     idle-timeout = infinite
   }
+
+  stream {
+
+    # Default materializer settings
+    materializer {
+
+      # Cleanup leaked publishers and subscribers when they are not used within a given
+      # deadline
+      subscription-timeout {
+        timeout = 25s
+      }
+    }
+  }
 }
--- a/mesos-client/src/main/scala/com/mesosphere/mesos/client/Session.scala
+++ b/mesos-client/src/main/scala/com/mesosphere/mesos/client/Session.scala
@@ -4,14 +4,16 @@ import java.net.URL
 import akka.NotUsed
 import akka.actor.{Actor, ActorRef, ActorSystem, Props, Stash, Status}
 import akka.http.scaladsl.Http
-import akka.http.scaladsl.model.headers.{Authorization, HttpCredentials}
 import akka.http.scaladsl.model._
+import akka.http.scaladsl.model.headers.{Authorization, HttpCredentials}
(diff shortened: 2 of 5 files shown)


jeschkies (approved):

mesos-client/src/main/resources/reference.conf
```diff
@@ -33,4 +33,17 @@ akka {
     # outgoing connection might be closed if no call is issued to Mesos for a while
     idle-timeout = infinite
   }
-}
+
+  stream {
```
timcharper: As a library, I feel like we shouldn't rely on putting global configuration here for Akka. We should instead be adding the appropriate catch-all error catching (which will expose exceptions properly).

timcharper (changes requested):

mesos-client/src/main/resources/reference.conf
ANeumann82: Yeah, that does make sense, and I think you're correct there. I couldn't really figure out why we actually need that increased timeout, something slows down the whole startup sequence, and we should probably clean that up instead of adjusting the timeout.

ANeumann82: @timcharper I've figured it out: The ConnectionPoolSettings creation does a reverse DNS lookup which is really slow on my macbook, which triggered the timeout. I've added a timeout check there and throw a reasonable error message now, so users of the lib will see something useful.

timcharper (changes requested):
Let's make the timeout configurable for the reverse-DNS lookup. Also, let's consider a function which takes `Future[Flow[I, O, M]]` and returns `Flow[I, O, M]`

mesos-client/src/main/scala/com/mesosphere/mesos/client/Session.scala
```diff
@@ -52,6 +54,28 @@ case class Session(url: URL, streamId: String, authorization: Option[Credentials
         Flow[Array[Byte]].map(createPostRequest(_, None)).via(connection)
     }
 
+  private def poolSettings():Future[ConnectionPoolSettings] = Future {
```
jeschkies, line 57: What is the reason for this being a future?

mesos-client/src/main/scala/com/mesosphere/mesos/client/Session.scala
```diff
@@ -52,6 +54,28 @@ case class Session(url: URL, streamId: String, authorization: Option[Credentials
...
+    // *IMPORTANT*: DO NOT CHANGE maxConnections OR pipeliningLimit! Otherwise, USI won't guarantee request order to Mesos!
+    ConnectionPoolSettings("").withMaxConnections(1).withPipeliningLimit(1)
+  }(ExecutionContext.global)
+
+  private def poolSettingsWithTimeout():ConnectionPoolSettings = {
+    // We create the pool settings with a timeout: It seems by default, the creation of the settings requires a reverse
+    // dns lookup, which may take more than 5 seconds which triggers the akka streams subscription timeout.
+    // If we let the creation here take the full time, we won't start without any reasonable logging message
+    try {
+      Await.result(poolSettings(), 2.seconds)
```
jeschkies, line 70: I don't get this. I'm sorry. Why can we not document the fact that apps should increase the Akka stream subscription timeout? It feels we are just introducing complexity here.

jeschkies (changes requested):
Could we move the DNS lookup timeout into a different pull request? DCOS_OSS-5633 is solved by shutting down the Akka Actor system. That is all that was required. This seems to be some scope creep to me.

DominikDary:
This PR is still relevant, right?

jeschkies:
@DominikDary, yes. @timcharper or me should take this over.

PR #122, opened 2019-11-04
Configure executor registration timeout.

jeschkies:
Summary:
If the executor needs too long to download a Docker image it's killed.
This should enable stabilizing tests.

### DIFF
--- a/test-utils/src/main/scala/com/mesosphere/utils/mesos/MesosTest.scala
+++ b/test-utils/src/main/scala/com/mesosphere/utils/mesos/MesosTest.scala
@@ -178,8 +178,8 @@ case class MesosCluster(
       ) ++ mesosFaultDomainAgentCmdOption.map(fd => s"--domain=$fd")
         ++ agentsConfig.seccompConfigDir.map(dir => s"--seccomp_config_dir=$dir")
         ++ agentsConfig.seccompProfileName.map(prf => s"--seccomp_profile_name=$prf")
-        ++ agentsConfig.executorRegistrationTimeout.map(t => s"--executor_registration_timeout=${t.toMinutes}mins")
-        ++ agentsConfig.fetcherStallTimeout.map(t => s"--fetcher_stall_timeout=${t.toMinutes}mins")
+        ++ agentsConfig.executorRegistrationTimeout.map(t => s"--executor_registration_timeout=${t.toSeconds}secs")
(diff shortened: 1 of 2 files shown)


test-utils/src/main/scala/com/mesosphere/utils/mesos/MesosTest.scala
```diff
@@ -31,14 +31,23 @@ import scala.util.Try
...
     imageProviders: Option[String] = None,
     seccompConfigDir: Option[String] = None,
-    seccompProfileName: Option[String] = None) {
+    seccompProfileName: Option[String] = None,
+    executorRegistrationTimeout: Option[Duration] = None,
+    fetcherStallTimeout: Option[Duration] = None) {
+
+  require(
+    executorRegistrationTimeout.getOrElse(1.minute) >= fetcherStallTimeout.getOrElse(1.minute),
+    s"Executor registration timeout $executorRegistrationTimeout must be bigger than fetcher stall timout $fetcherStallTimeout."
```
zen-dog:
```suggestion
    s"Executor registration timeout $executorRegistrationTimeout must be bigger or equal than fetcher stall timeout $fetcherStallTimeout."
```

test-utils/src/main/scala/com/mesosphere/utils/mesos/MesosTest.scala
```diff
@@ -31,14 +31,23 @@ import scala.util.Try
...
   */
 case class MesosAgentConfig(
     launcher: String = "posix",
     containerizers: String = "mesos",
     isolation: Option[String] = None,
     imageProviders: Option[String] = None,
     seccompConfigDir: Option[String] = None,
-    seccompProfileName: Option[String] = None) {
+    seccompProfileName: Option[String] = None,
+    executorRegistrationTimeout: Option[Duration] = None,
```
zen-dog: Where are `executorRegistrationTimeout` and `fetcherStallTimeout` set in tests? Maybe instead of `Option[Duration] = None` use `Duration = 1.seconds` directly?

jeschkies: The Jenkins Plugin uses them. The defaults are from Mesos. I think you are correct. `Option` is just confusing here.

zen-dog (approved):

PR #124, opened 2019-11-11
Multi-role, debounced, robust suppress and revive

timcharper:
We introduce a debounced suppress-and-revive mechanism which properly supports multiple Mesos roles (via the UpdateFramework calls).

Further, we introduce the factory pattern to decouple stream wiring logic from component initialization and respective dependencies.

Also, a generated-from-code mechanism is added to generate the visualization docs/scheduler-stream.svg; since this is generated from code, it will not be wrong. The graph is updated when the unit tests are run.


kaiwalyajoshi (dismissed):
An ASCII graph of the new streams would really help.

kaiwalyajoshi (approved):

core/src/main/scala/com/mesosphere/usi/core/SchedulerFactory.scala
```diff
@@ -0,0 +1,59 @@
...
+import com.mesosphere.usi.core.models.commands.SchedulerCommand
+import com.mesosphere.usi.core.revive.SuppressReviveHandler
+import com.mesosphere.usi.core.util.DurationConverters
+import com.mesosphere.usi.metrics.Metrics
+import com.mesosphere.usi.repository.PodRecordRepository
+import org.apache.mesos.v1.scheduler.Protos
+import org.apache.mesos.v1.scheduler.Protos.{Event => MesosEvent}
+import scala.concurrent.Future
+
+trait SchedulerLogicFactory {
```
jeschkies, line 18: Could you add docs?

core/src/main/scala/com/mesosphere/usi/core/SchedulerFactory.scala
```diff
@@ -0,0 +1,59 @@
...
+
+trait PersistenceFlowFactory {
+  private[usi] def newPersistenceFlow(): Flow[SchedulerEvents, SchedulerEvents, NotUsed]
+}
+
+trait SuppressReviveFactory {
+  def newSuppressReviveFlow: Flow[PodSpecUpdatedEvent, Protos.Call, NotUsed]
+}
+
+case class SchedulerFactory(
```
jeschkies, line 31: In which case would I not use this factory?

jeschkies (dismissed):
Thanks for the effort. I probably missed some discussion but this pull requests seems a little too big to me. Especially since it introduces multiple new concepts. The factories are orthogonal to the suppress-revive debouncing, isn't it?

core/src/main/scala/com/mesosphere/usi/core/revive/SuppressReviveHandler.scala
```diff
@@ -0,0 +1,192 @@
...
+  private[this] val reviveCountMetric: Counter = metrics.counter("usi.mesos.calls.revive")
+  private[this] val suppressCountMetric: Counter = metrics.counter("usi.mesos.calls.suppress")
+
+  private def defaultRoles = initialFrameworkInfo.getRolesList.asScala
+
+  /**
+    * Given a stream of PodSpecUpdatedEvent (which gives us the signal of which pods want offers), we emit a snapshot
+    * describing all roles to which the framework is subscribed, and the set of podIds which want offers for each role.
+    */
+  private[revive] def reviveStateFromPodSpecs: Flow[PodSpecUpdatedEvent, PodIdsWantingRoles, NotUsed] =
```
jeschkies, line 38: I have to admit that I'm not a big fan to split this logic out into a flow. The original `suppressAndRevive` was a simple easy to test method. What is the benefit of mixing the logic with streams?

core/src/main/scala/com/mesosphere/usi/core/util/RateLimiterFlow.scala
```diff
@@ -0,0 +1,70 @@
...
+import java.time.{Clock, Duration => JavaDuration}
+
+import akka.NotUsed
+import akka.stream.scaladsl.Flow
+import akka.stream.stage._
+import akka.stream.{Attributes, FlowShape, Inlet, Outlet}
+
+import scala.concurrent.duration.FiniteDuration
+
+private[usi] class RateLimiterFlow[U] private (rate: FiniteDuration, clock: Clock) extends GraphStage[FlowShape[U, U]] {
```
jeschkies, line 12: Nice. I actually wanted to point out that this was in Marathon already. What do you think about opening up a PR in Marathon to replace the Marathon implementation with this? We already have a test dependency on USI.

core/src/main/scala/com/mesosphere/usi/core/revive/SuppressReviveHandler.scala
timcharper: Debouncing

core/src/main/scala/com/mesosphere/usi/core/util/RateLimiterFlow.scala
timcharper: I'd rather not right now

core/src/main/scala/com/mesosphere/usi/core/SchedulerFactory.scala
timcharper: You rejected for this?

jeschkies:
> no actual rejection reason

🙁 man. There are a couple:
* The diff is too big and conflates multiple orthogonal changes.
* The interface change for the scheduler creation creates unnecessary work for plugin which has the highest priority as far as I am aware.
* Important parts were not documented.
* I am not a fan of the decision to move important logic into another flow.

PR #125, opened 2019-11-20
Accept optional domain filter for launching pods.

jeschkies:
JIRA issues: DCOS-61122

### DIFF
--- a/core-models/src/main/scala/com/mesosphere/usi/core/models/commands/SchedulerCommand.scala
+++ b/core-models/src/main/scala/com/mesosphere/usi/core/models/commands/SchedulerCommand.scala
@@ -1,6 +1,6 @@
 package com.mesosphere.usi.core.models.commands
 
-import com.mesosphere.usi.core.models.faultdomain.DomainFilter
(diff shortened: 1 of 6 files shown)


timcharper (approved):

core-models/src/main/scala/com/mesosphere/usi/core/models/commands/SchedulerCommand.scala
```diff
@@ -28,8 +29,10 @@ sealed trait SchedulerCommand
   *
   * @param podId
   * @param runSpec
+  * @param domainFilter An optional filter for default domains. See [[DomainFilter]] for details.
   */
-case class LaunchPod(podId: PodId, runSpec: RunTemplate) extends SchedulerCommand
+case class LaunchPod(podId: PodId, runSpec: RunTemplate, domainFilter: Option[DomainFilter] = None)
```
timcharper: I'd like to discuss this please; what are your thoughts on defaulting it to None vs just setting it HomeRegionFilter?

jeschkies: I did not give it much thought yesterday. `HomeRegionFilter` is much better here.

PR #126, opened 2019-11-21
Filter agent attributes.

kaiwalyajoshi:
Draft version of Attribute Filtering.


core-models/src/main/scala/com/mesosphere/usi/core/models/commands/SchedulerCommand.scala
```diff
@@ -29,7 +30,8 @@ sealed trait SchedulerCommand
   * @param podId
   * @param runSpec
   */
-case class LaunchPod(podId: PodId, runSpec: RunTemplate) extends SchedulerCommand
+case class LaunchPod(podId: PodId, runSpec: RunTemplate, agentFilter: AgentFilter = DefaultAgentFilter)
```
jeschkies: Nice. This is how I imagined it.

core-models/src/main/scala/com/mesosphere/usi/core/models/constraints/AgentAttributeFilter.scala
```diff
@@ -0,0 +1,51 @@
+package com.mesosphere.usi.core.models.constraints
+
+import org.apache.mesos.v1.Protos
+import collection.JavaConverters._
+
+import com.typesafe.scalalogging.StrictLogging
+
+case class AgentAttributeFilter(attributes: Map[String, Any]) extends AgentFilter with StrictLogging {
```
jeschkies: And now I get what @timcharper meant 🙂 If I understand an attribute correctly it can be a scalar, range or text. We could type that. However, for Jenkins we only need text if I'm not mistaken. What do you think about a simple `AgentStringAttributeFilter` that takes a `Map[String, String]`. It would just be false in case the attribute is not a string and issue a warning.

jeschkies (approved):

PR #127, opened 2019-11-21
Define any domain filter.

jeschkies:
Summary:
The `AnyDomain` filter will accept any fault domain.


core-models/src/main/scala/com/mesosphere/usi/core/models/faultdomain/AnyDomain.scala
```diff
@@ -0,0 +1,7 @@
+package com.mesosphere.usi.core.models.faultdomain
+import org.apache.mesos.v1.Protos
+
+object AnyDomain extends DomainFilter {
```
jeschkies, line 4: This would be used by the Jenkins plugin. See eg https://github.com/jenkinsci/mesos-plugin/pull/370

timcharper (approved):

PR #130, opened 2019-11-26
Provide collection of agentFilters

timcharper:
Summary:
In this change, we modify `agentFilters` to be a collection, rather than a single object. Each filter is applied with AND semantic.

Further, we introduce debug logging for `agentFilter` rejections.

Also, we introduce a convention for JAPI methods, and set a trend for moving away from case classes and default-values (two causes of unintentional binary breakages in Scala libraries): https://docs.scala-lang.org/overviews/core/binary-compatibility-for-library-authors.html#evolving-code-without-breaking-binary-compatibility


core/src/main/scala/com/mesosphere/usi/core/matching/OfferMatcher.scala
```diff
@@ -42,14 +43,23 @@ class OfferMatcher(masterDomainInfo: Mesos.DomainInfo) {
...
 
-      case podSpec :: rest if !(podSpec.domainFilter(masterDomainInfo, originalOffer.getDomain)) =>
-        matchPodSpecsTaskRecords(originalOffer, remainingResources, result, rest)
-
-      case podSpec :: rest if !(podSpec.agentFilter(originalOffer)) =>
-        matchPodSpecsTaskRecords(originalOffer, remainingResources, result, rest)
-
       case podSpec :: rest =>
-        maybeMatchResourceRequirements(remainingResources, Nil, podSpec.runSpec.allResourceRequirements) match {
+        Some(podSpec).filter { podSpec =>
```
kaiwalyajoshi: This feels more terse and less readable than the previous variant.

timcharper: This is a fair point

core-models/src/main/scala/com/mesosphere/usi/core/models/constraints/AgentFilter.scala
```diff
@@ -6,4 +6,10 @@ import org.apache.mesos.v1.Protos
 // Mostly used for attribute filtering.
 trait AgentFilter {
   def apply(offer: Protos.Offer): Boolean
+
+  /**
+    * Return a text description of the matcher
+    * @return
```
jeschkies: Nit:
```scala
/** @return a text description of the matcher */
```

core/src/main/scala/com/mesosphere/usi/core/matching/OfferMatcher.scala
```diff
@@ -38,14 +40,27 @@ class OfferMatcher(masterDomainInfo: Mesos.DomainInfo) {
       result: Map[RunningPodSpec, List[OfferMatcher.ResourceMatch]],
       pendingLaunchPodSpecs: List[RunningPodSpec]): Map[RunningPodSpec, List[OfferMatcher.ResourceMatch]] = {
 
+    def matchesAgentFilter(agentFilters: Iterable[AgentFilter]): Boolean = {
```
jeschkies:
```suggestion
    def matchesAgentFilters(agentFilters: Iterable[AgentFilter]): Boolean = {
```

Just me nit pickting.

core/src/main/scala/com/mesosphere/usi/core/matching/OfferMatcher.scala
```diff
@@ -38,14 +40,27 @@ class OfferMatcher(masterDomainInfo: Mesos.DomainInfo) {
...
+      agentFilters.find { filter =>
+        !filter(originalOffer)
+      } match {
+        case Some(nonMatchingfilter) =>
+          logger.debug(
+            s"Declining offer ${originalOffer.getId.getValue} for pod ${podSpec.id}; first non-matching agent filter: ${nonMatchingfilter.description}")
```
jeschkies: I'd prefer to show all filters that do not match.

jeschkies (approved):
This is so much cleaner. I have only a few nits.

core/src/main/scala/com/mesosphere/usi/core/matching/OfferMatcher.scala
timcharper: I know; I thought of this. However, this requires additional evaluation. Really worth it? Knowing that one filter didn't match is probably enough, and in Marathon / Metronome we just expose "constraints didn't match"

kaiwalyajoshi (approved):

PR #132, opened 2019-12-12
Fail Mesos client if master version is too old.

jeschkies:
Summary:
The Jenkins plugin reported `HTTP 400` errors. To ensure that the Mesos
master version is compatible we are checking it during the client
creation.


mesos-client/src/main/scala/com/mesosphere/mesos/conf/SemanticVersion.scala
```diff
@@ -0,0 +1,29 @@
+package com.mesosphere.mesos.conf
```
jeschkies, line 1: This is imported from Marathon.

kaiwalyajoshi (approved):

timcharper (approved):

PR #133, opened 2020-01-02
Explain why domain filter declined offer.

jeschkies:
Summary:
The domain filter would not give a reason why an offer was rejected.
This change introduces a similar log line to the agent filter.

### DIFF
--- a/core-models/src/main/scala/com/mesosphere/usi/core/models/faultdomain/RegionFilter.scala
+++ b/core-models/src/main/scala/com/mesosphere/usi/core/models/faultdomain/RegionFilter.scala
@@ -5,4 +5,6 @@ case class RegionFilter(region: String) extends DomainFilter {
   override def apply(masterDomain: Protos.DomainInfo, nodeDomain: Protos.DomainInfo): Boolean = {
     region == nodeDomain.getFaultDomain.getRegion.getName
   }
+
+  override def description: String = s"expect region $region"
 }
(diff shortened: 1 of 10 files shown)


core-models/src/main/scala/com/mesosphere/usi/core/models/faultdomain/RegionFilter.scala
```diff
@@ -5,4 +5,6 @@ case class RegionFilter(region: String) extends DomainFilter {
   override def apply(masterDomain: Protos.DomainInfo, nodeDomain: Protos.DomainInfo): Boolean = {
```
jeschkies, line 5: What do you think about a `MatchResult` type?

timcharper: Hmm, maybe?  I was hoping it'd be enough detail to have the descriptions of the filters that didn't match:

```
Offer did not match; 1 filter didn't match: "expect region 'windows' 
```

For the offer match debugging result that's processed by the implementation framework, the actual non-matching filters and matching filters will be passed

core/src/main/scala/com/mesosphere/usi/core/matching/OfferMatcher.scala
```diff
@@ -53,11 +54,19 @@ class OfferMatcher(masterDomainInfo: Mesos.DomainInfo) extends StrictLogging {
       }
     }
 
+    def matchesDomainFilter(podId: PodId, domainFilter: DomainFilter): Boolean = {
+      val result = domainFilter(masterDomainInfo, originalOffer.getDomain)
+      if (!result)
+        logger.debug(
+          s"Declining offer ${originalOffer.getId.getValue} for pod $podId; domain filter: ${domainFilter.description}.")
```
timcharper:
```suggestion
          s"Declining offer ${originalOffer.getId.getValue} for pod $podId; domain filter did not match: ${domainFilter.description}.")(
            LoggingArgs("podId" -> podId)
          )
```

timcharper (approved):

timcharper:
lgtm; can you please merge the suggestion?

PR #134, opened 2020-01-15
Include /slaves endpoint to list agents

timcharper:
Also expose active agent value


test-utils/src/main/scala/com/mesosphere/utils/mesos/MesosFormats.scala
```diff
@@ -60,9 +61,11 @@ object MesosFormats {
       (__ \ "unregistered_frameworks").format[Seq[String]]
   )(ITMesosState.apply, unlift(ITMesosState.unapply))
 
-  implicit lazy val ITaskFormat: Format[ITask] = Json.format[ITask]
+  implicit lazy val ITaskFormat: Format[ITTask] = Json.format[ITTask]
```
jeschkies, line 64: That's a fun misspelling 🙂

jeschkies (approved):

PR #136, opened 2020-01-22
Parse Mesos master ip properly.

jeschkies:
Summary:
The Mesos master IP is saved as a four bytes integer. It can be
eg `3355709612`. This is the unsigned representation of the four
bytes. The signed representation would be `939257684`.

### DIFF
--- a/mesos-master-detector/src/main/scala/com/mesosphere/mesos/MasterDetector.scala
+++ b/mesos-master-detector/src/main/scala/com/mesosphere/mesos/MasterDetector.scala
@@ -59,13 +59,21 @@ case class Zookeeper(master: String, metrics: Metrics) extends MasterDetector wi
 
   case class ZkUrl(auth: Option[String], servers: String, path: String)
 
+  // This is the JSON format for the Mesos master information saved in Zookeeper.
+  // According to the docs of `setIp`:
+  // The IP address (only IPv4) as a packed 4-bytes integer, stored in network order.
+  //
+  // Scala has no unsigned integer. That means we cannot read eg `3355709612` as int but as long. The `toInt` call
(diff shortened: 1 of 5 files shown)


mesos-master-detector/src/main/scala/com/mesosphere/mesos/MasterDetector.scala
```diff
@@ -59,13 +59,21 @@ case class Zookeeper(master: String, metrics: Metrics) extends MasterDetector wi
 
   case class ZkUrl(auth: Option[String], servers: String, path: String)
 
+  // This is the JSON format for the Mesos master information saved in Zookeeper.
+  // According to the docs of `setIp`:
+  // The IP address (only IPv4) as a packed 4-bytes integer, stored in network order.
+  //
+  // Scala has no unsigned integer. That means we cannot read eg `3355709612` as int but as long. The `toInt` call
+  // converts the long to a signed integer `-939257684`. Note that we do not care about the number but the bytes.
```
jeschkies, line 75: Please checkout my Rust example for converting the bytes https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=66105b1f4b3b2d27bb76076dd0629613

jongiddy:
I note that https://issues.apache.org/jira/browse/MESOS-1201 seems to imply that the top-level `ip` integer field can return IP's in either big-endian or little-endian order (the example seems to be in little-endian despite the Mesos docs saying it is in network order = big-endian).

Hence, I'm expecting that for reading a Mesos MasterInfo record, we should rely on the `address/ip` field which is a dotted decimal string.

timcharper (approved):

PR #138, opened 2020-01-24
Require that principal is set for authz.

jeschkies:
Summary:
This should ease some usage. The principal must be set in the framework
info when authorization is used. `UPDATE_FRAMEWORK` calls might fails
otherwise.


jeschkies:
I've found that we have a `Linting` object that is not used. We should bring this back. Eg if a domain filter is defined that is not `AnyDomain` or `HomeRegion` and the framework info does not defined the partition awareness, we should issue a warning.

kaiwalyajoshi (approved):

timcharper (approved):
This is completely reasonable

PR #144, opened 2020-04-02
Don't specify cgroups_enable_cfs option for non-linux

timcharper:
Also, refactor Agent class to accept and render the modeled agent
config, so it can be done more consistently


jeschkies (approved):
Much cleaner. Thanks!

PR #150, opened 2020-05-14
Throw exception if the login response is not OK.

jeschkies:


jeschkies:
I need to add a test tomorrow to see why the Actor would not log `Fetching the next IAM authentication token failed`.

jeschkies:
Ah, the initial `nextToken` fails [here](https://github.com/mesosphere/usi/blob/master/mesos-client/src/main/scala/com/mesosphere/mesos/client/MesosClient.scala#L184).

kaiwalyajoshi (approved):

